PR #4, opened 2014-04-02
[FIXED JENKINS-21669] Added File descriptor information for unix

christ66:
machines. This code will try to get the list of open files for the Jenkins instance.

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/FileDescriptorLimit.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/FileDescriptorLimit.java
@@ -1,0 +1,174 @@
+package com.cloudbees.jenkins.support.impl;
+
+import com.cloudbees.jenkins.support.api.Component;
+import com.cloudbees.jenkins.support.api.Container;
+import com.cloudbees.jenkins.support.api.Content;
+import com.sun.management.UnixOperatingSystemMXBean;
+import edu.umd.cs.findbugs.annotations.NonNull;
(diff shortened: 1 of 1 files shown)


src/main/java/com/cloudbees/jenkins/support/impl/FileDescriptorLimit.java
```diff
@@ -0,0 +1,174 @@
...
+    /**
+     * * This method executes the command "bash -c ulimit -a" on the machine. If an exception is thrown log it
+     * * to file and continue.
+     */
+    private static void getUlimit(OutputStream os) throws UnsupportedEncodingException {
+        final PrintWriter writer = new PrintWriter(new OutputStreamWriter(os, "utf-8"), true);
+        String osName = System.getProperty("os.name").toLowerCase();
+
+        if (osName != null && !osName.contains("win")) {
+            ProcessBuilder builder = new ProcessBuilder("bash", "-c", "ulimit -a");
```
stephenc: @christ66 I wonder if we shouldn't be getting both the hard and soft limits as I have seen cases where the limits reported by `-a` are not the actual soft limits but rather the hard limits, so you fail to see e.g. the file descriptor limit being reached

christ66: @stephenc What would be the suggested method for this instead.

Also this is generated on-the-fly, so most likely we won't see any issues unless they encounter it and immediately generate a bundle.

PR #11, opened 2014-10-08
[FIXED JENKINS-21668] Obtain root CA information for java instances

christ66:
This will only work if there is no password for obtaining the keystore.

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/RootCAs.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/RootCAs.java
@@ -42,15 +42,19 @@ import java.io.*;
 import java.security.KeyStore;
 import java.security.KeyStoreException;
 import java.util.Collections;
-import java.util.Map;
 import java.util.Set;
-import java.util.WeakHashMap;
 
 /**
  * @author schristou88
  */
 @Extension
-public class RootCAs extends Component{
+public class RootCAs extends Component {
+
+  @Override
+  public boolean isSelectedByDefault() {
+    return false;
+  }
+
   @NonNull
   @Override
   public Set<Permission> getRequiredPermissions() {
--- a/src/main/java/com/cloudbees/jenkins/support/impl/RootCAs.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/RootCAs.java
@@ -81,12 +81,6 @@ public class RootCAs extends Component {
     if (c == null) {
       return;
     }
-    if (c instanceof SlaveComputer && !Boolean.TRUE.equals(((SlaveComputer) c).isUnix())) {
-      return;
-    }
-    if (!node.createLauncher(TaskListener.NULL).isUnix()) {
-      return;
-    }
     String name;
     if (node instanceof Jenkins) {
       name = "master";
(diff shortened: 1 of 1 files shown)


src/main/java/com/cloudbees/jenkins/support/impl/RootCAs.java
```diff
@@ -0,0 +1,149 @@
...
+    } else {
+      name = "slave/" + node.getNodeName();
+    }
+    container.add(
+            new Content("nodes/" + name + "/RootCA.txt") {
+              @Override
+              public void writeTo(OutputStream os) throws IOException {
+                PrintWriter out = new PrintWriter(new BufferedWriter(new OutputStreamWriter(os, "utf-8")));
+                try {
+                  out.println(getRootCA(node));
```
christ66: I removed the call to `Future<String>`.

src/main/java/com/cloudbees/jenkins/support/impl/RootCAs.java
```diff
@@ -0,0 +1,149 @@
...
+      instance = KeyStore.getInstance(KeyStore.getDefaultType());
+      while (instance.aliases().hasMoreElements()) {
+        String s = instance.aliases().nextElement();
+        writer.append("========");
+        writer.append("Alias: " + s);
+        writer.append(instance.getCertificate(s).getPublicKey().toString());
+        writer.append("Trusted certificate: " + instance.isCertificateEntry(s));
+      }
+    } catch (KeyStoreException e) {
+      writer.write(ExceptionUtils.getFullStackTrace(e));
```
jglick: There is a similar method in `hudson.Functions` BTW.

src/main/java/com/cloudbees/jenkins/support/impl/RootCAs.java
```diff
@@ -0,0 +1,149 @@
...
+
+    private static final long serialVersionUID = 1L;
+  }
+
+  public static void getRootCAList(StringWriter writer) {
+    KeyStore instance = null;
+    try {
+      instance = KeyStore.getInstance(KeyStore.getDefaultType());
+      while (instance.aliases().hasMoreElements()) {
+        String s = instance.aliases().nextElement();
```
jglick: This is wrong. Each call to `aliases()` creates a new `Enumeration`. So I would expect this to be an infinite loop, if there is at least one alias.

src/main/java/com/cloudbees/jenkins/support/impl/RootCAs.java
```diff
@@ -0,0 +1,149 @@
...
+        writer.append("========");
+        writer.append("Alias: " + s);
+        writer.append(instance.getCertificate(s).getPublicKey().toString());
+        writer.append("Trusted certificate: " + instance.isCertificateEntry(s));
```
jglick: Why are you repeating these last two things for each alias? Should be outside the loop I think.

christ66: Not sure I understand. Obtaining the certificate requires the alias, so the last two things should be in the loop.

christ66:
@reviewbybees

ndeloof:
:bee: :+1:

stephenc:
:bee:

stephenc:
@reviewbybees done (as @christ66 is on vacation)

stephenc:
(wearing plugin maintainer's hat) :+1:

PR #14, opened 2015-01-02
[JENKINS-20542] Add summary of current build queue.

christ66:
The support bundle reporting the build queue would be helpful for debugging any elements that are pending or blocked in the queue.

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/BuildQueue.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/BuildQueue.java
@@ -35,6 +35,7 @@ import jenkins.model.Jenkins;
 
 import java.io.*;
 import java.util.Collections;
+import java.util.List;
 import java.util.Set;
 
 /**
@@ -66,8 +67,8 @@ public class BuildQueue extends Component {
         public void writeTo(OutputStream os) throws IOException {
           PrintWriter out = new PrintWriter(new BufferedWriter(new OutputStreamWriter(os, "utf-8")));
           try {
-            Queue.Item[] items = Jenkins.getInstance().getQueue().getItems();
-            out.println("Current build queue has " +  items.length + " item(s).");
+            List<Queue.Item> items = Jenkins.getInstance().getQueue().getApproximateItemsQuickly();
+            out.println("Current build queue has " +  items.size() + " item(s).");
             out.println("---------------");
 
             for (Queue.Item item : items) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/cloudbees/jenkins/support/impl/BuildQueue.java
```diff
@@ -0,0 +1,93 @@
...
+        @Override
+        public void writeTo(OutputStream os) throws IOException {
+          PrintWriter out = new PrintWriter(new BufferedWriter(new OutputStreamWriter(os, "utf-8")));
+          try {
+            List<Queue.Item> items = Jenkins.getInstance().getQueue().getApproximateItemsQuickly();
+            out.println("Current build queue has " +  items.size() + " item(s).");
+            out.println("---------------");
+
+            for (Queue.Item item : items) {
+              out.println(" * Name of item: " + item.task.getFullDisplayName());
```
jglick: Might require escaping. Check around in other components to see how display names are rendered.

jglick: Can also check for `item instanceof [hudson.model.]Item` and report `Item.fullName` instead, which (inside backticks) is better suited to support bundle files.

jglick:
Other than those minor things, LGTM.

src/test/java/com/cloudbees/jenkins/support/impl/BuildQueueTest.java
```diff
@@ -0,0 +1,94 @@
...
+    queue.addContents(createContainer(baos));
+
+    // Then
+    build.cancel(true);
+    List<String> output = new ArrayList<String>(Arrays.asList(baos.toString().split("\n")));
+
+
+    assertContains(output, "1 item");
+    assertContains(output, p.getDisplayName());
+    assertContains(output, "Waiting for next available executor");
```
jglick: FYI, I just got a random failure in this line while running prerelease tests.

```
[INFO] Failed tests: 
[INFO]   BuildQueueTest.verifyMinimumBuildQueue:75->assertContains:79 
[INFO] Expected: (a collection containing a string containing "Waiting for next available executor")
[INFO]      but: a collection containing a string containing "Waiting for next available executor" was "Current build queue has 1 item(s).", was "---------------", was " * Name of item: bar", was "    - In queue for: 4 ms", was "    - Is blocked: false", was "    - Why in queue: ???", was "    - Current queue trigger cause: Legacy code started this job.  No cause information is available"
```

PR #18, opened 2015-01-09
Add option to dump export tables for slave nodes.

christ66:
Could be useful for detecting certain types of memory leaks.

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/api/TruncatedContent.java
+++ b/src/main/java/com/cloudbees/jenkins/support/api/TruncatedContent.java
@@ -1,0 +1,97 @@
+/*
+ * The MIT License
+ * 
+ * Copyright (c) 2015 schristou88
+ * 
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ * 
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ * 
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package com.cloudbees.jenkins.support.api;
+
+import com.cloudbees.jenkins.support.timer.FileListCapComponent;
+
+import java.io.*;
+
+/**
+ * Content added to the support bundle that should be truncated.
+ *
+ * @author schristou88
+ * @since 2.26
+ */
+public abstract class TruncatedContent extends Content {
+  private int maxSize;
+
+  public TruncatedContent(String name) {
+    this(name, FileListCapComponent.MAX_FILE_SIZE);
+  }
+
+  public TruncatedContent(String name, int maxSize) {
+    super(name);
+    this.maxSize = maxSize;
+  }
+
+  @Override
+  public void writeTo(OutputStream os) throws IOException {
+    PrintWriter out = new PrintWriter(new BufferedWriter(new OutputStreamWriter(new TruncatedOutputStream(os, maxSize), "UTF-8")));
+    try {
+      printTo(out);
+    } finally {
+      out.flush();
+    }
+  }
+
+  /**
+   * {$Content} that will be printed to a specific {$PrintWriter}.
+   *
+   * @param out PrintWriter to print to
+   * @throws IOException
+   */
+  protected abstract void printTo(PrintWriter out) throws IOException;
+
+  public static class TruncatedOutputStream extends FilterOutputStream {
+    private final int max;
+    private int currentSize;
+
+    public TruncatedOutputStream(OutputStream os, int max) {
+      super(os);
+      this.max = max;
+      this.currentSize = 0;
+    }
+
+    @Override
+    public void write(int b) throws IOException {
+      if (max >= currentSize) {
+        out.write(b);
+        currentSize++;
+      } else {
+        throw new TruncationException("Max file size reached: " + max);
+      }
+    }
+
+    @Override
+    public void write(byte[] b, int off, int len) throws IOException {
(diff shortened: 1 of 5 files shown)


christ66:
@reviewbybees

tfennelly:
looks good :+1:

src/test/java/com/cloudbees/jenkins/support/SupportTestUtils.java
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * User: schristou88
+ * Date: 1/8/15
+ * Time: 4:35 PM
+ */
+public class SupportTestUtils {
+
+  /**
+   * Create a container from an output stream. Test units can use this to generate a
```
jglick: Truncated comment.

src/test/java/com/cloudbees/jenkins/support/SupportTestUtils.java
```diff
@@ -0,0 +1,32 @@
...
+      @Override
+      public void add(@CheckForNull Content content) {
+        try {
+          content.writeTo(os);
+        } catch (IOException e) {
+          e.printStackTrace();
+        }
+      }
+    };
+  }
```
jglick: Would be more convenient to take a `Component` and return a `String` (via `ByteArrayOutputStream`).

jglick:
Looks like it could be useful. The main risk is that the export table is huge and swamps the bundle. Is it in practice limited to some reasonable size by the way Remoting is used?

christ66:
Truncate would be nice, but might not capture everything unfortunately.

christ66:
@jglick Would it make sense for us to remove duplicates instead of truncating?

jglick:
> Would it make sense for us to remove duplicates instead of truncating?

Sure. Maybe nothing needs to be done—maybe the output as it stands is naturally limited to only a couple dozen lines or so. I am just asking what environments you have tested this in and whether you can be confident that the output will be reasonable, or whether some kind of truncation/deduplication might be necessary.

src/main/java/com/cloudbees/jenkins/support/impl/DumpExportTable.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.Collections;
+import java.util.Set;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Dump export table of nodes to detect potential
+ * memory leaks.
+ * 
+ * User: schristou88
```
stephenc: replace the template comments with an @since tag

stephenc:
It might help to put a time bound for each node and truncate if the time bound is exceeded... though as this is a local only operation from my reading of the code that may be an unnecessary concern

stephenc:
@christ66 where are you with closing this one out?

stephenc:
@christ66 ping!

jglick:
Seems to be stalled.

daniel-beck:
@christ66 Is this work in progress or can it be reviewed?

christ66:
@jglick @stephenc I added a `TruncateComponent` class which should handle the exportTable dump being very large. I wrote a unit test to confirm that the export table can contain duplicate content, and could grow to be very large. The test also confirms that my `TruncateComponent` will properly truncate the export table dump if it grows large. I _believe_ if the export table size is over 2MB in size that we would most likely have enough information to determine what _could potentially_ be causing a memory leak.

@daniel-beck I am removing the WIP and it can be reviewed.

src/main/java/com/cloudbees/jenkins/support/api/TruncatedContent.java
```diff
@@ -0,0 +1,86 @@
...
+  public void writeTo(OutputStream os) throws IOException {
+    PrintWriter out = new PrintWriter(new BufferedWriter(new OutputStreamWriter(new TruncatedOutputStream(os, maxSize), "UTF-8")));
+    try {
+      printTo(out);
+    } finally {
+      out.flush();
+    }
+  }
+
+  protected abstract void printTo(PrintWriter out) throws IOException;
```
jtnord: as this is intended to be implemented in sub-classes a little javadoc would go a long way...

stephenc:
:bug:

The problem that I have with the current truncated content and output stream implementation is that they do not stop the iteration of all the objects and potentially costly outputting of allocation and release stack traces... all they do is limit the size of content pushed to the bundle.

What I would do with truncation is actually throw an exception (subclassing IOException) when the limit is reached so that the write method bombs out (you can then catch the specific TruncationException and `//ignore`` it)

christ66:
@stephenc I implemented the `TruncationException`. Instead of ignoring I added a logging event that it happened.

stephenc:
:bee:

src/main/java/com/cloudbees/jenkins/support/impl/DumpExportTable.java
```diff
@@ -0,0 +1,73 @@
...
+
+  @NonNull
+  @Override
+  public String getDisplayName() {
+    return "Dump slave export tables (could reveal some memory leaks)";
+  }
+
+  @Override
+  public void addContents(@NonNull Container result) {
+    for (final Node node : Jenkins.getInstance().getNodes()) {
```
oleg-nenashev: Minor: Jenkins.getInstance() may return null

src/test/java/com/cloudbees/jenkins/support/SupportTestUtils.java
```diff
@@ -0,0 +1,37 @@
...
+import com.cloudbees.jenkins.support.api.Component;
+import com.cloudbees.jenkins.support.api.Container;
+import com.cloudbees.jenkins.support.api.Content;
+import edu.umd.cs.findbugs.annotations.CheckForNull;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+
+/**
+ * User: schristou88
```
oleg-nenashev: If you create Javadocs, please add information details there. Date and Time are useless, use <code>@since</code> to refer the plugin version, where the fix will be released

src/main/java/com/cloudbees/jenkins/support/api/TruncatedContent.java
```diff
@@ -0,0 +1,96 @@
...
+      super(os);
+      this.max = max;
+      this.currentSize = 0;
+    }
+
+    @Override
+    public void write(int b) throws IOException {
+      if (max > currentSize) {
+        out.write(b);
+        currentSize++;
```
oleg-nenashev: You start counting bytes from 0, so the check is  not correct. If max is 8000, 8001 bytes can be written

src/main/java/com/cloudbees/jenkins/support/api/TruncatedContent.java
```diff
@@ -0,0 +1,96 @@
...
+package com.cloudbees.jenkins.support.api;
+
+import com.cloudbees.jenkins.support.timer.FileListCapComponent;
+
+import java.io.*;
+
+/**
+ * Content added to the support bundle that should be truncated.
+ *
+ * @author schristou88
```
oleg-nenashev: Needs <code>@since</code>, because it's a useful API class

oleg-nenashev:
:bug: (ranges check)

src/main/java/com/cloudbees/jenkins/support/api/TruncatedContent.java
christ66: Fixed.

oleg-nenashev:
:bee: 
Could you squash the commits into a single one?

christ66:
@oleg-nenashev squashed!

oleg-nenashev:
:bee:  again

daniel-beck:
:bee:

christ66:
@reviewbybees done.

PR #20, opened 2015-01-19
[JENKINS-26483] Add JVM bit size to the support bundle.

christ66:
This will display 32/64 bit size of the masters JVM. This information can be obtained for slave nodes by looking for the `sun.arch.data.model` value in the `system.properties` file.

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
@@ -326,6 +326,7 @@ public class AboutJenkins extends Component {
             result.append(min).append(" Name:    ").append(System.getProperty("java.vm.name")).append("\n");
             result.append(min).append(" Vendor:  ").append(System.getProperty("java.vm.vendor")).append("\n");
             result.append(min).append(" Version: ").append(System.getProperty("java.vm.version")).append("\n");
+            result.append(min).append(" Bit size: ").append(System.getProperty("sun.arch.data.model")).append("\n");
             result.append(maj).append(" Operating system\n");
             result.append(min).append(" Name:         ").append(System.getProperty("os.name")).append("\n");
             result.append(min).append(" Architecture: ").append(System.getProperty("os.arch")).append("\n");


christ66:
@reviewbybees

src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
```diff
@@ -326,6 +326,7 @@ public String call() throws RuntimeException {
             result.append(min).append(" Name:    ").append(System.getProperty("java.vm.name")).append("\n");
             result.append(min).append(" Vendor:  ").append(System.getProperty("java.vm.vendor")).append("\n");
             result.append(min).append(" Version: ").append(System.getProperty("java.vm.version")).append("\n");
+            result.append(min).append(" Bit size: ").append(System.getProperty("sun.arch.data.model")).append("\n");
```
tfennelly: Is this always guaranteed to be a property? What if `null` is returned? You could supply a default I think, if that helped.

jglick:
Why is this needed at all? Existing bundles contain text such as

```
  * JVM Implementation
      - Name:    Java HotSpot(TM) 64-Bit Server VM
      - Vendor:  Oracle Corporation
      - Version: 24.51-b03
```

which seems enough.

src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
jglick: Definitely you need a fallback value for other JVMs.

ndeloof:
Last time we looked into a support bundle searching for this info we didn't found it. I don't have the bundle anymore to confirm this info was logged as JVM implementation Name

stephenc:
Is this information not already in nodes/master/system.properties?

src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
stephenc: Or if the system property is `null` then don't add the line... but whatever you do don't say `Bit size: null`

ndeloof:
indeed, but as JVM name : java.vm.name=Java HotSpot(TM) 64-Bit Server VM
Just didn't found it last time I was looking for it (so much mixed infos in support bundle ...)

jglick:
Of course it would be in `system.properties`, but you would expect to see this information in `about.md`. Which I do. So unless you are seeing something different (explain), I think this can be closed.

stephenc:
Well I'd like to see a guard against `Bit size: null` before I'm ok with it. I have no issue with adding the info to `about.md` (be nice to also add it `nodes.md` while we are at it for all the slaves) as long as it either prints `Bit size: unknown (non-sun JVM)` or doesn't print a bit size... just don't add `Bit size: null`

stephenc:
@christ66 are you planning on addressing the comments?

christ66:
Closing this pull request. This information is already given to us though the system properties. This also won't work with any other jdk implementations like openjdk, so it would be useless.

jglick:
To be clear, do you _not_ see this information in the _JVM Implementation_ section?

christ66:
@jglick I will get the output of both open and oracle jdks tomorrow. One other option we could do to also verify the bit size is to run a command like `$JAVA_HOME/java -d64 -version` and if it errors, then it means it's on a 32-bit JVM.

A quick test on my local machine:

```
[~/Jenkins@schristou88]
% java -d64 -version
Error: This Java instance does not support a 64-bit JVM.
Please install the desired version.
[~/Jenkins@schristou88]
% java -d32 -version
java version "1.7.0_76"
Java(TM) SE Runtime Environment (build 1.7.0_76-b13)
Java HotSpot(TM) Server VM (build 24.76-b04, mixed mode)
```

We can probably do a check for if the Bit Size is null, then run this as a "backup" method.

jglick:
-1 on running an external command for this; overkill, and not portable.

If `${java.vm.name}` (what is already listed in `about.md`) does not specify the bit size on some commonly used JVMs (which?), then print `${sun.arch.data.model}` if defined, else just skip it—this information is rarely relevant anyway.

christ66:
The IBM JVM is an example that does not show the bit size:

```
      - Java Runtime Specification
          + Name:    Java Platform API Specification
          + Vendor:  Sun Microsystems Inc.
          + Version: 1.7
      - JVM Specification
          + Name:    Java Virtual Machine Specification
          + Vendor:  Sun Microsystems Inc.
          + Version: 1.0
      - JVM Implementation
          + Name:    IBM J9 VM
          + Vendor:  IBM Corporation
          + Version: 2.6
```

however it is set in:

```
java.fullversion=JRE 1.7.0 IBM J9 2.6 Linux amd64-64
sun.arch.data.model=64
com.ibm.vm.bitmode=64
```

jglick:
Well, if someone is running Jenkins on J9 they probably have plenty of problems that have nothing to do with word length. We only see a handful of people on this JVM. I think we can just close this.

christ66:
Agree. Closing this issue.

PR #22, opened 2015-02-08
Add some Unix system metrics and configuration contents

ydubreuil:
@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/api/CommandOutputContent.java
+++ b/src/main/java/com/cloudbees/jenkins/support/api/CommandOutputContent.java
@@ -1,0 +1,106 @@
+package com.cloudbees.jenkins.support.api;
+
+import com.cloudbees.jenkins.support.AsyncResultCache;
+import hudson.model.Node;
+import hudson.remoting.Callable;
+import hudson.remoting.VirtualChannel;
+import jenkins.model.Jenkins;
+
+import java.io.*;
+import java.util.WeakHashMap;
+import java.util.logging.Level;
+import java.util.logging.LogRecord;
+import java.util.logging.Logger;
+
+/**
+ * Content of a command output. You can only instantiate this content with
+ * a builder method which launch the command immediately.
+ */
+public class CommandOutputContent extends StringContent {
+
+    private static final Logger LOGGER = Logger.getLogger(CommandOutputContent.class.getName());
+
+    private CommandOutputContent(String name, String value) {
+        super(name, value);
+    }
+
+    public static class CommandLauncher implements Callable<String, RuntimeException> {
+        final String[] command;
+
+        private CommandLauncher(String... command) {
+            this.command = command;
+        }
+
+        public String call() {
+            StringWriter bos = new StringWriter();
+            PrintWriter pw = new PrintWriter(bos);
+            try {
+                dumpCommandOutput(pw, command);
+            } catch (Exception e) {
+                e.printStackTrace(pw);
+            }
+            pw.flush();
+            return bos.toString();
+        }
+
+        @edu.umd.cs.findbugs.annotations.SuppressWarnings({"DM_DEFAULT_ENCODING", "OS_OPEN_STREAM"})
+        private static void dumpCommandOutput(PrintWriter writer, String... command) throws IOException {
+            InputStream is = new ProcessBuilder(command).start().getInputStream();
+            try {
+                // this is reading from the process so platform encoding is correct
+                BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(is));
+                String line;
+                while ((line = bufferedReader.readLine()) != null) {
+                    writer.println(line);
+                }
+            } finally {
+                is.close();
+            }
+        }
+    }
+
+    private static String getNodeName(Node node) {
+        return node instanceof Jenkins ? "master" : node.getNodeName();
+    }
+
+    public static CommandOutputContent runOnNode(Node node, String name, String... command) {
+        String content = "Exception occurred while retrieving command content";
+
+        VirtualChannel chan = node.getChannel();
+        if (chan == null) {
+            content = "No connection to node";
+        } else {
+            try {
+                content = chan.call(new CommandLauncher(command));
+            } catch (IOException e) {
+                final LogRecord lr = new LogRecord(Level.FINE, "Could not retrieve command content from {0}");
+                lr.setParameters(new Object[]{getNodeName(node)});
+                lr.setThrown(e);
+                LOGGER.log(lr);
+            } catch (InterruptedException e) {
+                final LogRecord lr = new LogRecord(Level.FINE, "Could not retrieve command content from {0}");
+                lr.setParameters(new Object[]{getNodeName(node)});
+                lr.setThrown(e);
+                LOGGER.log(lr);
+            }
+        }
+
+        return new CommandOutputContent(name, content);
+    }
+
+    public static CommandOutputContent runOnNodeAndCache(WeakHashMap<Node, String> cache, Node node, String name, String... command) {
+        String content = "Exception occurred while retrieving command content";
+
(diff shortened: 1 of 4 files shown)


ndeloof:
discussed about this with Yoann and made me a demo of the related Linux system stats he'd like to get in all support bundles,
so :+1:

daniel-beck:
Pretty sure this won't work on OS X, which is considered to be a Unix.

ndeloof:
@daniel-beck osx definitively is more unix-friendly than windows, but I won't consider it a Unix :-P
see hudson.tools.JDKInstaller.Platform
Anyway the target here is Linux.

daniel-beck:
@ndeloof Am I missing something here? Both `Launcher` and `SlaveComputer` consider OS X to be a Unix due to its `File.pathSeparatorChar`, which means it passes the checks in this code. Only to then fail to read from `/proc` as that does not exist there.

daniel-beck:
… which means any support bundle will contain eight files per OS X node that consist of a warning message and mostly identical 5KB stack trace.

	--- WARNING: Could not attach /proc/cpuinfo as it cannot currently be found ---

	java.io.FileNotFoundException: /proc/cpuinfo (No such file or directory)
		at java.io.FileInputStream.open(Native Method)
		at java.io.FileInputStream.<init>(FileInputStream.java:146)
		at hudson.util.IOUtils.copy(IOUtils.java:29)
		at com.cloudbees.jenkins.support.api.FileContent.writeTo(FileContent.java:63)
		at com.cloudbees.jenkins.support.SupportPlugin.writeBundle(SupportPlugin.java:342)
		at com.cloudbees.jenkins.support.SupportAction.doDownload(SupportAction.java:160)
		...

This does not look like it's deliberate, and is just asking for a bug report to be filed about it.

ndeloof:
indeed, must just skip producing content if platform does not match expectations

ydubreuil:
@daniel-beck you're right, it's better to detect the platform. It's fixed now.

I'm also limiting the dmesg output size and adding the output of 'id -a'

daniel-beck:
No longer generates useless data on OS X.

FWIW `/bin/sh -c "id -a"` would also work on OS X.

src/main/java/com/cloudbees/jenkins/support/api/CommandOutputContent.java
```diff
@@ -0,0 +1,106 @@
...
+        @edu.umd.cs.findbugs.annotations.SuppressWarnings({"DM_DEFAULT_ENCODING", "OS_OPEN_STREAM"})
+        private static void dumpCommandOutput(PrintWriter writer, String... command) throws IOException {
+            InputStream is = new ProcessBuilder(command).start().getInputStream();
+            try {
+                // this is reading from the process so platform encoding is correct
+                BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(is));
+                String line;
+                while ((line = bufferedReader.readLine()) != null) {
+                    writer.println(line);
+                }
```
jglick: Overkill. If you were starting with a `StringWriter` and an `InputStreamReader`, there is no need to look up line boundaries. A simple `IOUtils.copyStream` would suffice.

src/main/java/com/cloudbees/jenkins/support/impl/JVMProcessSystemMetricsContents.java
```diff
@@ -0,0 +1,103 @@
...
+        }
+        String name;
+        if (node instanceof Jenkins) {
+            name = "master";
+        } else {
+            name = "slave/" + node.getNodeName();
+        }
+
+        for (Map.Entry<String,String> procDescriptor : UNIX_PROC_CONTENTS.entrySet()) {
+            container.add(new FileContent("nodes/" + name + "/proc/" + procDescriptor.getValue(), new File(procDescriptor.getKey())));
```
jglick: This is already covered by another component IIRC.

src/main/java/com/cloudbees/jenkins/support/impl/SystemConfiguration.java
```diff
@@ -0,0 +1,110 @@
...
+
+    private static final Logger LOGGER = Logger.getLogger(SystemConfiguration.class.getName());
+
+    static Map<String,String> UNIX_PROC_CONTENTS;
+    static {
+        UNIX_PROC_CONTENTS = new HashMap<String,String>();
+        UNIX_PROC_CONTENTS.put("/proc/swaps", "swaps.txt");
+        UNIX_PROC_CONTENTS.put("/proc/cpuinfo", "cpuinfo.txt");
+        UNIX_PROC_CONTENTS.put("/proc/mounts", "mounts.txt");
+        UNIX_PROC_CONTENTS.put("/proc/sys/kernel/random/entropy_avail", "kernel_entropy_avail.txt");
```
jglick: Probably irrelevant after the fix of [JENKINS-20108](https://issues.jenkins-ci.org/browse/JENKINS-20108).

ydubreuil: True, system entropy is now only required to feed the initial seed

src/main/java/com/cloudbees/jenkins/support/api/CommandOutputContent.java
ydubreuil: You're right, I' simplifying the code

src/main/java/com/cloudbees/jenkins/support/impl/JVMProcessSystemMetricsContents.java
ydubreuil: I found a implementation in JDKInstaller, but nothing really reusable.

src/main/java/com/cloudbees/jenkins/support/api/CommandOutputContent.java
```diff
@@ -0,0 +1,93 @@
...
+        if (chan == null) {
+            content = "No connection to node";
+        } else {
+            try {
+                content = chan.call(new CommandLauncher(command));
+            } catch (IOException e) {
+                final LogRecord lr = new LogRecord(Level.FINE, "Could not retrieve command content from {0}");
+                lr.setParameters(new Object[]{getNodeName(node)});
+                lr.setThrown(e);
+                LOGGER.log(lr);
```
jglick: By the way you can save yourself some effort and just write

```java
LOGGER.log(Level.FINE, "Could not retrieve command content from " + getNodeName(node), e);
```

Yes this does string concatenation when `FINE` is not being logged, but that is going to be trivial compared to the work that went into the remote procedure call (or even stack trace creation), and we hope such exceptional cases are unusual to begin with. It is only worth bothering with the message format overloads of `log` for calls that are expected to be frequent.

src/main/java/com/cloudbees/jenkins/support/api/CommandOutputContent.java
```diff
@@ -0,0 +1,93 @@
+package com.cloudbees.jenkins.support.api;
```
jglick: Missing license/copyright header in this and other files.

src/main/java/com/cloudbees/jenkins/support/impl/JVMProcessSystemMetricsContents.java
jglick: I meant another `com.cloudbees.jenkins.support.api.Component`. Specifically, `FileDescriptorLimit`, which has `listAllOpenFileDescriptors`. There is no actual overlap with that component (it lists `/proc/self/fd/*` whereas this lists other things under `/proc/self/`), but they are clearly closed related, so it would seem reasonable to merge them into one `Component` to reduce the number of choices the user needs to make when picking things to include in the bundle.

jglick:
:+1: despite some minor unaddressed comments; will merge.

PR #24, opened 2015-02-25
Add networking interface information to support bundle

christ66:
Add an option to obtain information about the networking interfaces that are part of the computer.

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/NetworkInterfaces.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/NetworkInterfaces.java
@@ -37,6 +37,7 @@ import hudson.security.Permission;
 import hudson.util.HexBinaryConverter;
 import jenkins.model.Jenkins;
 import org.apache.commons.codec.binary.Hex;
+import org.codehaus.groovy.runtime.StackTraceUtils;
 
 import java.io.IOException;
 import java.io.OutputStream;
@@ -102,16 +103,15 @@ public class NetworkInterfaces extends Component {
 
     private static final class GetNetworkInterfaces implements Callable<String, RuntimeException> {
         public String call() {
-            StringWriter bos = new StringWriter();
-            PrintWriter pw = new PrintWriter(bos);
+            StringBuilder bos = new StringBuilder();
 
             try {
                 Enumeration<NetworkInterface> networkInterfaces = NetworkInterface.getNetworkInterfaces();
                 while (networkInterfaces.hasMoreElements()) {
-                    pw.println("-----------");
+                    bos.append("-----------\n");
 
                     NetworkInterface ni = networkInterfaces.nextElement();
-                    pw.println(" * Name " + ni.getDisplayName());
+                    bos.append(" * Name ").append(ni.getDisplayName()).append("\n");
 
                     byte[] hardwareAddress = ni.getHardwareAddress();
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/cloudbees/jenkins/support/impl/NetworkInterfaces.java
```diff
@@ -0,0 +1,144 @@
...
+                while (networkInterfaces.hasMoreElements()) {
+                    pw.println("-----------");
+
+                    NetworkInterface ni = networkInterfaces.nextElement();
+                    pw.println(" * Name " + ni.getDisplayName());
+
+                    byte[] hardwareAddress = ni.getHardwareAddress();
+
+                    // Do not have permissions or address does not exist
+                    if (hardwareAddress == null || hardwareAddress.length == 0)
```
jtnord: As jenkins may only be bound to the loopback I think we should not break but just not dump the hardware address (or dump something special)

src/main/java/com/cloudbees/jenkins/support/impl/NetworkInterfaces.java
```diff
@@ -0,0 +1,144 @@
...
+                        continue;
+
+                    pw.println(" ** Hardware Address - " + Util.toHexString(hardwareAddress));
+                    pw.println(" ** Index - " + ni.getIndex());
+                    pw.println(" ** Inet Address - " + ni.getInetAddresses());
```
jtnord: This is an array so prints an object :(

`** Inet Address - java.net.NetworkInterface$1checkedAddresses@2ce6ab88`

src/main/java/com/cloudbees/jenkins/support/impl/NetworkInterfaces.java
```diff
@@ -0,0 +1,144 @@
...
+        return AsyncResultCache.get(node,
+                networkInterfaceCache,
+                new GetNetworkInterfaces(),
+                "network interfaces",
+                "N/A: No connection to node, or no cache.");
+    }
+
+    private static final class GetNetworkInterfaces implements Callable<String, RuntimeException> {
+        public String call() {
+            StringWriter bos = new StringWriter();
```
jtnord: seems like an overkill when StringBuilder would work.

src/main/java/com/cloudbees/jenkins/support/impl/NetworkInterfaces.java
```diff
@@ -0,0 +1,146 @@
...
+                    byte[] hardwareAddress = ni.getHardwareAddress();
+
+                    // Do not have permissions or address does not exist
+                    if (hardwareAddress != null && hardwareAddress.length != 0)
+                        bos.append(" ** Hardware Address - ").append(Util.toHexString(hardwareAddress)).append("\n");
+
+                    bos.append(" ** Index - ").append(ni.getIndex()).append("\n");
+                    Enumeration<InetAddress> inetAddresses = ni.getInetAddresses();
+                    while (inetAddresses.hasMoreElements()) {
+                        NetworkInterface networkInterface =  networkInterfaces.nextElement();
```
jtnord: I think you mean to refer to inetAddress here.

jtnord:
:-1: also build failure needs investigating.

christ66:
@jtnord Fixed.

src/main/java/com/cloudbees/jenkins/support/impl/NetworkInterfaces.java
```diff
@@ -0,0 +1,146 @@
...
+        return "Networking Interface";
+    }
+
+    @Override
+    public void addContents(@NonNull Container result) {
+        result.add(
+                new Content("nodes/master/networkInterface.md") {
+                    @Override
+                    public void writeTo(OutputStream os) throws IOException {
+                        os.write(getNetworkInterface(Jenkins.getInstance()).getBytes("utf-8"));
```
jtnord: "utf-8" is implementation specific.  Use the mandated "UTF-8" instead.

src/main/java/com/cloudbees/jenkins/support/impl/NetworkInterfaces.java
```diff
@@ -0,0 +1,146 @@
...
+                    }
+                }
+        );
+
+        for (final Node node : Jenkins.getInstance().getNodes()) {
+            result.add(
+                    new Content("nodes/slave/" + node.getNodeName() + "/networkInterface.md") {
+                        @Override
+                        public void writeTo(OutputStream os) throws IOException {
+                            os.write(getNetworkInterface(node).getBytes("utf-8"));
```
jtnord: ditto above.

src/main/java/com/cloudbees/jenkins/support/impl/NetworkInterfaces.java
```diff
@@ -0,0 +1,146 @@
...
+        return "Networking Interface";
+    }
+
+    @Override
+    public void addContents(@NonNull Container result) {
+        result.add(
+                new Content("nodes/master/networkInterface.md") {
+                    @Override
+                    public void writeTo(OutputStream os) throws IOException {
+                        os.write(getNetworkInterface(Jenkins.getInstance()).getBytes("UTF-8"));
```
jglick: BTW there are at least two `UTF_8` constants: Apache Commons and Guava.

jglick:
:+1:

src/main/java/com/cloudbees/jenkins/support/impl/NetworkInterfaces.java
jtnord: I wouldn't use either of those and wait until JDK7 and use `StandardCharsets` which is even more efficient :-)

jtnord:
:+1:

PR #27, opened 2015-04-01
Use the CB UC whenever nectar-license is installed

jglick:
Otherwise I have seen

```
FROM jenkins:1.580.2
RUN mkdir -p /usr/share/jenkins/ref/plugins/
…
RUN curl -L $JENKINS_UC/plugins/nectar-license/6.0/nectar-license.hpi -o /usr/share/jenkins/ref/plugins/nectar-license.jpi
RUN curl -L $JENKINS_UC/plugins/nectar-rbac/4.12/nectar-rbac.hpi -o /usr/share/jenkins/ref/plugins/nectar-rbac.jpi
RUN curl -L $JENKINS_UC/plugins/nectar-vmware/4.3.2/nectar-vmware.hpi -o /usr/share/jenkins/ref/plugins/nectar-vmware.jpi
…
```

which will not work.

@reviewbybees but esp. @alecharp who seems to have worked on this recently. (A new release would be in order soon, no?)

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
@@ -703,19 +703,20 @@ public class AboutJenkins extends Component {
         @Override
         protected void printTo(PrintWriter out) throws IOException {
 
+            PluginManager pluginManager = Jenkins.getInstance().getPluginManager();
             String fullVersion = Jenkins.getVersion().toString();
             int s = fullVersion.indexOf(' ');
             if (s > 0 && fullVersion.contains("CloudBees")) {
                 out.println("FROM cloudbees/jenkins:" + fullVersion.substring(0, s));
-
-                out.println("ENV JENKINS_UC http://jenkins-updates.cloudbees.com");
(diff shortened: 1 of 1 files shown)


ndeloof:
I need to review this as well, as jenkins official docker image now only do offer LTS releases.
I plan to _not_ rely anymore on it but just fully build a Dockerfile to reproduce user installation

alecharp:
@jglick what if `nectar-license` is renamed? But I agree that is maybe a good way to see if the `Dockerfile` needs to have the CloudBees update center link.

ndeloof:
yes indeed, will integrate this in my incoming PR

jglick:
> I plan to not rely anymore on it

Sounds like a plan. Does that make this PR invalid, or do you just mean that you want to do something after it?

> what if `nectar-license` is renamed?

Plugin `shortName`s can never be changed once published.

ndeloof:
I'll write a distinct PR, but reuse the same detection technique, so I'm fine to merge
:+1:

alecharp:
Cannot be changed but could be deprecated / replace by another binary in the futur. But it's still :+1:

PR #28, opened 2015-04-21
Add file size limit to winsw logs

christ66:
Limit the winsw log files to `MAX_FILE_SIZE`. The `jenkins.err.log` file can be several MB, so we should limit the size.

@reviewbybees


stephenc:
:+1:

stephenc:
(BTW does that limit get the tail or the start?)

jglick:
:+1:

PR #29, opened 2015-04-24
Add load statistics to support bundle.

stephenc:
- Where possible we include PNG graphs for faster analysis
- Always include a CSV of the raw data

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/LoadStats.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/LoadStats.java
@@ -36,6 +36,8 @@ import hudson.model.MultiStageTimeSeries;
 import hudson.model.TimeSeries;
 import hudson.security.Permission;
 import jenkins.model.Jenkins;
+import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang.WordUtils;
 
 import javax.imageio.ImageIO;
 import java.awt.*;
@@ -224,7 +226,7 @@ public class LoadStats extends Component {
             out.println("#!/usr/bin/env gnuplot");
             out.println("set style line 101 lc rgb '#000000' lt 1 lw 1;");
             out.println("set border 3 front ls 101;");
-            out.println("set tics nomirror out scale;");
+            out.println("set tics nomirror out;");
             out.println("set format '%g';");
             out.println("set key box linestyle 101;");
             out.println("set key outside;");
@@ -234,14 +236,15 @@ public class LoadStats extends Component {
             out.println("set xdata time; ");
             out.println("set xtics rotate; ");
             out.println("set datafile sep \",\";");
-            out.println("set term png size 500,400;");
+            out.println("set term png font \"arial\" 9 size 500,400;");
             for (MultiStageTimeSeries.TimeScale scale : MultiStageTimeSeries.TimeScale.values()) {
                 String scaleName = scale.name().toLowerCase(Locale.ENGLISH);
                 out.printf("set output \"%s.png\";%n", scaleName);
                 int col = 2;
                 for (Field f : FIELDS) {
+                    String name = StringUtils.join(StringUtils.splitByCharacterTypeCamelCase(f.getName()), " ");
                     out.printf("%s \"%s.csv\" using 1:%d with lines title \"%s\"", col == 2 ? "plot" : ",", scaleName,
-                            col, f.getName());
+                            col, name.substring(0, 1).toUpperCase(Locale.ENGLISH) + name.substring(1).toLowerCase(Locale.ENGLISH));
                     col++;
                 }
                 out.println(";");
--- a/src/main/java/com/cloudbees/jenkins/support/impl/LoadStats.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/LoadStats.java
@@ -200,7 +200,7 @@ public class LoadStats extends Component {
                 maxLen = Math.max(maxLen, entry.getValue().length);
             }
             out.println();
-            for (int row = maxLen - 1, offset = 0; row >= 0; row--, offset++) {
+            for (int row = maxLen - 1, offset = 1; row >= 0; row--, offset++) {
                 out.print(dateFormat.format(new Date(time - clock * (maxLen - offset))));
                 for (float[] h : data.values()) {
                     out.print(',');
--- a/src/main/java/com/cloudbees/jenkins/support/impl/LoadStats.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/LoadStats.java
@@ -1,0 +1,155 @@
+package com.cloudbees.jenkins.support.impl;
+
+import com.cloudbees.jenkins.support.api.Component;
+import com.cloudbees.jenkins.support.api.Container;
+import com.cloudbees.jenkins.support.api.Content;
+import com.cloudbees.jenkins.support.api.PrintedContent;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import hudson.Extension;
+import hudson.model.Label;
+import hudson.model.LoadStatistics;
+import hudson.model.MultiStageTimeSeries;
+import hudson.model.TimeSeries;
+import hudson.security.Permission;
+import jenkins.model.Jenkins;
+
+import javax.imageio.ImageIO;
+import java.awt.image.BufferedImage;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.io.PrintWriter;
+import java.io.UnsupportedEncodingException;
+import java.lang.reflect.Field;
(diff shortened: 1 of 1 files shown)


src/main/java/com/cloudbees/jenkins/support/impl/LoadStats.java
```diff
@@ -0,0 +1,155 @@
+package com.cloudbees.jenkins.support.impl;
```
jtnord: OSS - doesn't this need a licence header?

stephenc: None of the other ones have one.

jtnord: You shouldn't lower high standards :-)

src/main/java/com/cloudbees/jenkins/support/impl/LoadStats.java
```diff
@@ -0,0 +1,155 @@
...
+    @NonNull
+    @Override
+    public String getDisplayName() {
+        return "Load Statistics";
+    }
+
+    @Override
+    public void addContents(@NonNull Container container) {
+        Jenkins jenkins = Jenkins.getInstance();
+        add(container, "unlabeled", jenkins.unlabeledLoad);
```
jtnord: nice lableled (sic) typo in Jenkins core ?

src/main/java/com/cloudbees/jenkins/support/impl/LoadStats.java
```diff
@@ -0,0 +1,155 @@
...
+            try {
+                add(container, String.format("label/%s", URLEncoder.encode(l.getName(), "UTF-8")), l.loadStatistics);
+            } catch (UnsupportedEncodingException e) {
+                // ignore UTF-8 is required by JLS specification
+            }
+        }
+    }
+
+    private void add(@NonNull Container container, String name, LoadStatistics stats) {
+        boolean headless = Boolean.getBoolean("java.awt.headless");
```
jtnord: this is the wrong way to check for a headless enviroment - you should check the GraphicsEnvironment.isHeadless()

src/main/java/com/cloudbees/jenkins/support/impl/LoadStats.java
```diff
@@ -0,0 +1,155 @@
...
+        for (MultiStageTimeSeries.TimeScale scale: MultiStageTimeSeries.TimeScale.values()) {
+            if (!headless) {
+                BufferedImage image = stats.createTrendChart(scale).createChart().createBufferedImage(500, 400);
+                container.add(new ImageContent(String.format("load-stats/%s/%s.png", name, scale.name().toLowerCase()),
```
jtnord: You should be able to generate graphs etc even if you are headless.

jtnord: toLowerCase() can be whacky on some locales - best to supply a Locale rather than rely on the default.

src/main/java/com/cloudbees/jenkins/support/impl/LoadStats.java
```diff
@@ -0,0 +1,180 @@
...
+            } catch (UnsupportedEncodingException e) {
+                // ignore UTF-8 is required by JLS specification
+            }
+        }
+    }
+
+    private void add(@NonNull Container container, String name, LoadStatistics stats) {
+        boolean headless = GraphicsEnvironment.isHeadless();
+        for (MultiStageTimeSeries.TimeScale scale: MultiStageTimeSeries.TimeScale.values()) {
+            if (!headless) {
```
jtnord: You should be able to generate graphs etc even if you are headless?

toLowerCase() can be whacky on some locales - best to supply a Locale rather than rely on the default.

jglick: Specify Turkish/Azeri. FindBugs will warn you about this.

src/main/java/com/cloudbees/jenkins/support/impl/LoadStats.java
```diff
@@ -0,0 +1,180 @@
...
+                BufferedImage image = stats.createTrendChart(scale).createChart().createBufferedImage(500, 400);
+                container.add(new ImageContent(String.format("load-stats/%s/%s.png", name, scale.name().toLowerCase()),
+                        image));
+            }
+            container.add(new CsvContent(String.format("load-stats/%s/%s.csv", name, scale.name().toLowerCase()), stats, scale));
```
jglick: (`toLowerCase` call here)

src/main/java/com/cloudbees/jenkins/support/impl/LoadStats.java
stephenc: @jtnord 

> You should be able to generate graphs etc even if you are headless?

Some JDKs cannot generate graphs if run headless (due to missing fonts), so I would rather just avoid them if in doubt.

daniel-beck: [JENKINS-25261](https://issues.jenkins-ci.org/browse/JENKINS-25261) has some related discussion. At the same time, [RPM installs](https://github.com/jenkinsci/packaging/blob/master/rpm/build/SOURCES/jenkins.sysconfig.in#L37) default to headless (but can generate graphs if there are fonts).

stephenc: For me, the graphs are a nice to have, it's the raw data that matters. What
I may do is output a gnu plot script to generate the PNGs if running
headless... Though that doesn't add much value as people consuming support
bundles should be able to have that script on standby... It would compress
well and restore value to sysadmins trying to self-diagnose

On Saturday, April 25, 2015, Daniel Beck <notifications@github.com> wrote:

> In src/main/java/com/cloudbees/jenkins/support/impl/LoadStats.java
> <https://github.com/jenkinsci/support-core-plugin/pull/29#discussion_r29099473>
> :
>
> > +        Jenkins jenkins = Jenkins.getInstance();
> > +        add(container, "unlabeled", jenkins.unlabeledLoad);
> > +        add(container, "overall", jenkins.overallLoad);
> > +        for (Label l: jenkins.getLabels()) {
> > +            try {
> > +                add(container, String.format("label/%s", URLEncoder.encode(l.getName(), "UTF-8")), l.loadStatistics);
> > +            } catch (UnsupportedEncodingException e) {
> > +                // ignore UTF-8 is required by JLS specification
> > +            }
> > +        }
> > +    }
> > +
> > +    private void add(@NonNull Container container, String name, LoadStatistics stats) {
> > +        boolean headless = GraphicsEnvironment.isHeadless();
> > +        for (MultiStageTimeSeries.TimeScale scale: MultiStageTimeSeries.TimeScale.values()) {
> > +            if (!headless) {
>
> JENKINS-25261 <https://issues.jenkins-ci.org/browse/JENKINS-25261> has
> some related discussion. At the same time, RPM installs
> <https://github.com/jenkinsci/packaging/blob/master/rpm/build/SOURCES/jenkins.sysconfig.in#L37>
> default to headless (but can generate graphs if there are fonts).
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jenkinsci/support-core-plugin/pull/29/files#r29099473>
> .
>


-- 
Sent from my phone

src/main/java/com/cloudbees/jenkins/support/impl/LoadStats.java
```diff
@@ -0,0 +1,181 @@
...
+import java.util.Date;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Set;
+import java.util.TimeZone;
+import java.util.TreeMap;
+
+/**
+ * @since
```
jtnord: A simple bit that says this is the Jenkins load statistics would be nice :)

stephenc:
Here is a graph if you are not headless:
sec10-orig

Here is a graph generated by gnuplot if you are headless:
sec10

jtnord:
the times on the plot seem to be out by 10 seconds - is this an off by one in your calculations of the time in the CSV?

Assuming that the issue with the plot missing the hours is to do with gnuplot as the `dataformat` looks ok to me.

Other than that looks good

stephenc:
@jtnord thanks for the off-by-one error... though the whole load stats graphs are potentially mis-aligned in core anyway as the `float[]` arrays can be updated while generating the graph and result in a skew for one series relative to the other, so being out by 10 seconds is largely going to be a non-issue... but for the per-minute and per-hour graphs it could get more confusing alright!

I think we are ready now, if we get the votes ;-)

jtnord:
:+1:

PR #30, opened 2015-04-30
Add root Jenkins config.xml to support bundle

joshshinn:
@reviewbybees


jglick:
:-1:, this could contain private information, and could be huge if (prior to 1.607 IIRC) there are a lot of slaves.

Any basic information is already in `about.md`.

christ66:
What if we disabled it by default? I still think having the option to download this via support bundle is useful.

jglick:
Is it not much more common to want job `config.xml`s? I mean, you can always request particular files you know you are interested in.

stephenc:
:-1: too much of a security risk, better to just ask for the file if you need it

stephenc:
Closing as maintainer

PR #32, opened 2015-05-09
Prevent NPE in unit test

christ66:
Fix a NPE when running unit tests at the start of the test:

```
Caused by: java.lang.NullPointerException
	at com.cloudbees.jenkins.support.SupportPlugin.loadConfig(SupportPlugin.java:384)
``` 

@reviewbybees


tfennelly:
:+1:

oleg-nenashev:
:+1: for tests. Does it make sense to add some logging if an issue somehow appears in the production?

PR #33, opened 2015-05-18
[JENKINS-24671] Add username and referrer to slow-requests

christ66:
This pull request adds the username of the person experiencing slow http requests. It also adds the referrer header in the event a plugin invokes the api and is unaware of it.

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/slowrequest/InflightRequest.java
+++ b/src/main/java/com/cloudbees/jenkins/support/slowrequest/InflightRequest.java
@@ -1,5 +1,9 @@
 package com.cloudbees.jenkins.support.slowrequest;
 
+import com.google.common.net.HttpHeaders;
+import jenkins.model.Jenkins;
+import org.kohsuke.stapler.StaplerRequest;
+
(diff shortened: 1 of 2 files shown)


alecharp:
:+1:

src/main/java/com/cloudbees/jenkins/support/slowrequest/InflightRequest.java
```diff
@@ -36,14 +40,27 @@
...
+     * Referer link to track any redirect urls.
+     */
+    final String referer;
 
     InflightRequest(HttpServletRequest req) {
         String query = req.getQueryString();
         url = req.getRequestURL() + (query == null ? "" : "?" + query);
         startTime = System.currentTimeMillis();
+        userName = Jenkins.getAuthentication().getName();
+        referer = ((StaplerRequest)req).getReferer();
```
oleg-nenashev: Unsafe cast. I would recommend to check the class and to default the field to something if it is not <code> StaplerRequest</code>

stephenc:
:+1:

oleg-nenashev:
:+1:

PR #36, opened 2015-06-16
Add null check before parsing agent header.

christ66:
@reviewbybees


daniel-beck:
:+1:

oleg-nenashev:
A ternary operator could simplify the code, but :+1: for the fix

oleg-nenashev:
<code>userAgent = agentHeader != null ? UADetectorServiceFactory.getResourceModuleParser().parse(agentHeader) : null;</code>

PR #38, opened 2015-06-23
[JENKINS-29034] - Modify filename to include date and time

recena:
https://issues.jenkins-ci.org/browse/JENKINS-29034

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/SupportAction.java
+++ b/src/main/java/com/cloudbees/jenkins/support/SupportAction.java
@@ -152,13 +152,14 @@ public class SupportAction implements RootAction {
         if (supportPlugin != null) {
             SupportProvider supportProvider = supportPlugin.getSupportProvider();
             if (supportProvider != null) {
-                SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd_HH.mm.ss");
-                dateFormat.setTimeZone(TimeZone.getTimeZone("GMT"));
                 // let the provider name it
-                filename = supportProvider.getName() + "_" + dateFormat.format(new Date());
+                filename = supportProvider.getName();
             }
         }
-        rsp.addHeader("Content-Disposition", "inline; filename=" + filename + ".zip;");
+
+        SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd_HH.mm.ss");
+        dateFormat.setTimeZone(TimeZone.getTimeZone("GMT"));
+        rsp.addHeader("Content-Disposition", "inline; filename=" + filename + "_" + dateFormat.format(new Date()) + ".zip;");
         final ServletOutputStream servletOutputStream = rsp.getOutputStream();
         try {
             SupportPlugin.setRequesterAuthentication(Jenkins.getAuthentication());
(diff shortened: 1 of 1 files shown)


recena:
@reviewbybees Could you review this?, specially @stephenc @aheritier @ndeloof and others from Support Engineering Team.

Read this to get more information https://github.com/cloudbees/cloudbees-support-plugin/pull/5

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -146,8 +152,10 @@ public void doDownload(StaplerRequest req, StaplerResponse rsp) throws ServletEx
         if (supportPlugin != null) {
             SupportProvider supportProvider = supportPlugin.getSupportProvider();
             if (supportProvider != null) {
+                SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd_HH.mm.ss");
+                dateFormat.setTimeZone(TimeZone.getTimeZone("GMT"));
                 // let the provider name it
-                filename = supportProvider.getName();
+                filename = supportProvider.getName() + "_" + dateFormat.format(new Date());
```
stephenc: You need the date always, not just for those coming from the supportProvider. Do the date injection in the addHeader line not here

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -146,8 +152,10 @@ public void doDownload(StaplerRequest req, StaplerResponse rsp) throws ServletEx
         if (supportPlugin != null) {
             SupportProvider supportProvider = supportPlugin.getSupportProvider();
             if (supportProvider != null) {
+                SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd_HH.mm.ss");
```
christ66: Creating a new `SimpleDateFormat` can be an expensive operation especially every time the `doDownload` link is clicked. I would recommend using a `static final` field since the format for the date/time never changes.

ndeloof: but SimpleDateFormat is not thread safe, so general recommendation is to have one and clone it for use. Anyway, as this isn't an intensively used URL all this would be premature optimization imho

stephenc: Agree with @ndeloof 

1. This action is not heavily used, we can incur the cost of creating a new date format every time it is invoked. Users typically only access this link once or twice when they have an issue... and they are not performance sensitive compared to the time they will spend downloading the bundle

2. The instance is not thread safe and is advertised as not thread safe, so if used form a static field you would need to add locking and synchronization which could potentially interfere with the issue the customer is encountering and prevent the generation of a bundle.

recena: @christ66 This [part of the plugin](https://github.com/jenkinsci/support-core-plugin/blob/master/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java#L622) is executed periodically and we use `SimpleDateFormat`.

recena:
@stephenc Could you review my last commit? Thanks.

christ66:
:+1:

stephenc:
:+1:

recena:
@stephenc @christ66 Thanks so much for your feedback.

@stephenc Could you merge it? Thanks in advance.

recena:
@christ66 Thanks so much.

PR #39, opened 2015-06-30
Ignore build queue unit test

christ66:
Currently this unit test has a race condition when performing a release and is currently preventing a new release. Unit tests work when running with `mvn test`, and `mvn install`, however fail consistently when a release is performed.

@reviewbybees

### DIFF
--- a/src/test/java/com/cloudbees/jenkins/support/impl/BuildQueueTest.java
+++ b/src/test/java/com/cloudbees/jenkins/support/impl/BuildQueueTest.java
@@ -30,6 +30,7 @@ import hudson.model.FreeStyleBuild;
 import hudson.model.FreeStyleProject;
 import hudson.model.labels.LabelAtom;
 import hudson.model.queue.QueueTaskFuture;
+import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.jvnet.hudson.test.JenkinsRule;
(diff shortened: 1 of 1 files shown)


christ66:
@jglick noticed this first: https://github.com/jenkinsci/support-core-plugin/pull/14#discussion_r29801137.

src/test/java/com/cloudbees/jenkins/support/impl/BuildQueueTest.java
```diff
@@ -51,6 +52,7 @@
 public class BuildQueueTest {
   @Rule public JenkinsRule j = new JenkinsRule();
 
+  @Ignore
```
jtnord: always add a reason when you @Ignore
e.g. `@Ignore("Fails under release and I need to make one even though this test fails and is telling me something is broken, I think I know better");`

christ66: Ah, thanks! Adding comment.

jtnord:
:bug: on the lack of comment for @Ignore.
I would like to see the failure commented somewhere (either in this PR or a gist link).

If there is a race condition would it not be better to fix the code that causes it or fix the unit test?  This test presumably has been working before now, and as such the fact that something is failing should be a blocker for release.

stephenc:
I suspect the race condition will be resolved once core is 1.607+

stephenc:
:bug: for `@Ignore` without comment

stephenc:
:bee:

stephenc:
would be even better if you had a `// TODO enable once Jenkins 1.607+` but I'll not hold up for that request

oleg-nenashev:
:bee:

jtnord:
:bee:

PR #41, opened 2015-10-01
Capture the full stacktrace for any deadlocked threads

stephenc:
Too often the actual cause of the deadlock is outside the first ThreadInfo.MAX_FRAMES (i.e. 8) thread frames.
When that happens the actual deadlock information is worse than useless.

This change uses the same formatter as the full thread dump which should result in more useful information

@reviewbybees


oleg-nenashev:
It's hard to review the code due to the method relocation.
BTW :bee:

michaelneale:
:bee:  pending tests passing, this is very much needed.

stephenc:
Not sure why that test failed, it was working for me...

oleg-nenashev:
:bee:

stephenc:
:bee: @reviewbybees done

PR #42, opened 2015-10-01
Add myself to developers list

christ66:
I would like to be added as a maintainer/developer to the support core plugin.


christ66:
@stephenc :smiley: ?

PR #45, opened 2015-11-10
[JENKINS-28216] Obtain slave logs using async callable.

christ66:
Previous implementation used a `Channel.call` which performs a synchronous callable. Instead we can use the `Channel.callAsync` in order to prevent the channel from blocking.

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/JenkinsLogs.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/JenkinsLogs.java
@@ -19,8 +19,9 @@ import hudson.logging.LogRecorder;
 import hudson.model.Computer;
 import hudson.model.Node;
 import hudson.model.PeriodicWork;
+import hudson.remoting.*;
 import hudson.remoting.Callable;
-import hudson.remoting.VirtualChannel;
+import hudson.remoting.Future;
 import hudson.security.Permission;
 import hudson.slaves.Cloud;
 import hudson.slaves.SlaveComputer;
@@ -45,11 +46,7 @@ import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
+import java.util.concurrent.*;
 import java.util.logging.Formatter;
 import java.util.logging.Handler;
 import java.util.logging.Level;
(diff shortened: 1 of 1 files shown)


stephenc:
:bee: AIUI

src/main/java/com/cloudbees/jenkins/support/impl/JenkinsLogs.java
```diff
@@ -112,9 +109,18 @@ protected void printTo(PrintWriter out) throws IOException {
                                     out.println("N/A");
                                 } else {
                                     try {
-                                        List<LogRecord> records = needHack
-                                                ? SlaveLogFetcher.getLogRecords(computer)
-                                                : computer.getLogRecords();
+                                        List<LogRecord> records = null;
+                                        if (needHack) {
+                                            if (computer.getChannel() != null) {
```
jtnord: and if we have a computer without a channel - no logging?

src/main/java/com/cloudbees/jenkins/support/impl/JenkinsLogs.java
```diff
@@ -112,9 +109,18 @@ protected void printTo(PrintWriter out) throws IOException {
...
+                                                Future<List<LogRecord>> future = SlaveLogFetcher.getLogRecords(computer.getChannel());
+                                                records = future.get(REMOTE_OPERATION_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+                                            }
+                                        }
+
+                                        if (records == null) {
```
jtnord: so recorderd could be null as the computer has no channel - so this will now be called - whereas before it was not.  
I can;t tell if this is intentional or not.

src/main/java/com/cloudbees/jenkins/support/impl/JenkinsLogs.java
stephenc: :bug: race condition here... the channel could go null before the next line, you need to capture it in a local variable before the null check

stephenc:
:bug: NPE if the slave goes off-line during the bundle capture

src/main/java/com/cloudbees/jenkins/support/impl/JenkinsLogs.java
christ66: Correct. The initial intention with this line was to "guard" against the potential that channel _could_ be null in which case the log records won't exist. It will then try to make an attempt to grab the `computer` version of `getLogRecords` which from what I can see does not return null, but an empty list. If needed I can write the for-loop in the same way as the initial implementation.

christ66:
@stephenc @jtnord fixed the NPE issue.

stephenc:
:bee:

src/main/java/com/cloudbees/jenkins/support/impl/JenkinsLogs.java
```diff
@@ -458,6 +465,15 @@ public static boolean isRequired() {
             }
         }
 
+        public static Future<List<LogRecord>> getLogRecords(@NonNull VirtualChannel channel) throws IOException {
+            return channel.callAsync(new SlaveLogFetcher());
```
kzantow: Does this make some ThreadLocal data available based on the VirtualChannel ?

jtnord:
:bee:

src/main/java/com/cloudbees/jenkins/support/impl/JenkinsLogs.java
christ66: Some data is: https://github.com/jenkinsci/remoting/blob/master/src/main/java/hudson/remoting/Channel.java#L157.

christ66:
@reviewbybees done.

PR #46, opened 2015-11-10
[JENKINS-31365] Display blocked + waiting time per thread.

christ66:
[JENKINS-31365](https://issues.jenkins-ci.org/browse/JENKINS-31365)

This PR adds to the thread-dump.txt file the blocked time + waiting time for each thread. In order for us to calculate this time we need to enable `ThreadContentionMonitoring` on the instance, so once the plugin is started, it will enable this and any _new_ threads will provide the contention information.

Example output:

``` java
"Computer.threadPoolForRemoting [#2]" id=80 (0x50) state=TIMED_WAITING cpu=68% time=13972ms
    - waiting on <0x49bd11ee> (a java.util.concurrent.SynchronousQueue$TransferStack)
    - locked <0x49bd11ee> (a java.util.concurrent.SynchronousQueue$TransferStack)
    at sun.misc.Unsafe.park(Native Method)
    at java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:215)
    at java.util.concurrent.SynchronousQueue$TransferStack.awaitFulfill(SynchronousQueue.java:460)
    at java.util.concurrent.SynchronousQueue$TransferStack.transfer(SynchronousQueue.java:362)
    at java.util.concurrent.SynchronousQueue.poll(SynchronousQueue.java:941)
    at java.util.concurrent.ThreadPoolExecutor.getTask(ThreadPoolExecutor.java:1066)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1127)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
```

@reviewbybees @jtnord

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/ThreadDumps.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/ThreadDumps.java
@@ -273,12 +273,16 @@ public class ThreadDumps extends Component {
             x.printStackTrace(writer);
             cpuPercentage = 0;
         }
-        writer.printf("\"%s\" id=%d (0x%x) state=%s cpu=%d%%",
+
+        long time = t.getBlockedTime() + t.getWaitedTime();
+
+        writer.printf("\"%s\" id=%d (0x%x) state=%s cpu=%d%% time=%dms",
                 t.getThreadName(),
                 t.getThreadId(),
                 t.getThreadId(),
                 t.getThreadState(),
-                cpuPercentage);
+                cpuPercentage,
+                time);
         final LockInfo lock = t.getLockInfo();
         if (lock != null && t.getThreadState() != Thread.State.BLOCKED) {
             writer.printf("\n    - waiting on <0x%08x> (a %s)",
--- a/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
+++ b/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
@@ -74,6 +74,8 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.io.PrintWriter;
 import java.io.StringWriter;
+import java.lang.management.ManagementFactory;
+import java.lang.management.ThreadMXBean;
 import java.text.MessageFormat;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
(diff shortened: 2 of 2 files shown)


src/main/java/com/cloudbees/jenkins/support/impl/ThreadDumps.java
```diff
@@ -273,12 +273,16 @@ public static void printThreadInfo(PrintWriter writer, ThreadInfo t, ThreadMXBea
             x.printStackTrace(writer);
             cpuPercentage = 0;
         }
-        writer.printf("\"%s\" id=%d (0x%x) state=%s cpu=%d%%",
+
+        long time = t.getBlockedTime() + t.getWaitedTime();
```
stephenc: what happens if thread contention monitoring is not supported?

jtnord: this value is pretty useless.   Object.wait() will show up here - and that is a normal thing.  Now being blocked is abnormal and is what we really want to know - ie how much time was spent when this thread should have run but could not due to contention.

src/main/java/com/cloudbees/jenkins/support/impl/ThreadDumps.java
```diff
@@ -273,12 +273,16 @@ public static void printThreadInfo(PrintWriter writer, ThreadInfo t, ThreadMXBea
...
+
+        writer.printf("\"%s\" id=%d (0x%x) state=%s cpu=%d%% time=%dms",
                 t.getThreadName(),
                 t.getThreadId(),
                 t.getThreadId(),
                 t.getThreadState(),
-                cpuPercentage);
+                cpuPercentage,
```
jtnord: cpuPercentage is a useless metric - it is the metric of `(now - t1)/uptime)` where `t1` is the time at which the thread was started (so all threads started with Jenkins will show as 100%, even if they have been completely idle for the whole period.

jtnord:
:bug: as these metrics are as much use as a chocolate fireguard for any meaningful analysis.

batmat:
@christ66 do you plan to address the comments, or should we finally close this?

stephenc:
@batmat given @chist66 has not responded in 4 months, perhaps we should just close

jglick:
I am not familiar with the details here, but since the reporter of the issue does not agree with the approach taken I think it is safe to close.

PR #47, opened 2015-11-11
[JENKINS-30117] Support bundle blocks with lots of data

christ66:
The format method should be thread safe as it could be invoked from
multiple threads. Instead of using a MessageFormatter which is not thread safe we
use the FastDateFormatter, and a StringBuffer both of which are thread
safe.

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/SupportLogFormatter.java
+++ b/src/main/java/com/cloudbees/jenkins/support/SupportLogFormatter.java
@@ -48,12 +48,11 @@ public class SupportLogFormatter extends Formatter {
             justification = "The exception wasn't thrown on our stack frame"
     )
     public String format(LogRecord record) {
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
:ant: <code>StringBuffer</code> is not required here from what I see. You use it in a method internally => it operates within one thread => It makes sense to replace it to <code>StringBuilder</code> for the sake of the better performance.

rsandell:
:bee:

christ66:
@oleg-nenashev I think you're right. It should be using `StringBuilder`.

christ66:
@oleg-nenashev fixed.

stephenc:
:bee:

oleg-nenashev:
:bee:

jtnord:
:bee:

PR #49, opened 2015-12-16
Minor fixes/improvements in the plugin build

aheritier:


aheritier:
@reviewbybees cc @christ66  @stephenc

pom.xml
```diff
@@ -76,6 +76,7 @@
     <project.build.outputEncoding>UTF-8</project.build.outputEncoding>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <maven.findbugs.failure.strict>true</maven.findbugs.failure.strict>
+    <maven-release-plugin.version>2.5.3</maven-release-plugin.version>
```
stephenc: This property anti-pattern must die IMHO

stephenc: As in, if it is used once and only once, don't use a property. Only use a property for cases where you have to keep at least two versions in sync *in the same pom*

aheritier: I completely disagree :-) Especially for such plugin properties. It allows to change the version of a plugin pre-defined in a parent pom with one line instead of 5 ?! For me the property is a way to make easily configurable stuffs using inheritance. It is also really useful for tests because locally you can just try another version of a plugin (or another config) with just a -DmyProperty=XXX

stephenc: @aheritier so you like that the pom in the tag and deployed to the repository is not actually the one that the plugin was built with?

All I need to do is remember how to quote for `mvn release:prepare release:perform -Darguments=-Dmaven-compiler-plugin.version=x.y` and now there is no record of the actual configuration that was built... way to go for reproducible builds

It's an anti-pattern... I'm not saying it doesn't have its uses... but I generally find that unless you have a very specific use case or a need to sync two or more versions in the same pom, use of a property causes more problems than the three extra lines of xml that it removes... plus it hides the convention of the pom with a convention of the project which hinders new developers

aheritier: It is always the same debate of "developers confort" vs "reproducibility/rules/.."

```
mvn release:prepare release:perform -Darguments=-Dmaven-compiler-plugin.version=x.y
```

You can always do many crappy stuffs even without this with `-Darguments=` :-)

I'm always supporting all efforts to enforce the quality of builds but it mustn't be again the maintainability and the developer efficiency. The root issue is that we don't publish the flat pom when we do a release ...

For me having a corporate/organisational pom to add some additional configurations/conventions is the way to go and it must be as possible extensible/configurable without being too verbose.

But I agree with you that we must follow as possible the default properties and settings which are known by maven users and plugins. For example in our parent pom ( https://github.com/jenkinsci/pom/blob/master/pom.xml#L586 ) hardcoding the java version at the compiler plugin level is a bad thing. We should override the default well known properties `maven.compiler.target` and `maven.compiler.source`.

oleg-nenashev:
:bee:

pom.xml
oleg-nenashev: I agree with @aheritier BTW. If a developer feels that version needs to be configurable, the approach is right

stephenc: Using the property *hides* the version change... I still am of the opinion that **where a version number is used is one place only** do not use a property

christ66:
@aheritier Should we merge this, or wait until https://github.com/jenkinsci/support-core-plugin/pull/53/files is finished and use that pom instead?

aheritier:
Let's close this one. It will be fixed by #53

PR #50, opened 2015-12-30
Add argument to increase max rotate log size.

christ66:
Sometimes it is necessary to increase the max log rotation size to maintain the history of logs for a few days. This allows you to keep a max rotation size to keep.

@reviewbybees


abayer:
:bee: LGTM

joshshinn:
:bee:

PR #52, opened 2016-01-13
[FIX JENKINS-32401] Record log levels

christ66:
Sometimes a logger could be causing performance issues because it is far too verbose. This will display which loggers are currently enabled, and what the log level is for each logger. 

@reviewbybees @daniel-beck


aheritier:
:bee:

daniel-beck:
:bee:

PR #40, opened 2015-07-15
[WiP] Enable FindBugs for the project

oleg-nenashev:

PR #53, opened 2016-01-14
[JENKINS-33668] Adapt to new plugin parent POM and upgrade baseline to 1.580.1

andresrc:
See [JENKINS-33668](https://issues.jenkins-ci.org/browse/JENKINS-33380) Adapt to new parent point and upgrade baseline to 1.580.1

* [x] Parent POM 2.3
* [x] Baseline 1.580.1
* [x] Adapt to new remoting interfaces.
* [x] Fix javadocs when building with Java 8.
* [x] Jelly XSS PI
* [x] Adapt to JTH 2.1
* [x] Additional findbugs fixes needed because of the baseline change.
* [x] Check build against 1.642.2 (PCT scenario).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,7 +29,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>1.554.1</version>
+    <version>2.3</version>
   </parent>
 
   <artifactId>support-core</artifactId>
@@ -60,10 +60,6 @@
     </developer>
   </developers>
 
-  <prerequisites>
-    <maven>2.2.1</maven>
-  </prerequisites>
-
   <scm>
     <connection>scm:git:git://github.com/jenkinsci/support-core-plugin.git</connection>
     <developerConnection>scm:git:git@github.com:jenkinsci/support-core-plugin.git</developerConnection>
@@ -72,11 +68,8 @@
   </scm>
 
   <properties>
-    <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-    <project.build.outputEncoding>UTF-8</project.build.outputEncoding>
-    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <findbugs-maven-plugin.version>3.0.3</findbugs-maven-plugin.version>
-    <findbugs.failOnError>true</findbugs.failOnError>
+    <jenkins.version>1.580.1</jenkins.version>
+    <java.level>6</java.level>
   </properties>
 
   <repositories>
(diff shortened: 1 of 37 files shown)


pom.xml
```diff
@@ -72,53 +68,45 @@
   </scm>
 
   <properties>
-    <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-    <project.build.outputEncoding>UTF-8</project.build.outputEncoding>
-    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <maven.findbugs.failure.strict>true</maven.findbugs.failure.strict>
+    <jenkins.version>1.554.1</jenkins.version>
```
jglick: I would not advise even trying to use the new parent with anything older than 1.580.1; to start with, we would not be able to use `jenkins-test-harness` 2.x.

That said, this would be a fine time to increment the dependency IMHO. 1.554.x is now quite old.

pom.xml
```diff
@@ -72,53 +68,45 @@
...
+    <!-- TODO check if we'd better add the PI. -->
+    <jelly.requirePI>false</jelly.requirePI>
```
jglick: Yes you should add the PI. The rule is that all `*.jelly` should explicitly say they escape output by default; where you have `something.jelly`

```xml
${something}
```

and `something` might _legitimately_ include HTML tags you want to render raw without risk of XSS (for example because it is coming from `something.properties`, or is a Java getter where the value is known not be “tainted” by user input), you can use

```xml
<j:out value="${something}"/>
```

By updating to a new parent POM which uses a new version of the HPI plugin that includes this `InjectedTest`, you confirm that you have reviewed all your views for XSS vulnerabilities.

src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
```diff
@@ -239,7 +240,7 @@ public GetSlaveDigest(FilePath rootPath) {
         public String call() {
             StringBuilder result = new StringBuilder();
             final File rootPath = new File(this.rootPathName);
-            for (File file : rootPath.listFiles()) {
+            for (File file : Preconditions.checkNotNull(rootPath.listFiles())) {
```
jglick: Throwing an `IllegalArgumentException` here is not really an improvement over a `NullPointerException` I think.

src/main/java/com/cloudbees/jenkins/support/impl/FileDescriptorLimit.java
```diff
@@ -144,6 +145,7 @@ public String call() {
     /**
      * * Using OperatingSystemMXBean, we can obtain the total number of open file descriptors.
      */
+    @IgnoreJRERequirement
```
jglick: Ought to be accompanied by a Java version check in `GetUlimit`, or at least catching `LinkageError` and reporting it gracefully.

src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
jglick: (ditto other places where `File.listFiles` or `File.list` is called)

src/main/java/com/cloudbees/jenkins/support/timer/FileListCapComponent.java
```diff
@@ -34,11 +35,11 @@ public void addContents(@NonNull Container container, FileListCap fileListCap) {
...
             // the FileListCap to delete files. So we lock it.
 
-            File[] files = fileListCap.getFolder().listFiles(new FilenameFilter() {
+            File[] files = Preconditions.checkNotNull(fileListCap.getFolder().listFiles(new FilenameFilter() {
                 public boolean accept(File dir, String name) {
                     return name.endsWith(".txt");
                 }
-            });
+            }));
             for (File f : Util.fixNull(Arrays.asList(files))) {
```
jglick: Probably the call to `Util.fixNull` is bogus here. The null would be the return value of `listFiles`, not an _element_ of the return value.

jglick:
Get #40 merged first (@oleg-nenashev ping), which should take care of the FB warnings, so you can focus on the POM.

andresrc:
Sure, I won't fail the build on findbugs error with a TODO to remove the switch when https://github.com/jenkinsci/support-core-plugin/pull/40 is merged.

andresrc:
The CI build error is because of findbugs. Expected. The build is not failed but marked unstable.

andresrc:
@reviewbybees

andresrc:
As before, failing CI is findbugs.

christ66:
@andresrc please rebase and there should be no more findbugs errors.

andresrc:
rebased

@reviewbybees

christ66:
:bee:

andresrc:
@reviewbybees done

PR #55, opened 2016-01-26
[JENKINS-32622] Switch to a ThreadLocal DateFormatter

jtnord:
Attempt to avoid a deadlock observed in remoting by not causing any remote
classes to be loaded when we are logging.

@reviewbybees 
[JENKINS-32622](https://issues.jenkins-ci.org/browse/JENKINS-32622)

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/SupportLogFormatter.java
+++ b/src/main/java/com/cloudbees/jenkins/support/SupportLogFormatter.java
@@ -24,12 +24,20 @@
 
 package com.cloudbees.jenkins.support;
 
+/***********************************************
+ * DO NOT INCLUDE ANY NON JDK CLASSES IN HERE.
+ * IT CAN DEADLOCK REMOTING - SEE JENKINS-32622
+ ***********************************************/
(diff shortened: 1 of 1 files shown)


christ66:
:bee:

oleg-nenashev:
:bee: , but I would like to see a Unit test, which checks dependencies of basic support core classes. Otherwise we will easily get into this issue again

jtnord:
@oleg-nenashev we can't prevent external dependencies here - we need to loading various Jenkins classes.  The only place we need to be strict is in the logging side IIUC - and short of grepping the source code or parsing bytecode I'm not sure what we could do to write a unit test that would actually be of benifit - after all there is nothing stopping someone adding another LogFOrmatter that would not be covered by this test.

jtnord:
@reviewbybees done

jtnord:
@reviewbybees again :-)
@oleg-nenashev does this comment address your primary concern?

andresrc:
:bee:

mslusarczyk:
:bee:

PR #56, opened 2016-02-04
Prevent null pointer exception reporting log level

MarkEWaite:
Null pointer exception is now reported in my jenkins.log.  This should prevent it.

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/LoggerManager.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/LoggerManager.java
@@ -48,9 +48,11 @@ public class LoggerManager extends Component {
                 Enumeration<String> loggerNames = logManager.getLoggerNames();
                 while (loggerNames.hasMoreElements()) {
                     String loggerName =  loggerNames.nextElement();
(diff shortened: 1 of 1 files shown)


src/main/java/com/cloudbees/jenkins/support/impl/LoggerManager.java
```diff
@@ -48,9 +48,11 @@ protected void printTo(PrintWriter out) throws IOException {
                 Enumeration<String> loggerNames = logManager.getLoggerNames();
                 while (loggerNames.hasMoreElements()) {
                     String loggerName =  loggerNames.nextElement();
-                    Level level = logManager.getLogger(loggerName).getLevel();
-                    if (level != null) {
-                        out.println(loggerName + " - " + level);
+                    if (logManager.getLogger(loggerName) != null) {
+                        Level level = logManager.getLogger(loggerName).getLevel();
```
christ66: Would it be possible to make `logManager.getLogger(loggerName)` a variable?

christ66:
Thank you @MarkEWaite for this fix!

PR #59, opened 2016-04-27
JENKINS-34462 - Support core plugin needs to obtain the status of whether CSRF protection is enabled

Evildethow:
@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
@@ -534,7 +534,6 @@ public class AboutJenkins extends Component {
             }
             final JenkinsLocationConfiguration jlc = JenkinsLocationConfiguration.get();
             out.println("  * Url:     " + (jlc != null ? jlc.getUrl() : "No JenkinsLocationConfiguration available"));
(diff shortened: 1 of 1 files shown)


Dohbedoh:
:bee:

mslusarczyk:
🐝 
what do you think about putting it in 'Important configuration' section ?

christ66:
:bug: as this could lead to a NPE (http://javadoc.jenkins-ci.org/jenkins/model/JenkinsLocationConfiguration.html#get()). Instead I recommend doing `Helper.getActiveInstance()` which never returns null and will instead throw a RTE (https://github.com/jenkinsci/support-core-plugin/blob/master/src/main/java/com/cloudbees/jenkins/support/util/Helper.java#L20-L21)

I agree with @mslusarczyk that this is a security configuration, so it should be in the important configuration section.

Evildethow:
@christ66 @mslusarczyk Agreed.

Evildethow:
@christ66 @mslusarczyk Code review changes made. Let me know if you are happy and ill rebase to clean up the extra commit.

mslusarczyk:
lgtm 🐝

christ66:
:bee:

PR #60, opened 2016-05-09
[FIXED JENKINS-26409] If the stream is terminated early log this info…

christ66:
…rmation in FINE level.


christ66:
Ops, forgot @reviewbybees and @fbelzunc who made the initial fix.

PR #61, opened 2016-05-09
[FIXED JENKINS-31952] Change dockerfile from location.

christ66:


christ66:
Ops, forgot @reviewbybees

mslusarczyk:
🐝

andresrc:
:bee:

aheritier:
hummm that's crappy/hardcoded :( We couldn't have an extension point to allow vendors to redefine the image coordinate to use ?
Otherwise :bee: better than nothing

christ66:
@aheritier made it a system property.

aheritier:
thanks @christ66 
But a test is failing thus 🐛

batmat:
I think a test would be good to check the behaviour, the code is simple, but so the test would be very easy too.

PR #62, opened 2016-05-10
Separate slave logs, and master logs.

christ66:
This should help improve the performance of generating a support bundle and the size if you do not require logs from each individual slave. Some Jenkins masters have a few hundred slaves, so generating a support bundle might take a long time. Also some of the machines might have logs which might be several MB in size.

This pull request disables obtaining slave logs by default. Unless the user(s) have had issues with the slave machines, the log information for each individual slave is most likely is not required.

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/JenkinsLogs.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/JenkinsLogs.java
@@ -60,9 +60,9 @@ import org.jenkinsci.remoting.RoleChecker;
 import org.apache.commons.io.FileUtils;
 
 /**
- * Log files from the different nodes
+ * Log files from the master node only.
  */
-@Extension(ordinal = 100.0) // put this first as largest content and can let the slower ones complete
+@Extension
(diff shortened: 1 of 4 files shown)


aheritier:
@christ66 Jacoco is breaking the test ? https://jenkins.ci.cloudbees.com/job/plugins/job/support-core-plugin/316/org.jenkins-ci.plugins$support-core/testReport/com.cloudbees.jenkins.support/SupportActionTest/takeSnapshotAndMakeSureSomethingHappens/
Otherwise +1 if the test failure isn't related to your change

christ66:
I think the test failure is related to my change. I can reproduce the same error message locally now.

aheritier:
🐝 🐝 🐝 hourra !!!

christ66:
Needs license headers

christ66:
I added a new license header. Could I get a re-review @reviewbybees ?

stephenc:
:bee:

aheritier:
re 🐝

christ66:
@reviewbybees done

PR #64, opened 2016-06-03
[JENKINS-33090] Ease the management bundles by the administrator

minudika:
[JENKINS-33090](https://issues.jenkins-ci.org/browse/JENKINS-33090)

This pull request is to keep track of the progress of GSoC project, Support Core Plugin Improvements.
@aheritier @christ66


aheritier:
The test failure isn't due by the PR ... unstable test

aheritier:
@minudika a best practice with `git`(for me) when you are managing such "long life" feature branch is to not do merges ("Merge branch 'master' of https://github.com/jenkinsci/support-core...") but instead to rebase regularly on top of the new master (which may requires also manage conflicts but at least you are fixing them in each commit you did)

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -68,14 +68,7 @@
...
 
-import java.io.BufferedOutputStream;
-import java.io.File;
-import java.io.FileOutputStream;
-import java.io.FilenameFilter;
-import java.io.IOException;
-import java.io.OutputStream;
-import java.io.PrintWriter;
-import java.io.StringWriter;
+import java.io.*;
```
aheritier: you need to reconfigure your IDE to not do this. It is better to keep all imports (at least as far as you don't have more than 10 from the same package)

christ66: See http://stackoverflow.com/a/3348855.

src/main/java/com/cloudbees/jenkins/support/BundleBrowser.java
```diff
@@ -0,0 +1,108 @@
+package com.cloudbees.jenkins.support;
```
oleg-nenashev: @aheritier @christ66 Is it fine for GSoC? I have some doubts about it, so maybe creating a community package would be preferable

src/main/java/com/cloudbees/jenkins/support/BundleBrowser.java
```diff
@@ -0,0 +1,108 @@
...
+       List<File> list = new ArrayList<File>();
+        List<File> zipFileList = getZipFileList();
+        for(Integer i : selectedFileIndices){
+            list.add(zipFileList.get(i));
+        }
+        return  list;
+    }
+
+    public void compressFiles(List<File>fileNames){
+        byte[] buffer = new byte[1024*1024*10];
```
oleg-nenashev: do you really need to create a new buffer for each operation? 10Mb is a big number, so parallel generations may hammer JVM

oleg-nenashev: 🐜 for the magic constant in the code. I've a;so seen bigger support bundles

src/main/java/com/cloudbees/jenkins/support/BundleBrowser.java
```diff
@@ -0,0 +1,108 @@
...
+
+                FileInputStream in =
+                        new FileInputStream(file);
+
+                int len;
+                while ((len = in.read(buffer)) > 0) {
+                    zos.write(buffer, 0, len);
+                }
+
+                in.close();
```
oleg-nenashev: It causes a risk of unclosed input file descriptors if a write operation fails. 🐜

src/main/java/com/cloudbees/jenkins/support/BundleBrowser.java
```diff
@@ -0,0 +1,108 @@
...
+            }
+
+            zos.closeEntry();
+            zos.close();
+
+            System.out.println("Done");
```
oleg-nenashev: Please use java.util.logging.Logger. System outputs should not be used in the production code

src/main/java/com/cloudbees/jenkins/support/BundleBrowser.java
```diff
@@ -0,0 +1,108 @@
...
+        }catch(IOException ex){
+            ex.printStackTrace();
```
oleg-nenashev: Should go to the system log as well

src/main/java/com/cloudbees/jenkins/support/BundleBrowser.java
```diff
@@ -0,0 +1,108 @@
...
+        }
+    }
+
+   /* public static void main(String args[]) throws IOException {
```
oleg-nenashev: dead code

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -65,6 +68,7 @@
 public class SupportAction implements RootAction {
 
     public static final Permission CREATE_BUNDLE = SupportPlugin.CREATE_BUNDLE;
+    private final String  bundleCollectionName = "BundleCollection";
```
oleg-nenashev: static? Also will need variable name case change

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -205,4 +209,73 @@ public boolean isSelected() {
...
     }
+
+    //=================================================================
+
+    public List<File> getList() throws IOException {
+        BundleBrowser bundleBrowser = new BundleBrowser();
+        return bundleBrowser.getZipFileList();
+    }
+
+    public ListBoxModel doFillGoalTypeItems() throws IOException {
```
oleg-nenashev: Please add <code>@Restricted(NoExternalUse)</code> to all stapler methods

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -205,4 +209,73 @@ public boolean isSelected() {
...
+        for(File file:list){
+            fileList.add(file);
+            items.add(file.getName(),file.getName());
+        }
+        return items;
+    }
+
+    public void doDownloadBundle(StaplerRequest req, StaplerResponse rsp) throws ServletException, IOException {
+        ArrayList<Integer> selectedFileIndices = new ArrayList<Integer>();
+        if (!"POST".equals(req.getMethod())) {
```
oleg-nenashev: There is a <code>@RequirePOST</code> annotation for stapler, which implements it

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -205,4 +209,73 @@ public boolean isSelected() {
...
+        File[] list = bundleBrowser.getFileList();
+
+        for(File file:list){
+            fileList.add(file);
+            items.add(file.getName(),file.getName());
+        }
+        return items;
+    }
+
+    public void doDownloadBundle(StaplerRequest req, StaplerResponse rsp) throws ServletException, IOException {
```
oleg-nenashev: 🐜 Missing security check

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
oleg-nenashev: Also makes sense to check Jenkins read permissions at least

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -385,6 +378,36 @@ public void close() throws IOException {
...
+                FileInputStream in =
+                        new FileInputStream(file);
+
+                int len;
+                buffer = new byte[1024*1024];
+                while ((len = in.read(buffer)) > 0) {
+                    zipOutputStream.write(buffer, 0, len);
+                }
+
+                in.close();
```
oleg-nenashev: Improper handling of the input stream

src/main/resources/com/cloudbees/jenkins/support/SupportAction/BundleBrowser.jelly
```diff
@@ -0,0 +1,57 @@
+<!--
+ ~ The MIT License
+ ~
+ ~ Copyright (c) 2013, CloudBees, Inc.
```
oleg-nenashev: ORLY?

src/main/resources/com/cloudbees/jenkins/support/SupportAction/BundleBrowser.jelly
```diff
@@ -0,0 +1,57 @@
...
+        </j:forEach>
+        <f:entry>
+          <f:submit value="${%Generate Bundle}"/>
+        </f:entry>
+      </f:form>
+
+
+    </l:main-panel>
+  </l:layout>
+</j:jelly>
```
oleg-nenashev: Good practice: add spaces to the end

src/main/resources/com/cloudbees/jenkins/support/SupportAction/index.jelly
```diff
@@ -37,31 +37,11 @@
...
-              <input style="display:none" name="name" value="${component.id}"/>
-            </div>
-          </f:entry>
-        </j:forEach>
-        <f:entry>
-          <f:submit value="${%Generate Bundle}"/>
-        </f:entry>
-      </f:form>
+
+      <a href="BundleGenerator">Download Bundles</a>
```
oleg-nenashev: Nice to have: internalization support. https://wiki.jenkins-ci.org/display/JENKINS/Internationalization

src/main/java/com/cloudbees/jenkins/support/BundleBrowser.java
christ66: @oleg-nenashev It's probably best we ask the mentor mailing list about this. Personal preference is to keep everything in the same package (searching gets harder if it is split).

src/main/java/com/cloudbees/jenkins/support/BundleBrowser.java
christ66: 🐜 as this should be a field which can be changed in the future easier.

In terms of the size I think this is too large for a buffer size. Buffers should be several Bytes, or KB (8,192 is the default size). Is there a reason for going beyond the default size?

src/main/java/com/cloudbees/jenkins/support/BundleBrowser.java
christ66: It might be better if we switch to java 7 which allows for the try-with-resources which will simplify this code.

src/main/java/com/cloudbees/jenkins/support/BundleBrowser.java
minudika: that buffer size was used just for testing purposes since I was not aware of a max size for a bundle. I will change it.

src/main/java/com/cloudbees/jenkins/support/BundleBrowser.java
aheritier: I think that all changes done by @minudika for GSOC could land in a 2.0 release, thus it may be the opportunity for such change too. We may start a 2.0 branch (or in master) with J7/Jenkins 1.617+ requirement. WDYT ?

oleg-nenashev: 1.625.x baseline is fine for me

src/main/java/com/cloudbees/jenkins/support/BundleBrowser.java
oleg-nenashev: @christ66 I've checked the GSoC rules, and there is no explicit guideline about it. Seems you can go forward with this naming

oleg-nenashev:
@minudika From what I see the most of my comments have not been really addressed yet.
They were just overriden by formatting. 

Any plans to polish the code?

minudika:
@oleg-nenashev  I am working on it. Will commit the polished code soon.

oleg-nenashev:
OK, thanks!

src/main/resources/com/cloudbees/jenkins/support/SupportAction/BundleBrowser.jelly
christ66: This should probably be changed to 2016 and remove `CloudBees, Inc.` in the title.

oleg-nenashev:
seems the PR is stuck :(

src/main/java/com/cloudbees/jenkins/support/BundleBrowser.java
```diff
@@ -0,0 +1,189 @@
...
+            File file = new File("config.txt");
+
+            if (file.createNewFile()){
+                FileWriter fileWriter = new FileWriter(file);
+                fileWriter.append("100");
+                fileWriter.close();
+                return 100;
+            }else{
+                FileReader fileReader = new FileReader(file);
+                BufferedReader bufferedReader = new BufferedReader(fileReader);
```
christ66: Need to make sure this stream is closed (use try/finally block)

src/main/java/com/cloudbees/jenkins/support/BundleBrowser.java
```diff
@@ -0,0 +1,189 @@
...
+
+    public void deleteBundle(List<Integer>indices){
+        for(Integer i : indices){
+            zipFileList.get(i).delete();
+            zipFileList.remove(i);
+        }
+    }
+
+    public String getJsonFileList() throws IOException {
+        String json="{'core' : { " +
```
christ66: Use `StringBuilder` as this concatenation can be slow.

oleg-nenashev:
@christ66 @minudika That's why I grumble about the status of this PR. My comments have been invalidated by formatting, and now new reviewers spend time on discovering same issues.

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
minudika: Could you please explain this? I didn't get that ;)

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -205,4 +213,151 @@ public boolean isSelected() {
...
+            } catch (FileNotFoundException ex) {
+                logger.log(Level.FINE, ex.getMessage(), ex);
+            } catch (IOException ex) {
+                logger.log(Level.FINE, ex.getMessage(), ex);
+            }
+        }catch (Exception e) {
+            PrintWriter out = rsp.getWriter();
+            rsp.setContentType("text/html");
+            out.println("<script type=\"text/javascript\">");
+            out.println("alert('Input must be an integer!');");
```
christ66: This should be instead replaced with a `doCheck` method as we're only checking for the version number (see https://youtu.be/eUzYZZsNBIA?t=37m24s)

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -205,4 +213,151 @@ public boolean isSelected() {
...
+    @Restricted(NoExternalUse.class)
+    public void doConfigure(StaplerRequest req,StaplerResponse rsp) throws ServletException, IOException {
+        String fileName = "config.txt";
+        String numberOfDays = req.getSubmittedForm().getString("numberOfDays");
+        try {
+            Integer.parseInt(numberOfDays);
+            try {
+                File file = new File(fileName);
+                FileWriter fileWriter = new FileWriter(file);
+                fileWriter.append(numberOfDays);
```
christ66: Implement `Saveable` interface instead.

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -205,4 +213,151 @@ public boolean isSelected() {
...
+        List<Integer> indexList = getSelectedIndices(req);
+        BundleBrowser bundleBrowser = BundleBrowser.getBundleBrowser();
+        bundleBrowser.deleteBundle(indexList);
+        PrintWriter out = rsp.getWriter();
+        out.println("</script>");
+        rsp.forwardToPreviousPage(req);
+
+        rsp.setContentType("text/html");
+        out.println("<script type=\"text/javascript\">");
+        out.println("alert('Bundle(s) Deleted!');");
```
christ66: I would instead create a new IOException which the message says that the bundle was already deleted. Let the exception bubble up as Jenkins knows how to handle these exceptions.

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -205,4 +213,151 @@ public boolean isSelected() {
...
+        if (!json.has("components")) {
+            rsp.sendError(HttpServletResponse.SC_BAD_REQUEST);
+            return;
+        }
+
+        List<Integer> indexList = getSelectedIndices(req);
+        BundleBrowser bundleBrowser = BundleBrowser.getBundleBrowser();
+        bundleBrowser.deleteBundle(indexList);
+        PrintWriter out = rsp.getWriter();
+        out.println("</script>");
```
christ66: ?

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
oleg-nenashev: * http://stapler.kohsuke.org/apidocs/?org/kohsuke/stapler/interceptor/RequirePOST.html
* Example: https://github.com/jenkinsci/jenkins/blob/31ad77005ea1cb9ab07e2312b6dadd1eb15ab72b/core/src/main/java/hudson/model/MyView.java#L65

christ66: You can use the `@RequirePOST` annotation instead of the check for if it is not a `POST`.

jglick:
`SupportActionTest.takeSnapshotAndMakeSureSomethingHappens` is failing, perhaps due to JaCoCo barfing on `SACParserCSS3TokenManager`.

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -205,4 +213,151 @@ public boolean isSelected() {
...
+            doDownloadBundles(req,rsp);
+        }
+        else{
+            doDeleteBundles(req,rsp);
+        }
+    }
+
+    @Restricted(NoExternalUse.class)
+    public void doDownloadBundles(StaplerRequest req, StaplerResponse rsp) throws ServletException, IOException {
+        if (!"POST".equals(req.getMethod())) {
```
jglick: Just use `@RequirePOST` and this is automatic.

src/main/resources/com/cloudbees/jenkins/support/SupportAction/Test.jelly
```diff
@@ -0,0 +1,61 @@
...
+    <l:main-panel>
+      <h1>
+        <img src="${resURL}/plugin/support-core/images/32x32/support.png" alt=""/>
+        ${it.actionTitle}
+      </h1>
+      <br/>
+
+      <table>
+        <tr>
+          <td><a href="BundleGenerator"><img src="${resURL}/plugin/support-core/images/commons/CreateBundle.png" alt=""/></a></td>
```
jglick: Conventionally Stapler views are named like Java methods, so e.g. `bundleGenerator`, or lowercase/hyphenated, e.g. `bundle-generator`.

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
minudika: Are you suggesting to use this [1] interface or implement the memento design pattern?
[1] http://help.eclipse.org/mars/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fapi%2Forg%2Feclipse%2Fui%2FSaveable.html

minudika:
I have sent a PR with all the changes I've made without the gsoc description. Please have a look

oleg-nenashev:
@minudika Thanks! I'll update the metadata of that PR

PR #65, opened 2016-06-17
[WIP] Support CLI to specify a directory as an argument.

christ66:


oleg-nenashev:
Since the permission is not restricted to Jenkins.ADMINISTER, @batmat has a point.
Actually I see no reason in overriding tempdir, the command should be able to return the bundle to the requester's filesystem. E.g. `generate-bundle > bundle.zip` or something more convenient

christ66:
Re-thinking this I think it should put the support bundle in the location where the command was called instead of temp directory or being able to specify a location.

recampbell:
This should probably be closed now that the remoting CLI protocol is being deprecated. When using non-remoting, the output is sent to stdout which is more usable anyway.

PR #66, opened 2016-06-21
Upgrade parent 2.3 -> 2.11

aheritier:
Signed-off-by: Arnaud HERITIER <aheritier@apache.org>


pom.xml
```diff
@@ -98,6 +98,13 @@
       <artifactId>metrics</artifactId>
       <version>3.0.0</version>
     </dependency>
+    <!-- test dependencies -->
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpcore</artifactId>
+      <version>4.4.1</version>
+      <scope>test</scope>
```
jglick: Do you rather mean to `<exclude>` an old version of this dep from some “nearer” dependency chain? CC @andresrc

jglick:
:+1:

pom.xml
aheritier: No @jglick This dep is used in some tests but it's not anymore in the dependency tree with recent version of test harness. It remains that it is a good practice to always declare at project level all the dependencies used (mvn dependency:analyze is more or less your friend)

andresrc: @aheritier Please see [JENKINS-35445](https://issues.jenkins-ci.org/browse/JENKINS-35445). This specific dependency was causing problems in the `hpi` packaging for some dependency graphs, so it has been shaded in the JTH.

If the dependency does not appear on any trail of the compile dependencies, :+1:

aheritier: AH AH AH @andresrc @jglick I just found that the dependency was required by an orphan import cc @christ66 
I will update my PR

aheritier:
Build failure isn't related to my PR but to the CI infra :(

oleg-nenashev:
👍 , but can be 2.13 now

andresrc:
:+1:

PR #67, opened 2016-06-28
Jenkins 33090

minudika:
[JENKINS-33090](https://issues.jenkins-ci.org/browse/JENKINS-33090)

Comments from @oleg-nenashev: It's a reopened version of https://github.com/jenkinsci/support-core-plugin/pull/64 . That PR contains much review comments from different contributors


config.txt
```diff
@@ -0,0 +1 @@
+100
```
aheritier: what is it ?

minudika: That is used to keep values related to the configuration such as purging age of an old bundle etc..

PR #68, opened 2016-07-05
Diagnose the startup performance

christ66:
In certain situations Jenkins will take a very long time to start up. We add some new diagnostic information when the Jenkins plugin is loaded to obtain more information about what threads are currently running which could be causing the slow startup. The diagnostic information is only gathered until the `InitMilestone.COMPLETED` is reached in which case the thread dump stops being generated. By default this is disabled and the default time interval is 60 seconds, but can be customized via java arguments.

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
+++ b/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
@@ -30,6 +30,7 @@ import com.cloudbees.jenkins.support.api.Content;
 import com.cloudbees.jenkins.support.api.StringContent;
 import com.cloudbees.jenkins.support.api.SupportProvider;
 import com.cloudbees.jenkins.support.api.SupportProviderDescriptor;
+import com.cloudbees.jenkins.support.impl.ThreadDumps;
 import com.cloudbees.jenkins.support.util.Helper;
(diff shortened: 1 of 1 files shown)


abayer:
:bug: on the findbugs warnings, otherwise fine.

christ66:
@abayer fixed findbugs issues

abayer:
:bee:

stephenc:
:bee:

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -396,6 +402,59 @@ public static void loadConfig() throws IOException {
             instance.load();
     }
 
+    private static final boolean logStartupPerformanceIssues = Boolean.getBoolean(SupportPlugin.class.getCanonicalName() + ".threadDumpStartup");
```
christ66: @stephenc would it be easier if we piggybacked off of the environment variable: `-Djenkins.model.Jenkins.logStartupPerformance=true` instead of creating a custom one?

stephenc: Whatever you want, but if this causes issues, may need to turn it off separately to my mind

PR #69, opened 2016-07-06
PSE Support Bundle

spnchian:
Added an action that will make it easier for PSE to request a support bundle from the Support Core Plugin. The added action is very similar to doDownload in the way it generates the support bundle.

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/SupportAction.java
+++ b/src/main/java/com/cloudbees/jenkins/support/SupportAction.java
@@ -42,6 +42,7 @@ import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.StaplerRequest;
 import org.kohsuke.stapler.StaplerResponse;
 
+import javax.servlet.Servlet;
 import javax.servlet.ServletException;
 import javax.servlet.ServletOutputStream;
 import javax.servlet.http.HttpServletResponse;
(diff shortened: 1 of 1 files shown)


spnchian:
@reviewbybees

Vlatombe:
🐛  support-core shouldn't mention PSE at all. Given your new method takes the same parameters as the doDownload method and is roughly making components optional, I would refactor the `doDownload` to do what you need.

aheritier:
Yes support-core must remain generic. If required you can add new extension points. At CloudBees we have a specific plugin implementing these extension points to manage everything which is specific to our product

stephenc:
If this is a CloudBees specific generic extension it should go in CloudBees-support

If this is a PSE specific extension the PSE should host it in either a specific CloudBees-PSE-support plugin or in an existing PSE plugin

:bug: on this code going here

jglick:
This is not an extension, it is an alternate download mode. :bug: as per @Vlatombe.

oleg-nenashev:
🐛 to just join the party. Private-source extensions do not belong to OSS

spnchian:
As stated above, this was originally written to be an alternate download. I am trying to access this plugin using curl commands and unfortunately, the form submission format was not working in the `doDownload`. I can't seem to be able to format the form correctly to allow the curl command to be processed through. Could someone give an example of what would work?

spnchian:
Nevermind. I got it to work.

PR #71, opened 2016-07-26
[Jenkins-33394] Generate support bundle before providing download

minudika:
When downloading the created zip file, it will be compressed and will be downloaded as a zip file in a zip file).


minudika:
There are some already merged commits have been added to this PR when I get a pull from master branch. Will it be ok?

oleg-nenashev:
@minudika should be OK if maintainers squash commits. But it's not ideal

PR #72, opened 2016-08-14
[FIX JENKINS-33090] Add UI Changes to make support bundle management easier

christ66:
@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/SupportAction.java
+++ b/src/main/java/com/cloudbees/jenkins/support/SupportAction.java
@@ -29,7 +29,13 @@ import com.cloudbees.jenkins.support.api.SupportProvider;
 
 import com.cloudbees.jenkins.support.util.Helper;
 import hudson.Extension;
+import hudson.FilePath;
+import hudson.Util;
+import hudson.model.Action;
+import hudson.model.DirectoryBrowserSupport;
+import hudson.model.InvisibleAction;
 import hudson.model.RootAction;
+import hudson.model.Run;
 import hudson.security.ACL;
 import hudson.security.Permission;
 import jenkins.model.Jenkins;
@@ -37,16 +43,26 @@ import net.sf.json.JSONObject;
 
 import org.acegisecurity.context.SecurityContext;
 import org.acegisecurity.context.SecurityContextHolder;
+import org.apache.commons.io.IOUtils;
 import org.jvnet.localizer.Localizable;
 import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.HttpResponse;
 import org.kohsuke.stapler.StaplerRequest;
 import org.kohsuke.stapler.StaplerResponse;
+import org.kohsuke.stapler.interceptor.RequirePOST;
 
+import javax.annotation.Nonnull;
 import javax.servlet.ServletException;
 import javax.servlet.ServletOutputStream;
 import javax.servlet.http.HttpServletResponse;
 
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FilenameFilter;
 import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.OutputStream;
+import java.nio.file.Files;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Date;
@@ -57,6 +73,7 @@ import java.util.Set;
 import java.util.TimeZone;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.zip.ZipOutputStream;
 
 /**
  * Main root action for generating support.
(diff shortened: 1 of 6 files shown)


src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -29,25 +29,40 @@
...
 import org.acegisecurity.context.SecurityContextHolder;
+import org.apache.commons.io.IOUtils;
 import org.jvnet.localizer.Localizable;
 import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.HttpResponse;
 import org.kohsuke.stapler.StaplerRequest;
 import org.kohsuke.stapler.StaplerResponse;
 import org.kohsuke.stapler.interceptor.RequirePOST;
 
+import javax.annotation.Nonnull;
```
jtnord: should use the deprecated findbugs annotations.

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -114,6 +130,72 @@ public Localizable getActionBlurb() {
         return SupportPlugin.getComponents();
     }
 
+    public class Bundle {
+        File location;
```
jtnord: private

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -114,6 +130,72 @@ public Localizable getActionBlurb() {
...
+                Util.copyStreamAndClose(new FileInputStream(location), os);
+            } finally {
+                IOUtils.closeQuietly(os);
+            }
+        }
+    }
+
+    public final class BundleList extends ArrayList<Bundle> {
+        public void doBundle(StaplerRequest req, StaplerResponse rsp) throws IOException {
+            String restOfPath = req.getRestOfPath();
```
jtnord: Permission check - bundle may contain information that an arbitrary user should not be able to see?

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -114,6 +130,72 @@ public Localizable getActionBlurb() {
...
+            try {
+                rsp.setContentType("application/zip");
+                Util.copyStreamAndClose(new FileInputStream(location), os);
+            } finally {
+                IOUtils.closeQuietly(os);
+            }
+        }
+    }
+
+    public final class BundleList extends ArrayList<Bundle> {
```
jtnord: Should be static as does not use any out class methods.

Also not convinced that this should extend the Array List, rather than have a method in the parent and just store a `ArrayList<Bundle>` in the parent

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -114,6 +130,72 @@ public Localizable getActionBlurb() {
...
+                // don't serve anything other than files in the artifacts dir
+                rsp.sendError(HttpServletResponse.SC_BAD_REQUEST);
+                return;
+            }
+            if (restOfPath.charAt(0) == '/') {
+                restOfPath = restOfPath.substring(1);
+            }
+            for (Bundle b : this) {
+                if (b.getName().equals(restOfPath)) {
+                    b.doDownload(req, rsp);
```
jtnord: assuming that no too bundles have the same name - which they could if they where in a different (sub) directory

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -114,6 +130,72 @@ public Localizable getActionBlurb() {
         return SupportPlugin.getComponents();
     }
 
+    public class Bundle {
```
jtnord: should be static or in a separate file - does not use any of the outer classes internals.

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -114,6 +130,72 @@ public Localizable getActionBlurb() {
...
+            }
+        }
+    }
+
+    public @Nonnull BundleList getBundles() {
+        BundleList bundles = new BundleList();
+        for (File f : SupportPlugin.getRootDirectory().listFiles(new FilenameFilter() {
+            @Override
+            public boolean accept(File dir, String name) {
+                return name.contains(".zip");
```
jtnord: so `foo.zip.blah` should be accepted?  `endsWith` is probably more correct and more performant.

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -114,6 +130,72 @@ public Localizable getActionBlurb() {
...
+                if (b.getName().equals(restOfPath)) {
+                    b.doDownload(req, rsp);
+                }
+            }
+        }
+    }
+
+    public @Nonnull BundleList getBundles() {
+        BundleList bundles = new BundleList();
+        for (File f : SupportPlugin.getRootDirectory().listFiles(new FilenameFilter() {
```
jtnord: OT: you could change this and the other parts of the class to the more efficient [NIO2](https://docs.oracle.com/javase/7/docs/api/java/nio/file/DirectoryStream.html)

stephenc:
Link to RFE [JENKINS-33090](https://issues.jenkins-ci.org/browse/JENKINS-33090)

stephenc:
Would probably be simpler to have the bundles as a sub-action from the support action with its own page rather than reusing the crappy tab-bar

Additionally moving to a separate page would make it easier to enforce permission restrictions such that the bundles link and pages would only be accessible to the Administrator (which is necessary as the auto-generated bundles will certainly include information that the administrator *may* want to restrict from general availablity)

As such :bug: on the current implementation, but I am 👍 on the idea

christ66:
>  rather than reusing the crappy tab-bar

I like the tab-bar ;).

Will close and use sub-actions instead.

PR #73, opened 2016-08-29
[JENKINS-37772] - Get rid of Initializer(after =  InitMilestone.COMPLETED)

oleg-nenashev:
https://issues.jenkins-ci.org/browse/JENKINS-37772

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
+++ b/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
@@ -68,6 +68,8 @@ import org.apache.commons.lang.StringUtils;
 import org.apache.tools.zip.ZipEntry;
 import org.apache.tools.zip.ZipOutputStream;
 import org.jenkinsci.remoting.RoleChecker;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.kohsuke.stapler.StaplerRequest;
 
 import java.io.BufferedOutputStream;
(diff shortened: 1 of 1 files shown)


src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -427,7 +429,11 @@ public static void threadDumpStartup() throws Exception {
         Thread t = new Thread("Support core plugin startup diagnostics") {
             @Override
             public void run() {
-                while (!milestonesCompleted) {
+                while (true) {
+                    final Jenkins jenkins =Jenkins.getInstance();
+                    if (jenkins == null || jenkins.getInitLevel() != InitMilestone.COMPLETED) {
+                        continue;
```
stephenc: Should you not have even a `Thread.yield()` in here (better a 10ms sleep perhaps) to avoid CPU spin of death)

Also looks like you may have invert the logic

oleg-nenashev: Just fixed the code without performance polishing.
@christ66 and @aheritier may enhance it later

aheritier:
LGTM :bee:

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
aheritier: 👍 to apply @stephenc improvement in another PR

recena:
🐝 but I agree with @stephenc

oleg-nenashev:
@reviewbybees done
Would be great to have it in the master in order to avoid occasional release without this fix

PR #75, opened 2016-09-28
Optional parameter to generate a support bundle when a node goes offline

christ66:
@reviewbybees esp. @oleg-nenashev

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
+++ b/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
@@ -683,7 +683,9 @@ public class SupportPlugin extends Plugin {
                         PeriodicWork.all().get(PeriodicWorkImpl.class).doRun();
                     }
                 } catch (Exception e) {
-                    e.printStackTrace();
+                    Logger.getLogger(SupportPlugin.class.getName())
+                            .log(Level.WARNING,
+                            "Error when generating support bundle while on offline node", e);
                 } finally {
                     if (locked)
                         lock.unlock();
--- a/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
+++ b/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
@@ -677,7 +677,8 @@ public class SupportPlugin extends Plugin {
 
         @Override
         public void onOffline(@Nonnull final Computer c, @CheckForNull OfflineCause cause) {
-            if (Boolean.getBoolean(SupportPlugin.class.getName() + ".generateOnOfflineNode")) {
+            if (Boolean.getBoolean(SupportPlugin.class.getName() + ".generateOnOfflineNode")
+                    && cause != null && cause.getClass().getName().contains("RestartCause")) {
                 Timer.get().schedule(new java.util.concurrent.Callable<Void>() {
                     @Override
                     public Void call() throws Exception {
--- a/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
+++ b/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
@@ -54,6 +54,7 @@ import hudson.security.Permission;
 import hudson.security.PermissionGroup;
 import hudson.security.PermissionScope;
 import hudson.slaves.ComputerListener;
+import hudson.slaves.OfflineCause;
 import hudson.util.IOUtils;
 import hudson.util.TimeUnit2;
 import jenkins.metrics.impl.JenkinsMetricProviderImpl;
@@ -72,6 +73,7 @@ import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.kohsuke.stapler.StaplerRequest;
 
+import javax.annotation.Nonnull;
 import java.io.BufferedOutputStream;
 import java.io.File;
 import java.io.FileNotFoundException;
@@ -646,6 +648,7 @@ public class SupportPlugin extends Plugin {
 
     @Extension
     public static class ComputerListenerImpl extends ComputerListener {
+        private static final Logger LOGGER = Logger.getLogger(ComputerListenerImpl.class.getCanonicalName());
         @Override
         public void onOnline(Computer c, TaskListener listener) throws IOException, InterruptedException {
             final Node node = c.getNode();
@@ -666,6 +669,29 @@ public class SupportPlugin extends Plugin {
             } catch (RuntimeException e) {
                 Logger.getLogger(SupportPlugin.class.getName()).log(Level.WARNING,
                         "Could not install root log handler on node: " + c.getName(), e);
+            }
+        }
+
+        @Override
+        public void onOffline(@Nonnull Computer c, @CheckForNull OfflineCause cause) {
+            if (Boolean.getBoolean(SupportPlugin.class.getCanonicalName() + ".generateOnOfflineNode")) {
+                SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd_HH.mm.ss");
+                dateFormat.setTimeZone(TimeZone.getTimeZone("GMT"));
+
+                final String bundlePrefix = "support";
+                File file = new File(SupportPlugin.getRootDirectory(),
+                        bundlePrefix + "_" + dateFormat.format(new Date()) + ".zip");
+                FileOutputStream fos = null;
+                try {
+                    fos = new FileOutputStream(file);
+                    SupportPlugin.writeBundle(fos);
+                } catch (FileNotFoundException e) {
+                    Logger.getLogger(SupportPlugin.class.getName()).log(Level.WARNING, e.getMessage(), e);
+                } catch (IOException e) {
+                    Logger.getLogger(SupportPlugin.class.getName()).log(Level.WARNING, e.getMessage(), e);
+                } finally {
+                    IOUtils.closeQuietly(fos);
+                }
             }
         }
     }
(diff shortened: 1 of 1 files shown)


src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -668,6 +671,29 @@ public void onOnline(Computer c, TaskListener listener) throws IOException, Inte
                         "Could not install root log handler on node: " + c.getName(), e);
             }
         }
+
+        @Override
+        public void onOffline(@Nonnull Computer c, @CheckForNull OfflineCause cause) {
+            if (Boolean.getBoolean(SupportPlugin.class.getCanonicalName() + ".generateOnOfflineNode")) {
```
oleg-nenashev: Maybe makes sense to delay the generation for a minute. Just to get info about the impact (job termination, triggers, etc., etc.).

christ66: Fixed by adding locking in latest commit.

oleg-nenashev: 🐛 It blocks all listeners

christ66:
The new commit uses the `PeriodicWorkImpl` class which spawns a new thread. We would still need the locking to prevent N threads from being created in the event all are disconnected at once.

oleg-nenashev:
🐝 even though there is no explicit quiet time for the task

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -668,6 +672,26 @@ public void onOnline(Computer c, TaskListener listener) throws IOException, Inte
...
+                boolean locked = false;
+                try {
+                    locked = lock.tryLock();
+                    if (locked) {
+                        PeriodicWork.all().get(PeriodicWorkImpl.class).doRun();
+                    }
+                } catch (Exception e) {
+                    e.printStackTrace();
```
recampbell: 🐜 This should go to a logger instead

christ66: Ah I missed that! TY

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
jglick: Indeed, this ought to schedule the task asynchronously, and I agree that waiting a minute would be appropriate. Easy enough to schedule via `Timer.get()`. (I hope `PeriodicWorkImpl` is thread-safe already…)

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -668,6 +672,26 @@ public void onOnline(Computer c, TaskListener listener) throws IOException, Inte
...
         }
+
+        @Override
+        public void onOffline(@Nonnull Computer c, @CheckForNull OfflineCause cause) {
+            if (Boolean.getBoolean(SupportPlugin.class.getCanonicalName() + ".generateOnOfflineNode")) {
+                boolean locked = false;
+                try {
+                    locked = lock.tryLock();
+                    if (locked) {
+                        PeriodicWork.all().get(PeriodicWorkImpl.class).doRun();
```
recampbell: It would be better to log something like "Node xxx has gone offline, so a support bundle was generated. Please check $JENKINS_HOME/support for a bundle generated at this time.

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -668,6 +672,28 @@ public void onOnline(Computer c, TaskListener listener) throws IOException, Inte
                         "Could not install root log handler on node: " + c.getName(), e);
             }
         }
+
+        @Override
+        public void onOffline(@Nonnull Computer c, @CheckForNull OfflineCause cause) {
+            if (Boolean.getBoolean(SupportPlugin.class.getCanonicalName() + ".generateOnOfflineNode")) {
```
batmat: Somehow NIT: Though in that case `SupportPlugin.class.getCanonicalName()` will yield the same result as getName(), I'd use getName() since pretty much every calls with that pattern in the core use getName().

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -668,6 +674,42 @@ public void onOnline(Computer c, TaskListener listener) throws IOException, Inte
                         "Could not install root log handler on node: " + c.getName(), e);
             }
         }
+
+        @Override
+        public void onOffline(@Nonnull final Computer c, @CheckForNull OfflineCause cause) {
+            if (Boolean.getBoolean(SupportPlugin.class.getName() + ".generateOnOfflineNode")) {
```
oleg-nenashev: Maybe makes sense to generate bundles only when the cause is not null and not RestartCause (which means the termination is abnormal). https://github.com/jenkinsci/jenkins/search?utf8=%E2%9C%93&q=%22extends+OfflineCause%22&type=Code . RestartCause just may produce much flood since it happens when Jenkins performs restart.

Kinda 🐜

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
christ66: Fixed.

oleg-nenashev:
:bee:

christ66:
@jglick could I get a re-review?

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -668,6 +674,43 @@ public void onOnline(Computer c, TaskListener listener) throws IOException, Inte
                         "Could not install root log handler on node: " + c.getName(), e);
             }
         }
+
+        @Override
+        public void onOffline(@Nonnull final Computer c, @CheckForNull OfflineCause cause) {
+            if (Boolean.getBoolean(SupportPlugin.class.getName() + ".generateOnOfflineNode")
```
jglick:
```java
// TODO 2.4+ SystemProperties
```

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -668,6 +674,43 @@ public void onOnline(Computer c, TaskListener listener) throws IOException, Inte
...
+                    && cause != null && !cause.getClass().getName().contains("RestartCause")) {
```
jglick: If you are looking for `jenkins.model.Jenkins$MasterRestartNotifyier$RestartCause` please check whether the name is `equals` to that.

jglick: What about if `cause == null`? I would think you would want to generate a bundle in that case, right? But here you are not.

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -668,6 +674,43 @@ public void onOnline(Computer c, TaskListener listener) throws IOException, Inte
...
+                                                        " was generated. Please check: " +
+                                                        SupportPlugin.getRootDirectory().getAbsolutePath() +
+                                                        " at this time.");
+                            }
+                        } catch (Exception e) {
+                            Logger.getLogger(SupportPlugin.class.getName())
+                                    .log(Level.WARNING,
+                                            "Error when generating support bundle while on offline node", e);
+                        } finally {
+                            if (locked)
```
jglick: Simpler to put the `try`-`finally` block inside the `if` block.

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -668,6 +674,43 @@ public void onOnline(Computer c, TaskListener listener) throws IOException, Inte
...
+            if (Boolean.getBoolean(SupportPlugin.class.getName() + ".generateOnOfflineNode")
+                    && cause != null && !cause.getClass().getName().contains("RestartCause")) {
+                Timer.get().schedule(new java.util.concurrent.Callable<Void>() {
+                    @Override
+                    public Void call() throws Exception {
+                        boolean locked = false;
+                        try {
+                            locked = lock.tryLock();
+                            if (locked) {
+                                PeriodicWork.all().get(PeriodicWorkImpl.class).doRun();
```
jglick: IIUC this will generate a new bundle for every disconnection event, even if you get a bunch of them in sequence. Intentional? I thought we discussed coalescing these somehow?

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -668,6 +674,43 @@ public void onOnline(Computer c, TaskListener listener) throws IOException, Inte
...
+        @Override
+        public void onOffline(@Nonnull final Computer c, @CheckForNull OfflineCause cause) {
+            if (Boolean.getBoolean(SupportPlugin.class.getName() + ".generateOnOfflineNode")
+                    && cause != null && !cause.getClass().getName().contains("RestartCause")) {
+                Timer.get().schedule(new java.util.concurrent.Callable<Void>() {
+                    @Override
+                    public Void call() throws Exception {
+                        boolean locked = false;
+                        try {
+                            locked = lock.tryLock();
```
jglick: Should this lock be shared with `PeriodicWorkImpl` so we do not try to generate two background bundles at once?

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
batmat: Isn't `SystemProperties` `@Restricted(NoExternalUse.class)`?

PR #76, opened 2016-10-03
[FIX JENKINS-38572] Only run if async-http-client plugin is installed.

christ66:
@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/UpdateCenter.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/UpdateCenter.java
@@ -70,6 +70,8 @@ public class UpdateCenter extends Component {
                             // Only do this part of the async-http-client plugin is installed.
                             if (Helper.getActiveInstance().getPlugin("async-http-client") != null) {
                                 addProxyInformation(out);
+                            } else {
+                                out.println("Proxy: 'async-http-client' not installed so proxy info available.");
                             }
                         } finally {
                             out.flush();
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
Maybe it needs to be decoupled to a separate method to get it 100% safe from JIT stuff, but I'm not 100% sure

christ66:
@oleg-nenashev fixed.

oleg-nenashev:
🐝

src/main/java/com/cloudbees/jenkins/support/impl/UpdateCenter.java
```diff
@@ -66,16 +67,9 @@ public void writeTo(OutputStream os) throws IOException {
...
-                                out.println(" - Port: " + proxyServer.getPort());
-
-                                out.println(" - No Proxy Hosts: ");
-                                for (String noHost : proxyServer.getNonProxyHosts()) {
-                                    out.println(" * " + noHost);
-                                }
+                            // Only do this part of the async-http-client plugin is installed.
+                            if (Helper.getActiveInstance().getPlugin("async-http-client") != null) {
+                                addProxyInformation(out);
                             }
```
cblomqvist: You could also output that "async-http-client" is not installed and that is the reason why you are not outputting any proxy info. It could help in a troubleshooting scenario. That could mean moving the if statement into the `addProxyInformation` method to leave it to that method to also print something like 
```
out.println("Proxy: 'async-http-client' not installed so proxy info available.");
```
or something like that.

christ66:
@cblomqvist Added your comment.

stephenc:
:bee: (but given that the proxy details could be retrieved outside of using async-http-client I'd like to understand why the dependency is introduced

christ66:
@stephenc not sure. It could be the way I'm retrieving the proxy information that the dependency is required.

stephenc:
Then I suspect you are retrieving the proxy information wrong.

There is an idiomatic way of resolving the proxy information on a Jenkins master. That is all that AHCUtils is doing. Unless you are actually using AHC I suspect you are doing it wrong...

Until I see evidence to the contrary

:bug:

PR #77, opened 2016-10-21
collect boot metrics

ndeloof:
Collect boot metrics introduced by https://github.com/jenkinsci/jenkins/pull/2592
@reviewbybees


batmat:
@ndeloof I guess that as you closed the upstream one, this one cannot be considered anymore?...

PR #78, opened 2016-10-21
[FIXED JENKINS-39150] report remoting diagnostics when it's available

kohsuke:
Since this plugin cannot assume the version of core, access the method
in question via reflection (and also report the failure to find that
method, so that we have evidence either way.)

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/RemotingDiagnostics.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/RemotingDiagnostics.java
@@ -1,0 +1,51 @@
+package com.cloudbees.jenkins.support.impl;
+
+import com.cloudbees.jenkins.support.api.Component;
+import com.cloudbees.jenkins.support.api.Container;
+import com.cloudbees.jenkins.support.api.PrintedContent;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import hudson.remoting.Channel;
+import hudson.security.Permission;
+import jenkins.model.Jenkins;
+
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.lang.reflect.Method;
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -104,6 +104,12 @@
       <version>1.7.8</version>
       <optional>true</optional>
     </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci</groupId>
+      <artifactId>remoting</artifactId>
+      <version>2.62.3-SNAPSHOT</version>
```
stephenc: is this a left-over from testing?

kohsuke: Oops

stephenc:
:bee:

src/main/java/com/cloudbees/jenkins/support/impl/RemotingDiagnostics.java
```diff
@@ -0,0 +1,51 @@
...
+    @Override
+    public void addContents(@NonNull Container container) {
+        container.add(new PrintedContent("channel-diagnostics.md") {
+            @Override
+            protected void printTo(PrintWriter out) throws IOException {
+                // this method is new in remoting. see JENKINS-39150 change in remoting
+                try {
+                    Method m = Channel.class.getMethod("dumpDiagnosticsForAll", PrintWriter.class);
+                    m.invoke(null,out);
+                } catch (Exception e) {
```
oleg-nenashev: 🐜 I see no need to pollute output in the case of `MethodNotFoundException`. Maybe better to just write a warning and say which remoting version is required

jglick: Agreed, better to `catch (MethodNotFoundException)` here.

PR #80, opened 2016-10-26
[JENKINS-33394] Generate support bundle before providing download

minudika:
Please review this.
Implemented 3 options.
-Generate a bundle
-Download the latest existing bundle
-Generate and download (as how this works now)


src/main/resources/com/cloudbees/jenkins/support/SupportAction/index.jelly
```diff
@@ -58,8 +58,15 @@
             </div>
           </f:entry>
         </j:forEach>
+        <f:entry name="goalType"  field="goalType">
+          <select name="goalType">
+            <option value="generateBundle">${%Generate Bundle}</option>
+            <option value="download">${%Download Latest Bundle}</option>
+            <option value="download">${%Generate and Download}</option>
```
jglick: Two options, both `download`? Sounds like a mistake here somewhere.

Probably the UI should continue to just have a **Generate Bundle** button but the implementation should be rewritten to use JavaScript so that this triggers the generation, then a progress bar appears, and when the bundle is ready a download is initiated automatically.

batmat:
Conflicted and blocked since soon 1 year. Closing for clarity. Please don't hesitate to reopen if you can work on this again.
Thanks!

PR #81, opened 2016-10-28
Fixed some buglets in SupportLogFormatter

jglick:
See https://github.com/jenkinsci/workflow-job-plugin/pull/31 for an example.
- missing newline between header and start of exception
- gratuitous newline after stack trace
- spurious `: null` when a logger reported an exception without a message

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/SupportLogFormatter.java
+++ b/src/main/java/com/cloudbees/jenkins/support/SupportLogFormatter.java
@@ -33,6 +33,7 @@ import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.text.SimpleDateFormat;
 import java.util.Date;
+import java.util.TimeZone;
 import java.util.logging.Formatter;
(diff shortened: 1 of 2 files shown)


jglick:
Hmm, my call to `TimeZone.setDefault` works locally, but not on CI for some reason…

abayer:
:bee:

src/main/java/com/cloudbees/jenkins/support/SupportLogFormatter.java
```diff
@@ -28,10 +28,12 @@
  * DO NOT INCLUDE ANY NON JDK CLASSES IN HERE.
  * IT CAN DEADLOCK REMOTING - SEE JENKINS-32622
  ***********************************************/
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
```
christ66: @jtnord Is it ok to have a suppressFBWarning? I assume this should only impact compile time, but just want to check if this is ok.

PR #82, opened 2016-10-31
[JENKINS-39377] Feature/js errors

michaelneale:
Ticket https://issues.jenkins-ci.org/browse/JENKINS-39377

For discussion/ideas 

cc @christ66 

Looking at a way to include client side in an adjunct (in blue ocean it probably will have to add client side itself, and only once, as it doesn't use Jenkins classic js)


michaelneale:
FYI I consider this a WIP - that is missing is how to integrate client side (and if it is acceptable in its logging form).

src/main/java/com/cloudbees/jenkins/support/impl/BrowserErrorCollector.java
```diff
@@ -0,0 +1,72 @@
...
+ *            }
+ *
+ *
+ *  </code>
+ *
+ *         This will attempt to log the error with a maybe helpful JS stacktrace.
+ *
+ */
+@Extension
+public class BrowserErrorCollector implements RootAction, StaplerProxy {
```
christ66: Wondering if having it be an `UnprotectedRootAction` might be better, because if I have authentication issues and it's related to javascript, then information could still be obtained.

It might also be worth doing this instead as just an `InvisibleAction` (or `InvisibleRootAction`), so users do not see this all the time on the main Jenkins page. You can still navigate to it, but there wouldn't be an image in the UI.

michaelneale: good idea re invisible. 

The reason for protected is unprotected means people can just throw anything in the log, which may be opening a slight can of worms (if it was logged separately to others, sure). Obviously log ration etc would need to protect. 

In the case of anon read only access, yes would be nice to have this, but at a slight increased risk.

christ66: I might be imagining things...I thought `InvisibleRootAction` existed, but it looks like it doesn't.

PR #83, opened 2016-10-31
[FIXED JENKINS-39381] Add a support component for XML files

varmenise:


src/main/java/com/cloudbees/jenkins/support/configfiles/ConfigFileComponent.java
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Support component to add config.xml to the support bundle.
+ */
+@Extension
+public class ConfigFileComponent extends Component {
+    @NonNull
+    @Override
+    public Set<Permission> getRequiredPermissions() {
+        return Collections.singleton(Jenkins.ADMINISTER);    }
```
recena: 🐜 code style

src/main/java/com/cloudbees/jenkins/support/configfiles/OtherConfigFilesComponent.java
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * Support component for adding xml files to the support bundle.
+ */
+@Extension
+public class OtherConfigFilesComponent extends Component {
+    @NonNull
+    @Override
+    public Set<Permission> getRequiredPermissions() {
+        return Collections.singleton(Jenkins.ADMINISTER);    }
```
recena: 🐜 ditto

src/test/java/com/cloudbees/jenkins/support/configfiles/OtherConfigFilesComponentTest.java
```diff
@@ -0,0 +1,78 @@
...
+ * Created by valentina on 28/10/16.
+ */
+public class OtherConfigFilesComponentTest {
+
+    @Rule
+    public JenkinsRule r = new JenkinsRule();
+
+    private String xml;
+
+    private final String expectedXml = "<com.cloudbees.plugins.credentials.SystemCredentialsProvider plugin=\"credentials@1.18\">\n" +
```
recena: Why not store on a file?

varmenise: cause I generate the password during the execution.

src/main/java/com/cloudbees/jenkins/support/configfiles/OtherConfigFilesComponent.java
```diff
@@ -0,0 +1,69 @@
...
+    @NonNull
+    @Override
+    public String getDisplayName() {
+        return "Other Jenkins Configuration Files";
+    }
+
+    @Override
+    public void addContents(@NonNull Container container) {
+        Jenkins jenkins = Jenkins.getInstance();
+        if (jenkins != null) {
```
batmat: unnecessary null-check. Jenkins instance can (almost) never be null. See Stephen's explanation [here](https://github.com/jenkinsci/jenkins/commit/bb7c8fcedbcc9b51c5b1bb5b32810af5ac6b1ffb) e.g. (note: that commit was reverted afterwards for other reasons :))

src/main/java/com/cloudbees/jenkins/support/configfiles/OtherConfigFilesComponent.java
```diff
@@ -0,0 +1,69 @@
...
+            File dir = jenkins.getRootDir();
+            File[] files = dir.listFiles(new FilenameFilter() {
+                public boolean accept(File dir, String name) {
+                    //all xml files but credentials.xml
+                    return name.toLowerCase().endsWith(".xml") && !name.equals("credentials.xml");
+                }
+            });
+            if (files != null) {
```
batmat: Reading the javadoc, I don't think that code path should ever yield a null `files`. That would basically mean ` jenkins.getRootDir();` returns garbage, which seems unlikely.

Maybe we want to reverse the logic to log a warning (or even an ERROR) here if this happens, or even throw an exception?

src/main/java/com/cloudbees/jenkins/support/configfiles/OtherConfigFilesComponent.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Override
+    public void addContents(@NonNull Container container) {
+        Jenkins jenkins = Jenkins.getInstance();
+        if (jenkins != null) {
+            File dir = jenkins.getRootDir();
+            File[] files = dir.listFiles(new FilenameFilter() {
+                public boolean accept(File dir, String name) {
+                    //all xml files but credentials.xml
+                    return name.toLowerCase().endsWith(".xml") && !name.equals("credentials.xml");
```
batmat: IIUC what you did here, shouldn't you here *exclude* the config.xml as it's already handled separately in the other class?

src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
```diff
@@ -0,0 +1,90 @@
...
+ */
+public class SecretHandler {
+
+    /**
+     * our placeholder
+     */
+    protected static final String SECRET_MARKER = "#secret#";
+
+    /**
+     * fine the secret in the xml file and replace it with the place holder
```
batmat: typo "fine"

src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
```diff
@@ -0,0 +1,90 @@
...
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.StringReader;
+
+import org.xml.sax.helpers.XMLFilterImpl;
+import org.xml.sax.helpers.XMLReaderFactory;
+
+/**
+ * Secret Handler for xml files to add to the support bundle.
+ * We want to use a place holder instead of Secrets.
```
batmat: I _think_ place holder should be _placeholder_

src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
```diff
@@ -0,0 +1,90 @@
...
+                tagName = "";
+                super.endElement(uri, localName, qName);
+            }
+
+            @Override
+            public void characters(char[] ch, int start, int length) throws SAXException {
+                if (!tagName.equals("")) {
+                    String value = new String(ch, start, length).trim();
+                    //if it's a secret, then use a place holder
+                    if (!value.equals("") && Secret.decrypt(value) != null) {
```
batmat: `"".equals(value)` is IMO generally safer to not ever encounter an NPE, always putting static strings on the left hand side.

src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
```diff
@@ -0,0 +1,90 @@
...
+     * fine the secret in the xml file and replace it with the place holder
+     * @param xmlFile we want to parse
+     * @return the patched xml files without secrets
+     * @throws ParserConfigurationException
+     * @throws SAXException
+     * @throws IOException
+     * @throws XMLStreamException
+     * @throws TransformerException
+     */
+    public static File findSecrets(File xmlFile) throws ParserConfigurationException, SAXException, IOException, XMLStreamException, TransformerException {
```
batmat: I wonder if we shouldn't better hide those exceptions here or at least handle some locally, to rule out the comment I made above about using a catch(Exception) when calling that method.

src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
```diff
@@ -0,0 +1,90 @@
...
+            }
+
+            public void endElement(String uri, String localName, String qName) throws SAXException {
+                tagName = "";
+                super.endElement(uri, localName, qName);
+            }
+
+            @Override
+            public void characters(char[] ch, int start, int length) throws SAXException {
+                if (!tagName.equals("")) {
```
batmat: `!"".equals(tagName)` so that NPE is impossible

src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
```diff
@@ -0,0 +1,90 @@
...
+                super.characters(ch, start, length);
+            }
+        };
+        Source src = new SAXSource(xr, new InputSource(new StringReader(FileUtils.readFileToString(xmlFile))));
+        File patchedFile = File.createTempFile("patched", ".xml");
+        Result res = new StreamResult(new FileOutputStream(patchedFile));
+        Transformer transformer = TransformerFactory.newInstance().newTransformer();
+        //omit xml declaration because of https://bugs.openjdk.java.net/browse/JDK-8035437
+        transformer.setOutputProperty(OutputKeys.OMIT_XML_DECLARATION, "yes");
+        transformer.transform(src, res);
```
batmat: Typically, we should probably catch `TransformerException` here, log it, and just rethrow an IOException chained with or something in case it happens.

src/main/java/com/cloudbees/jenkins/support/configfiles/OtherConfigFilesComponent.java
varmenise: findbugs complains otherwise. This depends on an old version of jenkins core

src/test/java/com/cloudbees/jenkins/support/configfiles/OtherConfigFilesComponentTest.java
varmenise: cause I am comparing strings eventually, one being dynamic since I generate the secret on runtime. It's an overkill  IMHO

src/main/java/com/cloudbees/jenkins/support/configfiles/ConfigFileComponent.java
```diff
@@ -0,0 +1,53 @@
...
+    @Override
+    public void addContents(@NonNull Container container) {
+        Jenkins jenkins = Jenkins.getInstance();
+        if(jenkins != null) {
+            File configFile = new File(jenkins.getRootDir(), "config.xml");
+            if (configFile.exists()) {
+                try {
+                    File patchedXmlFile = SecretHandler.findSecrets(configFile);
+                    container.add(new FileContent("jenkins-root-configuration-files/" + configFile.getName(), patchedXmlFile));
+                } catch (Exception e) {
```
varmenise: 👍

src/main/java/com/cloudbees/jenkins/support/configfiles/OtherConfigFilesComponent.java
```diff
@@ -0,0 +1,69 @@
...
+                }
+            });
+            if (files != null) {
+            for (int i = 0; i < files.length; i++) {
+                File configFile = files[i];
+                if (configFile.exists()) {
+                    try {
+                        File patchedXmlFile = SecretHandler.findSecrets(configFile);
+                        container.add(new FileContent("jenkins-root-configuration-files/" + configFile.getName(), patchedXmlFile));
+                    } catch (Exception e) {
```
varmenise: 👍

src/main/java/com/cloudbees/jenkins/support/configfiles/ConfigFileComponent.java
```diff
@@ -0,0 +1,53 @@
...
+                    logger.log(Level.WARNING, "could not add the {0} configuration file to the support bundle because of: {1}", new Object[]{configFile.getName(), e});
```
varmenise: well, it's still a support bundle we're trying to create...we don't want to scare users but we will be able to see that if this happens..I think it should stay a warning but maybe we can have some other inputs here

src/main/java/com/cloudbees/jenkins/support/configfiles/ConfigFileComponent.java
```diff
@@ -0,0 +1,53 @@
...
+                }
+            }
```
varmenise: I think it does not exists at startup before you save it the first time...

src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
varmenise: well, we're handling everything on the other site..why do you want me to log it here?

src/main/java/com/cloudbees/jenkins/support/configfiles/ConfigFileComponent.java
varmenise: if I start the plugin with hpi:run then until I save the config, the file does not exists. This does not happen if I normally upload the plugin on jenkins. So something may have changed in later versions..anyways I am adding a warning

varmenise:
@reviewbybees

aheritier:
cc @christ66 (I didn't yet reviewed it myself but can you schedule some time next days to review this PR please ?)

src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
```diff
@@ -0,0 +1,90 @@
...
+                tagName = "";
+                super.endElement(uri, localName, qName);
+            }
+
+            @Override
+            public void characters(char[] ch, int start, int length) throws SAXException {
+                if (!"".equals(tagName)) {
+                    String value = new String(ch, start, length).trim();
+                    //if it's a secret, then use a place holder
+                    if (!"".equals(value) && Secret.decrypt(value) != null) {
```
stephenc: May also want to handle SecretBytes from latest credentials plugin

src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
```diff
@@ -87,4 +91,8 @@ public void characters(char[] ch, int start, int length) throws SAXException {
 
         return patchedFile;
     }
+
+    private static boolean isSecretBytes(String value) {
+        return (SecretBytes.decrypt(Base64.decode(value.substring(1, value.length() - 1))) != null);
```
stephenc: Will have issues if the string is length 1.

You could check first that the string starts with `{` ends with `}` and is at least 8 characters long (may be able to increase the min length further if we made assumptions about the salt length)

src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
```diff
@@ -0,0 +1,98 @@
...
+        Transformer transformer = TransformerFactory.newInstance().newTransformer();
+        //omit xml declaration because of https://bugs.openjdk.java.net/browse/JDK-8035437
+        transformer.setOutputProperty(OutputKeys.OMIT_XML_DECLARATION, "yes");
+        transformer.transform(src, res);
+
+        return patchedFile;
+    }
+
+    private static boolean isSecretBytes(String value) {
+        return (SecretBytes.decrypt(Base64.decode(value.substring(1, value.length() - 1))) != null);
```
stephenc: I think you are doing this wromg

src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
varmenise: aren't they alwasy saved on the xml file with the toString() method? the toString method adds the `{}`

varmenise: https://github.com/jenkinsci/credentials-plugin/blob/master/src/main/java/com/cloudbees/plugins/credentials/SecretBytes.java#L297

varmenise:
I will move the `isSecretBytes ` logic in the `credential-plugin` as discussed offline with Stephen

src/main/java/com/cloudbees/jenkins/support/configfiles/ConfigFileComponent.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @NonNull
+    @Override
+    public String getDisplayName() {
+        return "Jenkins Global Configuration File";
+    }
+
+    @Override
+    public void addContents(@NonNull Container container) {
+        Jenkins jenkins = Jenkins.getInstance();
```
christ66: :bug: Use https://github.com/jenkinsci/support-core-plugin/blob/master/src/main/java/com/cloudbees/jenkins/support/util/Helper.java#L21 `Helper.getActiveInstance()`

src/main/java/com/cloudbees/jenkins/support/configfiles/OtherConfigFilesComponent.java
christ66: Use https://github.com/jenkinsci/support-core-plugin/blob/master/src/main/java/com/cloudbees/jenkins/support/util/Helper.java#L21 (`Helper.getActiveInstance()`)

src/main/java/com/cloudbees/jenkins/support/configfiles/ConfigFileComponent.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.Set;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+
+/**
+ * Support component to add config.xml to the support bundle.
+ */
+@Extension
+public class ConfigFileComponent extends Component {
```
christ66: :bug: `isSelectedByDefault` should be set to `false`

src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
```diff
@@ -0,0 +1,101 @@
...
+                            start = 0;
+                            length = ch.length;
+                        }
+                    }
+                }
+                super.characters(ch, start, length);
+            }
+        };
+        Source src = new SAXSource(xr, new InputSource(new StringReader(FileUtils.readFileToString(xmlFile))));
+        File patchedFile = File.createTempFile("patched", ".xml");
```
christ66: :bug: Should be `File.createTempFile("patched", ".xml", SupportPlugin.getRootDirectory());` as this will put the temporary file in the support directory. Without specifying the root directory it will place the files in the default `java.io.tmpdir` which some users might have access to, and not the `JENKINS_HOME` directory
:bug: the temp file should be deleted.

src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
christ66: Errors are bubbled up, so I think that's good.

varmenise:
waiting for https://github.com/jenkinsci/credentials-plugin/pulls to be merged and use that method instead of the local `isSecretBytes` method

amuniz:
:bee:

varmenise:
@stephenc @christ66 can I have your review after addressing your comments? Thanks!

christ66:
@varmenise https://github.com/jenkinsci/support-core-plugin/pull/83#discussion_r86581163 ?

stephenc:
So anyone who has used to "write secret to disk as plain text and Jenkins will turn to encrypt on next write" to populate their Jenkins home (e.g. See https://issues.jenkins-ci.org/browse/JENKINS-39317) will have those plaintext secrets exposed in the bundle.

I cannot see a good way around that however... is there some way we could add a help note to that effect on these new components?

varmenise:
@christ66 sorry I had missed that, done! @stephenc see if that note can work

pom.xml
```diff
@@ -67,11 +67,6 @@
     <tag>HEAD</tag>
   </scm>
 
-  <properties>
-    <jenkins.version>1.580.1</jenkins.version>
```
batmat: FWIW, this is bumping the baseline from 1.580.1 to 1.625.3, and JDK 7 instead of 6 previously. Intended?

varmenise: it's intended: otherwise it complains on some jenkins.getActiveInstance coming from credentials plugin 2.1.5 which does not exists in 1.580. I need the latest on credentials plugin

stephenc: More correctly, credentials 2.1.x requires 1.609 as the baseline, hence the need to jump

christ66:
I'm still 🐝 on this PR. The added jump to java 7 will make work a little easier in the future.

batmat:
Still :+1:/:bee: for me too, that jump is good thing anyway. And JDK 8 is coming, so time has come ;-).

batmat:
@stephenc your "review status" is still "requested changes", can you re-review or confirm you're :bee:ing so that @christ66 can merge it?

src/main/java/com/cloudbees/jenkins/support/configfiles/ConfigFileComponent.java
```diff
@@ -0,0 +1,73 @@
...
+    @NonNull
+    @Override
+    public Set<Permission> getRequiredPermissions() {
+        return Collections.singleton(Jenkins.ADMINISTER);
+    }
+
+    @NonNull
+    @Override
+    public String getDisplayName() {
+        return "Jenkins Global Configuration File (secrets are filtered out unless saved in clear in the config.xml file)";
```
stephenc: The text is grammatically incorrect.

`Encrypted secrets *are* redacted`

Is probably a safer and shorter statement

src/main/java/com/cloudbees/jenkins/support/configfiles/OtherConfigFilesComponent.java
```diff
@@ -0,0 +1,76 @@
...
+    @NonNull
+    @Override
+    public Set<Permission> getRequiredPermissions() {
+        return Collections.singleton(Jenkins.ADMINISTER);
+    }
+
+    @NonNull
+    @Override
+    public String getDisplayName() {
+        return "Other Jenkins Configuration Files (secrets are filtered out unless saved in clear in the .xml files)";
```
stephenc: Text is slightly better but I still think it is better to focus on the positive, e.g. We redact the encrypted secrets... which leaves unsaid but implies the non-redaction of "secrets" that are not redacted

varmenise:
@stephenc I gave up on trying to make *are* bold from jelly sorry. but the rest should be okey.

can someone also remove the work-in-progress label?

aheritier:
done @varmenise

batmat:
@reviewbybees done

PR #85, opened 2016-11-07
Add node monitoring data for all nodes.

christ66:
@reviewbybees


kuisathaverat:
🐝

aheritier:
🐝 LGTM with 2 remarks @christ66: 
* Wasn't it better/possible to directly add this in nodes.md ?
* You didn't add the file/license header (but not sure it is on all files...)

alecharp:
:bee:

christ66:
> Wasn't it better/possible to directly add this in nodes.md ?

Possibly, but the nodes.md file contains different types of data. nodes.md is more about configurations, while nodes-monitor.md is only about monitoring.

> You didn't add the file/license header (but not sure it is on all files...)

I created https://github.com/jenkinsci/support-core-plugin/pull/87 to address the license issue. No need to keep putting in the license header if it's never going to change.

PR #86, opened 2016-11-07
Check if Jenkins is quieting down.

christ66:
@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/BuildQueue.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/BuildQueue.java
@@ -101,6 +101,7 @@ public class BuildQueue extends Component {
               out.println("----");
               out.println();
             }
+            out.println("Is quieting down: " + Helper.getActiveInstance().isQuietingDown());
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
🐝

aheritier:
🐝

src/main/java/com/cloudbees/jenkins/support/impl/BuildQueue.java
```diff
@@ -101,6 +101,7 @@ public void writeTo(OutputStream os) throws IOException {
               out.println("----");
               out.println();
             }
+            out.println("Is quieting down: " + Helper.getActiveInstance().isQuietingDown());
```
carlosrodlop: 🐝 
Suggestion:
out.println("Is quieting down: " + Helper.getActiveInstance().isQuietingDown() + "new builds are not being executed");

jglick:
BTW you could take the opportunity to correct the missing information noted in [JENKINS-38514](https://issues.jenkins-ci.org/browse/JENKINS-38514).

PR #87, opened 2016-11-07
Add license file.

christ66:
All files in the support core plugin should be licensed under MIT (unless specified otherwise).

@reviewbybees esp @stephenc and @jglick


oleg-nenashev:
:bee:

PR #89, opened 2016-11-28
NPE in GCLogs.addContents

jglick:
Observed by just installing this plugin and restarting Jenkins.

```
… com.cloudbees.jenkins.support.SupportPlugin writeBundle
WARNING: Could not get content from 'Garbage Collection Logs' for support bundle
java.lang.NullPointerException
	at java.io.File.<init>(File.java:277)
	at com.cloudbees.jenkins.support.impl.GCLogs.addContents(GCLogs.java:84)
	at com.cloudbees.jenkins.support.SupportPlugin.writeBundle(SupportPlugin.java:311)
	at com.cloudbees.jenkins.support.SupportPlugin.writeBundle(SupportPlugin.java:274)
	at com.cloudbees.jenkins.support.SupportPlugin$PeriodicWorkImpl$1.run(SupportPlugin.java:724)
```

`Component.isEnabled` is ignored by `SupportPlugin.writeBundle(OutputStream)`. Possibly `GCLogs` should not even bother overriding it—just treat it as “enabled” but do nothing if `-Xloggc` is not currently passed.

@reviewbybees esp. @batmat

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/GCLogs.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/GCLogs.java
@@ -76,7 +76,9 @@ public class GCLogs extends Component {
     public void addContents(@NonNull Container result) {
         LOGGER.fine("Trying to gather GC logs for support bundle");
         String gcLogFileLocation = getGcLogFileLocation();
-        assert gcLogFileLocation != null; // non nullable here 'cause isEnabled() already checks it
+        if (gcLogFileLocation == null) {
+            return;
+        }
 
         if (isGcLogRotationConfigured()) {
             handleRotatedLogs(gcLogFileLocation, result);


svanoort:
🐝

batmat:
:bee: as this is quite an urgent fix, but shouldn't it be fixed also/instead upstream? I mean, is this expected that the component be called when it's supposed to be disabled?

batmat:
@reviewbybees done

jglick:
> this is quite an urgent fix

Well, it does not block startup or even bundle generation, but it is ugly.

> is this expected that the component be called when it's supposed to be disabled?

See the PR description. Only `GCLogs` overrides `isEnabled`, and I think it is probably wrong to do so—this is apparently only supposed to mean whether the checkbox is selected in the GUI, not whether the component _could_ be used if specifically requested (say, via the CLI).

batmat:
@christ66 you might want to re-release quickly the plugin with that fix. (Though looking at the code, this should be caught and just ignored, but not nice in the logs)

batmat:
> isEnabled, and I think it is probably wrong to do so—this is apparently only supposed to mean whether the checkbox is selected in the GUI, not whether the component could be used if specifically requested (say, via the CLI).

No, this makes the collector totally disappear from the UI. What you say is [`isSelectedByDefault()`](https://github.com/jenkinsci/support-core-plugin/blob/6b47348009fe53370e162695951ef70883e80d56/src/main/java/com/cloudbees/jenkins/support/api/Component.java#L102-L104)'s role.

jglick:
Either way, I suspect you could just delete the `isEnabled` override (maybe make it `final` in the base class?) without losing anything.

jglick:
> looking at the code, this should be caught and just ignored

Yes, see log excerpt in PR description.

aheritier:
:bee: whatever option you choose to do the fix.

batmat:
Could we please merge & release a fixed version? @christ66 or @jglick (as you are in the current persons allowed to release)

Thanks a lot

batmat:
@jglick To answer your comment about simply removing `isEnabled()`, ATM I think there's still value to leave the code as-is: when using the support bundle generation through the UI, this makes the option not even be shown. Hence I guess it's a bit less cognitive disruption currently.

jglick:
Released as 2.35. @batmat file a request to get yourself added to the releasers list for this plugin.

jglick:
> I think there's still value to leave the code as-is: when using the support bundle generation through the UI, this makes the option not even be shown.

No; it *is* shown, but in strikethrough text, with the completely misleading tooltip **Requires: Overall/Administer**.

PR #91, opened 2016-11-29
[JENKINS-40098] The support bundle naming strategy should be modifiable

batmat:
[JENKINS-40098](https://issues.jenkins-ci.org/browse/JENKINS-40098)

Introduced a change enabling both overriding through using a sysprop,
or implementing a new Extension Point.

The approach with Extension Points should enable more dynamic needs,
when the sysprop one will be immediately usable by anyone without
having to develop an extension.

Note: if found, the extension is preferred over the sysprop.

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/BundleNamePrefixProvider.java
+++ b/src/main/java/com/cloudbees/jenkins/support/BundleNamePrefixProvider.java
@@ -1,0 +1,67 @@
+package com.cloudbees.jenkins.support;
+
+import com.google.common.annotations.VisibleForTesting;
+import hudson.ExtensionList;
+import hudson.ExtensionPoint;
+
+import javax.annotation.Nonnull;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * <p>Extension point allowing to customize the support bundle name prefixing strategy.</p>
+ * It will work the following way:
+ * <ol>
+ * <li>If an implementation of {@link BundleNamePrefixProvider} is found, it will be used.<br>
+ * <strong>WARNING: </strong>if many are found, then a warning will be issued, and the first extension found will
+ * be used.</li>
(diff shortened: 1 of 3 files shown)


src/main/java/com/cloudbees/jenkins/support/BundleNamePrefixProvider.java
```diff
@@ -0,0 +1,67 @@
...
+        public String getPrefix() {
+            return System.getProperty(SUPPORT_BUNDLE_NAMING_PREFIX_PROPERTY, "support");
+        }
+    };
+
+    /* package */
+    static BundleNamePrefixProvider getInstance() {
+        final ExtensionList<BundleNamePrefixProvider> all = ExtensionList.lookup(BundleNamePrefixProvider.class);
+        final int extensionCount = all.size();
+        System.out.println(extensionCount);
```
amuniz: Delete or use a logger.

amuniz:
:bee:

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -713,7 +713,7 @@ public void run() {
                                 SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd_HH.mm.ss");
                                 dateFormat.setTimeZone(TimeZone.getTimeZone("GMT"));
 
-                                final String bundlePrefix = "support";
+                                final String bundlePrefix = BundleNamePrefixProvider.getInstance().getPrefix();
```
jglick: This needs to be reconciled somehow with the existing SPI, `SupportProvider.getName`.

batmat: Thanks for the heads-up, discovering it indeed. And BTW, in that case (the periodic auto generation) seems like it's ignored contrary to when using the UI. Digging...

batmat:
Preparing another PR that will supersede this one. I need to refactor the existing code too since the filename retrieval logic is not currently centralized.

PR #92, opened 2016-11-30
[JENKINS-40098] Bundle naming strategy should be able to specify an instance type

batmat:
[JENKINS-40098](https://issues.jenkins-ci.org/browse/JENKINS-40098)

Refactored the existing code to use the same logic for choosing the bundle filename, and introduced a change enabling both overriding through using a sysprop, or implementing a new Extension Point.

The approach with Extension Points should enable more dynamic needs, when the sysprop one will be immediately usable by anyone without having to develop an extension.

Note: if found, the extension is preferred over the sysprop.

@reviewbybees esp. @jglick who pointed me to SupportProvider.getName() I had missed in the first attempt. 

Supersedes #91

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/BundleNameInstanceTypeProvider.java
+++ b/src/main/java/com/cloudbees/jenkins/support/BundleNameInstanceTypeProvider.java
@@ -26,11 +26,11 @@ package com.cloudbees.jenkins.support;
 
 import com.cloudbees.jenkins.support.api.SupportProvider;
 import com.google.common.annotations.VisibleForTesting;
+import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.Extension;
 import hudson.ExtensionList;
 import hudson.ExtensionPoint;
 
-import javax.annotation.Nonnull;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -57,7 +57,7 @@ public abstract class BundleNameInstanceTypeProvider implements ExtensionPoint {
 
     private static final Logger LOGGER = Logger.getLogger(BundleNameInstanceTypeProvider.class.getName());
 
-    @Nonnull
+    @NonNull
     static BundleNameInstanceTypeProvider getInstance() {
         final ExtensionList<BundleNameInstanceTypeProvider> all = ExtensionList.lookup(BundleNameInstanceTypeProvider.class);
         final int extensionCount = all.size();
@@ -91,7 +91,7 @@ public abstract class BundleNameInstanceTypeProvider implements ExtensionPoint {
      *
      * @return the instance type specification to be used for generated support bundles.
      */
-    @Nonnull
+    @NonNull
     public abstract String getInstanceType();
 
     // We want this to be always picked up last in case others are found Double.MIN_VALUE does NOT work
--- a/src/main/java/com/cloudbees/jenkins/support/BundleNameInstanceTypeProvider.java
+++ b/src/main/java/com/cloudbees/jenkins/support/BundleNameInstanceTypeProvider.java
@@ -75,7 +75,7 @@ public abstract class BundleNameInstanceTypeProvider implements ExtensionPoint {
 
         final BundleNameInstanceTypeProvider chosen = all.get(0);
         if (extensionCount > 1) {
-            LOGGER.log(Level.INFO, "Using {0} as BundleNameInstanceTypeProvider implementation", chosen.getClass().getName());
+            LOGGER.log(Level.FINE, "Using {0} as BundleNameInstanceTypeProvider implementation", chosen.getClass().getName());
         }
         return chosen;
     }
--- a/src/main/java/com/cloudbees/jenkins/support/BundleNameInstanceTypeProvider.java
+++ b/src/main/java/com/cloudbees/jenkins/support/BundleNameInstanceTypeProvider.java
@@ -1,0 +1,78 @@
+package com.cloudbees.jenkins.support;
+
+import com.cloudbees.jenkins.support.api.SupportProvider;
+import com.google.common.annotations.VisibleForTesting;
+import hudson.ExtensionList;
+import hudson.ExtensionPoint;
+
+import javax.annotation.Nonnull;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * <p>Extension point allowing to customize the support bundle naming strategy.</p>
+ * <p>
+ * It will work the following way:
+ * </p>
+ * <ol>
+ * <li>If an implementation of {@link BundleNameInstanceTypeProvider} is found, it will be used.<br>
+ * <strong>WARNING: </strong>if many are found, then a warning will be issued, and the first extension found will
+ * be used.</li>
+ * <li>If not, then it will check for the presence of the {@link #SUPPORT_BUNDLE_NAMING_INSTANCE_SPEC_PROPERTY}
+ * system property, and will use its value if provided.</li>
+ * <li>If not, then will fallback to the original behaviour, which is simply an empty String</li>
+ * </ol>
+ *
+ * @see SupportProvider#getName() for prefixing.
+ */
+public abstract class BundleNameInstanceTypeProvider implements ExtensionPoint {
+
+    @VisibleForTesting
+    static final String SUPPORT_BUNDLE_NAMING_INSTANCE_SPEC_PROPERTY = SupportPlugin.class.getName() + ".instanceType";
+
+    private static final Logger LOGGER = Logger.getLogger(BundleNameInstanceTypeProvider.class.getName());
+
+    private static final BundleNameInstanceTypeProvider DEFAULT_STRATEGY = new BundleNameInstanceTypeProvider() {
+
+        @Override
+        public String getInstanceType() {
+            return System.getProperty(SUPPORT_BUNDLE_NAMING_INSTANCE_SPEC_PROPERTY, "");
+        }
+    };
+
+    /* package */
+    @Nonnull
+    static BundleNameInstanceTypeProvider getInstance() {
+        final ExtensionList<BundleNameInstanceTypeProvider> all = ExtensionList.lookup(BundleNameInstanceTypeProvider.class);
+        final int extensionCount = all.size();
+        if (extensionCount < 1) {
+            LOGGER.fine("No alternative strategy provided for support bundle prefixing.");
+            return DEFAULT_STRATEGY;
+        }
+
+        if (all.size() > 1) {
+            if (LOGGER.isLoggable(Level.WARNING)) {
+                LOGGER.warning("{0} implementations found for support bundle prefix naming strategy. " +
+                        "Can be only 0 or 1. Choosing the first found.");
+                for (BundleNameInstanceTypeProvider bundlePrefixProvider : all) {
+                    LOGGER.log(Level.WARNING, "class '{0}' found", bundlePrefixProvider.getClass().getName());
+                }
+            }
+        }
+        return all.get(0);
+    }
+
+    /**
+     * Returns the <strong>non-null</strong> instance type to be used for generated support bundle names.
+     * Aims to provide informational data about the generated bundles.
+     *
+     * <p>
+     * <p><b>Will be used for file name generation, so avoid funky characters.
+     * Please stay in <code>[a-zA-Z-_]</code></b>. Also consider the file name length, you probably want to be defensive
+     * and not return crazily long strings. Something below 20 characters or so might sound reasonable.</p>
+     *
+     * @return the instance type specification to be used for generated support bundles.
+     */
+    @Nonnull
+    public abstract String getInstanceType();
+}
--- a/src/main/java/com/cloudbees/jenkins/support/BundleNameInstanceTypeProvider.java
+++ b/src/main/java/com/cloudbees/jenkins/support/BundleNameInstanceTypeProvider.java
@@ -97,7 +97,7 @@ public abstract class BundleNameInstanceTypeProvider implements ExtensionPoint {
     public abstract String getInstanceType();
 
     // We want this to be always picked up last in case others are found Double.MIN_VALUE does NOT work
-    @Extension(ordinal = -1000000)
+    @Extension(ordinal = Integer.MIN_VALUE)
     public static final class DEFAULT_STRATEGY extends BundleNameInstanceTypeProvider {
 
         @Override
--- a/src/main/java/com/cloudbees/jenkins/support/BundleNameInstanceTypeProvider.java
+++ b/src/main/java/com/cloudbees/jenkins/support/BundleNameInstanceTypeProvider.java
@@ -81,12 +81,14 @@ public abstract class BundleNameInstanceTypeProvider implements ExtensionPoint {
     }
 
     /**
-     * Returns the <strong>non-null</strong> instance type to be used for generated support bundle names.
-     * Aims to provide informational data about the generated bundles.
-     * <p>
-     * <p>
-     * <p><b>Will be used for file name generation, so avoid funky characters.
-     * Please stay in <code>[a-zA-Z-_]</code></b>. Also consider the file name length, you probably want to be defensive
+     * Returns the <strong>non-null</strong> and non empty (default value is empty) instance type to be used for
+     * generated support bundle names.
+     *
+     * <p>Aims to provide informational data about the generated bundles.</p>
+     * <p><strong>Will be used for file name generation, so avoid funky characters.
+     * Please ideally stay in <code>[a-zA-Z-_.]</code></strong>.
+     *
+     * Also consider the file name length, you probably want to be defensive
      * and not return crazily long strings. Something below 20 characters or so might sound reasonable.</p>
      *
      * @return the instance type specification to be used for generated support bundles.
--- a/src/main/java/com/cloudbees/jenkins/support/BundleNameInstanceTypeProvider.java
+++ b/src/main/java/com/cloudbees/jenkins/support/BundleNameInstanceTypeProvider.java
@@ -1,7 +1,32 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
 package com.cloudbees.jenkins.support;
 
 import com.cloudbees.jenkins.support.api.SupportProvider;
 import com.google.common.annotations.VisibleForTesting;
+import hudson.Extension;
 import hudson.ExtensionList;
 import hudson.ExtensionPoint;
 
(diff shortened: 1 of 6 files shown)


src/main/java/com/cloudbees/jenkins/support/BundleNameInstanceTypeProvider.java
```diff
@@ -0,0 +1,78 @@
+package com.cloudbees.jenkins.support;
```
jtnord: (c) header please

src/main/java/com/cloudbees/jenkins/support/BundleNameInstanceTypeProvider.java
```diff
@@ -0,0 +1,78 @@
...
+        final ExtensionList<BundleNameInstanceTypeProvider> all = ExtensionList.lookup(BundleNameInstanceTypeProvider.class);
+        final int extensionCount = all.size();
+        if (extensionCount < 1) {
+            LOGGER.fine("No alternative strategy provided for support bundle prefixing.");
+            return DEFAULT_STRATEGY;
+        }
+
+        if (all.size() > 1) {
+            if (LOGGER.isLoggable(Level.WARNING)) {
+                LOGGER.warning("{0} implementations found for support bundle prefix naming strategy. " +
```
jtnord: :bug: you are missing a parameter here.
NIT: from a stylistic point I think its better to always use Logger.log(Level.XXX ...) rather than the Logger.xxx(...)

src/main/java/com/cloudbees/jenkins/support/BundleNameInstanceTypeProvider.java
```diff
@@ -0,0 +1,78 @@
...
+    @Nonnull
+    static BundleNameInstanceTypeProvider getInstance() {
+        final ExtensionList<BundleNameInstanceTypeProvider> all = ExtensionList.lookup(BundleNameInstanceTypeProvider.class);
+        final int extensionCount = all.size();
+        if (extensionCount < 1) {
+            LOGGER.fine("No alternative strategy provided for support bundle prefixing.");
+            return DEFAULT_STRATEGY;
+        }
+
+        if (all.size() > 1) {
```
jtnord: why not use `extensionCount`

src/main/java/com/cloudbees/jenkins/support/BundleNameInstanceTypeProvider.java
```diff
@@ -0,0 +1,78 @@
...
+     * Aims to provide informational data about the generated bundles.
+     *
+     * <p>
+     * <p><b>Will be used for file name generation, so avoid funky characters.
+     * Please stay in <code>[a-zA-Z-_]</code></b>. Also consider the file name length, you probably want to be defensive
+     * and not return crazily long strings. Something below 20 characters or so might sound reasonable.</p>
+     *
+     * @return the instance type specification to be used for generated support bundles.
+     */
+    @Nonnull
```
jtnord: is this really the annotations that this code already uses.... If not then swtich.

src/test/java/com/cloudbees/jenkins/support/BundleNamePrefixTest.java
```diff
@@ -0,0 +1,71 @@
...
+    public JenkinsRule rule = new JenkinsRule();
+
+    @Test
+    public void checkOriginalBehaviour() throws Exception {
+        assertThat(SupportPlugin.getBundleFileName(), startsWith("support_" + CURRENT_YEAR));
+    }
+
+    @Test
+    public void checkWithOneProvider() throws Exception {
+        assertThat(SupportPlugin.getBundleFileName(), startsWith("support_pouet_" + CURRENT_YEAR));
```
amuniz: Oh, ok, I see the method name match with the `@TestExtension` parameter.

src/main/java/com/cloudbees/jenkins/support/BundleNameInstanceTypeProvider.java
```diff
@@ -0,0 +1,78 @@
...
+ * @see SupportProvider#getName() for prefixing.
+ */
+public abstract class BundleNameInstanceTypeProvider implements ExtensionPoint {
+
+    @VisibleForTesting
+    static final String SUPPORT_BUNDLE_NAMING_INSTANCE_SPEC_PROPERTY = SupportPlugin.class.getName() + ".instanceType";
+
+    private static final Logger LOGGER = Logger.getLogger(BundleNameInstanceTypeProvider.class.getName());
+
+    private static final BundleNameInstanceTypeProvider DEFAULT_STRATEGY = new BundleNameInstanceTypeProvider() {
```
jtnord: why not have this as an extension with very low priority and avoid the extra complexity below?

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -763,6 +774,33 @@ public boolean accept(File dir, String name) {
             }
         }
 
+        private void cleanupOldBundles(File bundleDir, File justGenerated) {
+            thread.setName(String.format("%s periodic bundle generator: tidying old bundles since %s",
+                    SupportPlugin.class.getSimpleName(), new Date()));
+            File[] files = bundleDir.listFiles(new FilenameFilter() {
+                public boolean accept(File dir, String name) {
+                    return name.startsWith(SupportPlugin.getBundlePrefix()) && name.endsWith(".zip");
```
jtnord: this will fail to clean up old bundles if the provider is changed.

Why not just check for "support*.zip"?

src/main/java/com/cloudbees/jenkins/support/BundleNameInstanceTypeProvider.java
```diff
@@ -0,0 +1,78 @@
...
+            LOGGER.fine("No alternative strategy provided for support bundle prefixing.");
+            return DEFAULT_STRATEGY;
+        }
+
+        if (all.size() > 1) {
+            if (LOGGER.isLoggable(Level.WARNING)) {
+                LOGGER.warning("{0} implementations found for support bundle prefix naming strategy. " +
+                        "Can be only 0 or 1. Choosing the first found.");
+                for (BundleNameInstanceTypeProvider bundlePrefixProvider : all) {
+                    LOGGER.log(Level.WARNING, "class '{0}' found", bundlePrefixProvider.getClass().getName());
```
jtnord: :bug:  `'` needs to be escaped other wise this will log `"class {0} found`"
https://docs.oracle.com/javase/7/docs/api/index.html?java/text/MessageFormat.html

src/main/java/com/cloudbees/jenkins/support/BundleNameInstanceTypeProvider.java
batmat: Just because I wasn't absolutely sure about the behaviour of the priority, going to check, and if that works and the doc confirms, will do.

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
batmat: Because the SupportProvider.getName() could make that pattern totally different anyway, files could be named `whatever-thedate.zip`.

A different approach might be to just select every zip files like `$JENKINS_HOME/support/*.zip`, but this is a bit risky if someone has taken the habit to put unrelated files there...

batmat:
@jtnord I believe I addressed all your comments, but https://github.com/jenkinsci/support-core-plugin/pull/92#discussion_r90266258. For that one, I guess the only "risk" is to have uncleaned lefovers that users may have to clean up manually. The alternative is to work more largely, but we take then IMO an even more risky approach: wiping out unrelated zip files. WDYT?

If you have a better idea or think this is fine to just consider nobody should have put things in that directory (which, well, I would agree on...) then we can make it simple(r).

amuniz:
:bee:

src/main/java/com/cloudbees/jenkins/support/BundleNameInstanceTypeProvider.java
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+ * <p>Extension point allowing to customize the support bundle naming strategy.</p>
+ * <p>
+ * It will work the following way:
+ * </p>
+ * <ol>
+ * <li>If an implementation of {@link BundleNameInstanceTypeProvider} is found, it will be used.<br>
+ * <strong>WARNING: </strong>if many are found, then a warning will be issued, and the first extension found will
+ * be used.</li>
```
jglick: So, where is this nondefault implementation? For any newly introduced API I expect to see a PR from another plugin repository using a `SNAPSHOT` dependency which consumes that API. (Obviously public use cases are preferred whenever possible, but note that a PR in a private repository may refer in its summary etc. to a public PR, in which case GitHub automatically creates a backlink in the public PR visible to users who have permission to view the private repo.)

src/main/java/com/cloudbees/jenkins/support/BundleNameInstanceTypeProvider.java
```diff
@@ -0,0 +1,106 @@
...
+     * Please stay in <code>[a-zA-Z-_]</code></b>. Also consider the file name length, you probably want to be defensive
+     * and not return crazily long strings. Something below 20 characters or so might sound reasonable.</p>
+     *
+     * @return the instance type specification to be used for generated support bundles.
+     */
+    @NonNull
+    public abstract String getInstanceType();
+
+    // We want this to be always picked up last in case others are found Double.MIN_VALUE does NOT work
+    @Extension(ordinal = -1000000)
```
christ66: `@Extension(ordinal = Integer.MAX_VALUE)`

batmat: Well, why not, but this is still weird, since the type is `double` anyway...

src/main/java/com/cloudbees/jenkins/support/SupportCommand.java
```diff
@@ -107,7 +98,7 @@ protected int run() throws Exception {
         }
 
         public OutputStream call() throws IOException {
-            File f = File.createTempFile(filename, ".zip");
+            File f = File.createTempFile(filename, "");
```
jglick: This is wrong. Even if `filename` ends in `.zip`, you will end up with something like this:

```
$ jrunscript -e 'println(java.io.File.createTempFile("foo.zip", ""))'
/tmp/foo.zip6023275810588304403
```

src/main/java/com/cloudbees/jenkins/support/SupportCommand.java
```diff
@@ -107,7 +98,7 @@ protected int run() throws Exception {
...
             System.err.println("Creating: " + f);
             return new RemoteOutputStream(new FileOutputStream(f));
         }
```
jglick: BTW not introduced in this PR, but while you are here please delete the bogus `checkRoles` implementation and extend `MasterToSlaveCallable`.

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -563,6 +561,46 @@ public void run() {
...
+    @NonNull
+    public static String getBundleFileName() {
+        StringBuilder filename = new StringBuilder();
+        filename.append(getBundlePrefix());
+
+        SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd_HH.mm.ss");
+        dateFormat.setTimeZone(TimeZone.getTimeZone("GMT"));
+        filename.append("_").append(dateFormat.format(new Date()));
+
+        filename.append(".zip");
```
jglick: As noted above, better to have the extension appended by the caller.

src/main/java/com/cloudbees/jenkins/support/BundleNameInstanceTypeProvider.java
```diff
@@ -0,0 +1,106 @@
...
+
+        final BundleNameInstanceTypeProvider chosen = all.get(0);
+        if (extensionCount > 1) {
+            LOGGER.log(Level.INFO, "Using {0} as BundleNameInstanceTypeProvider implementation", chosen.getClass().getName());
+        }
+        return chosen;
+    }
+
+    /**
+     * Returns the <strong>non-null</strong> instance type to be used for generated support bundle names.
```
jglick: You need to document that the empty string is allowed, and in fact the default.

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -563,6 +561,46 @@ public void run() {
...
+     * @return the prefix of the bundle name.
+     */
+    private static String getBundlePrefix() {
+        String filename = "support"; // default bundle filename
+        final SupportPlugin instance = getInstance();
+        if (instance != null) {
+            SupportProvider supportProvider = instance.getSupportProvider();
+            if (supportProvider != null) {
+                // let the provider name it
+                filename = supportProvider.getName();
```
jglick: Still leaves me wondering about why we have both systems. In the case of a vendor Acme implementing both extension points in a Blaster product, we will get something like

    acme-support_acme-blaster_2016-12-07_16.12.47.zip

which seems redundant.

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
jglick: > if someone has taken the habit to put unrelated files there

Well don’t do that then.

I agree it would be better to just delete all `*.zip`.

src/main/java/com/cloudbees/jenkins/support/SupportCommand.java
jglick: Though if we are always adding the current date down to second resolution anyway, then perhaps we can omit the generated suffix and just save `${java.io.tmpdir}/${filename}`.

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
batmat: Well, in that case, that vendor would just want to not put Acme everywhere. Which would be more:

`acme-support_blaster_2016-12-07_16.12.47.zip`

which seems less redundant than your example?

src/main/java/com/cloudbees/jenkins/support/BundleNameInstanceTypeProvider.java
```diff
@@ -0,0 +1,106 @@
...
+                        "Choosing the first found among the following:", extensionCount);
+                for (BundleNameInstanceTypeProvider nameProvider : all) {
+                    LOGGER.log(Level.WARNING, "Class {0} found", nameProvider.getClass().getName());
+                }
+            }
+        }
+
+        final BundleNameInstanceTypeProvider chosen = all.get(0);
+        if (extensionCount > 1) {
+            LOGGER.log(Level.INFO, "Using {0} as BundleNameInstanceTypeProvider implementation", chosen.getClass().getName());
```
batmat: fixed

src/main/java/com/cloudbees/jenkins/support/BundleNameInstanceTypeProvider.java
batmat: fixed. I actually explained it's expected to be not empty (contrary to the default value)

src/main/java/com/cloudbees/jenkins/support/BundleNameInstanceTypeProvider.java
```diff
@@ -0,0 +1,106 @@
...
+        }
+        return chosen;
+    }
+
+    /**
+     * Returns the <strong>non-null</strong> instance type to be used for generated support bundle names.
+     * Aims to provide informational data about the generated bundles.
+     * <p>
+     * <p>
+     * <p><b>Will be used for file name generation, so avoid funky characters.
```
batmat: fixed

src/main/java/com/cloudbees/jenkins/support/BundleNameInstanceTypeProvider.java
```diff
@@ -0,0 +1,106 @@
...
+     * Please stay in <code>[a-zA-Z-_]</code></b>. Also consider the file name length, you probably want to be defensive
```
batmat: fixed

src/main/java/com/cloudbees/jenkins/support/BundleNameInstanceTypeProvider.java
batmat: fixed, though I don't really like it. But in the end with the comment above, ppl puzzled by it should understand.

src/main/java/com/cloudbees/jenkins/support/SupportCommand.java
batmat: done

src/main/java/com/cloudbees/jenkins/support/SupportCommand.java
batmat: Done. I settled on using 

```java
Path path = Files.createFile(Paths.get(System.getProperty("java.io.tmpDir"), filename));
```

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
batmat: so, finally used `filename` directly instead of a temp file created by the dedicated API.

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
batmat: done

src/main/java/com/cloudbees/jenkins/support/SupportCommand.java
```diff
@@ -99,23 +91,18 @@ protected int run() throws Exception {
...
-            return new RemoteOutputStream(new FileOutputStream(f));
-        }
-
-        /** {@inheritDoc} */
         @Override
-        public void checkRoles(RoleChecker checker) throws SecurityException {
-            // TODO: do we have to verify some role?
+        public OutputStream call() throws IOException {
+            Path path = Files.createFile(Paths.get(System.getProperty("java.io.tmpDir"), filename));
+            System.err.println("Creating: " + path);
```
christ66: 🐛 `System.err`

christ66: Ah wait this is for the SupportCommand, nvm

src/main/java/com/cloudbees/jenkins/support/SupportCommand.java
```diff
@@ -90,23 +91,18 @@ protected int run() throws Exception {
...
-            System.err.println("Creating: " + f);
-            return new RemoteOutputStream(new FileOutputStream(f));
-        }
-
-        /** {@inheritDoc} */
         @Override
-        public void checkRoles(RoleChecker checker) throws SecurityException {
-            // TODO: do we have to verify some role?
+        public OutputStream call() throws IOException {
+            Path path = Files.createFile(Paths.get(System.getProperty("java.io.tmpDir"), filename));
```
jglick: It is `tmpdir` not `tmpDir`. Did you try running this command?

src/main/java/com/cloudbees/jenkins/support/BundleNameInstanceTypeProvider.java
```diff
@@ -0,0 +1,108 @@
...
+
+    private static final Logger LOGGER = Logger.getLogger(BundleNameInstanceTypeProvider.class.getName());
+
+    @NonNull
+    static BundleNameInstanceTypeProvider getInstance() {
+        final ExtensionList<BundleNameInstanceTypeProvider> all = ExtensionList.lookup(BundleNameInstanceTypeProvider.class);
+        final int extensionCount = all.size();
+
+        if (extensionCount > 2) {
+            if (LOGGER.isLoggable(Level.WARNING)) {
```
jglick: We log `INFO` by default. So no, it is not terribly useful, but harmless I suppose.

batmat:
Fixed @jglick yes I tried it locally using jrunscript, but didn't adjusted after fixing the jrunscript. Sorry for the noise.

jglick:
> I tried it locally using jrunscript

Not sure what you mean. You need to

    mvn hpi:run &
    wget -O /tmp/cli.jar http://localhost:8080/jenkins/jnlpJars/jenkins-cli.jar
    java -jar /tmp/cli.jar -s http://localhost:8080/jenkins/ support

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -763,6 +770,38 @@ public boolean accept(File dir, String name) {
...
+                    SupportPlugin.class.getSimpleName(), new Date()));
+            File[] files = bundleDir.listFiles(new FilenameFilter() {
+                public boolean accept(File dir, String name) {
+                    return name.endsWith(".zip");
+                }
+            });
+            long pivot = System.currentTimeMillis();
+            for (long l = 1; l * 2 > 0; l *= 2) {
+                boolean seen = false;
+                for (File f : files) {
```
jglick: NPE if `listFiles` returns null, as it is permitted to do even if you believe it currently exists.

batmat: @jglick I just moved that code in a dedicated method to clarify the code, didn't touch it. But will happily file a separate PR about it.
Thanks for the review !

christ66:
@batmat @jglick @jtnord I am wondering if this PR should be reverted. This only makes the naming of the support bundles significantly longer.

Before this PR my support bundles use to be:

```
cloudbees-support_2017-02-02_18.16.45
```

which is already long enough and unique (as in the time it was created IMO is random and unique enough), but now they are:

```
cloudbees-support_UNKNOWN-localhost_8080_jenkins_2017-01-24_20.02.19
```

Having this long of a name for a folder I do not believe is right. Just a lot of extra noise which does not need to exist in the name.

jtnord:
not sure why that comes up as `UNKNOWN` but the name is really really really useful when you have lots of bundles from the entire cluster in a support ticket that are updated at various occasions during the lifecycle...
The fact it says` cloudbees-support` would possibly imply you are running a cloudbees distribution without some required plugins that should be installed by default on every instance?

PR #94, opened 2016-12-06
Simply ignore if the GC logging options are not enabled

batmat:
@reviewbybees esp. @jglick as you rightly advised for that change in https://github.com/jenkinsci/support-core-plugin/pull/89#issuecomment-264277965

State currently:
image


raul-arabaolaza:
AFAIU :bee:

oleg-nenashev:
OK 🐝

batmat:
@reviewbybees done

PR #97, opened 2016-12-13
Max file size for agent logs.

christ66:
@reviewbybees


src/main/java/com/cloudbees/jenkins/support/impl/SlaveLaunchLogs.java
```diff
@@ -49,6 +49,11 @@
  */
 public class SlaveLaunchLogs extends Component{
 
+    /**
+     * Max agent file size of 500mb.
+     */
+    public /*static final*/ long max_file_size = Long.getLong(SlaveLaunchLogs.class.getCanonicalName() + ".max_file_size", 5 * 100000);
```
batmat: :bug: 500 * 1000 * 1000 is probably what you want reading the Javadoc

christ66:
Decided to use the `FileListCapComponent.MAX_FILE_SIZE` which is 2mb by default. Logs shouldn't exceed 2mb.

PR #101, opened 2017-02-03
Use UTC for all timestamps

jglick:
When trying to evaluate a complex support bundle, or series of bundles, perhaps attempting to align different log files to determine the root cause of a single event, possibly crossing machines…it is very tricky to keep track of multiple local timezones plus UTC. This patch just forces the support bundle to use UTC consistently: most importantly, in log timestamps, and in the name of the bundle ZIP itself.

@reviewbybees


aheritier:
LGTM 🐝

christ66:
LGTM also :bee:

PR #102, opened 2017-02-04
[JENKINS-26731] Convert support bundle to use YAML

christ66:
This is an initial POC to see if it is possible to convert the support bundle files from markdown to YAML. YAML is just as easy to read as markdown, however it provides an easy method for converting back to java objects. More files should be converted as time goes on.

Also there are some file renames, and optimizations happening which will break backward compatibility. I do not believe anyone is actually using the support core plugin in their dependency list (from a quick github search), so these changes should be safe to implement.

@reviewbybees


michaelmccaskill:
🐝  Looks good to me.

pom.xml
```diff
@@ -67,6 +67,11 @@
     <tag>HEAD</tag>
   </scm>
 
+  <properties>
+    <jenkins.version>1.625</jenkins.version>
+    <java.level>7</java.level>
```
jglick: This is the default.

pom.xml
```diff
@@ -87,11 +92,16 @@
       <artifactId>uadetector-resources</artifactId>
       <version>2013.09</version>
     </dependency>
+    <dependency>
+      <groupId>org.projectlombok</groupId>
+      <artifactId>lombok</artifactId>
```
jglick: RUN AWAY IN FEAR

src/main/java/com/cloudbees/jenkins/support/impl/AboutBrowserComponent.java
```diff
@@ -0,0 +1,100 @@
...
+public class AboutBrowserComponent extends Component {
+    @Override
+    @NonNull
+    public String getDisplayName() {
+        return "About browser";
+    }
+
+    @Override
+    public Set<Permission> getRequiredPermissions() {
+        return Collections.emptySet();
```
jglick: Why does this not show up as a rename in the diff?

src/main/java/com/cloudbees/jenkins/support/model/BuildQueue.java
```diff
@@ -0,0 +1,49 @@
+package com.cloudbees.jenkins.support.model;
+
+import lombok.Data;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Created by stevenchristou on 2/3/17.
```
jglick: Configure your IDE to not add these useless comments.

src/main/java/com/cloudbees/jenkins/support/impl/AdministrativeMonitorsComponent.java
```diff
@@ -60,62 +62,56 @@
...
+                AdminMonitors.OldDataMonitor olddm = new AdminMonitors.OldDataMonitor();
+                OldDataMonitor odm = (OldDataMonitor) monitor;
+                for (Map.Entry<Saveable,OldDataMonitor.VersionRange> entry : odm.getData().entrySet()) {
+
+                    AdminMonitors.OldDataMonitor.OldData data = new AdminMonitors.OldDataMonitor.OldData();
+                    data.setProblematicObject(entry.getKey().toString());
+
+                    OldDataMonitor.VersionRange value = entry.getValue();
+                    data.setRange(value.toString());
+                    data.setExtra(value.extra);
```
jglick: Missing the `StringUtils.isBlank` call from the original.

I suppose `YAMLFactory` will now take care of the possibility of a multiline stack trace.

jglick:
Some legitimate FindBugs warnings; for example, `UserAgent` must implement `Serializable`.

pom.xml
christ66: :rofl:

stephenc: Oh my god! Please DO NOT USE LOMBOK!

src/main/java/com/cloudbees/jenkins/support/impl/AboutBrowserComponent.java
christ66: Not sure. Nothing has changed there.

christ66:
@stephenc @jglick @batmat please re-review as I've removed lombok, and fixed a lot more issues.

christ66:
@stephenc not sure if you implemented this, but do you know why methods like: https://github.com/jenkinsci/support-core-plugin/blob/f819d772483a9a48b6d06dd7f011f45c6b7838f6/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java#L490 are synchronized? From initially looking at the code I don't believe synchronized is necessary, but I could be wrong.

stephenc:
If you have experience computing cumulative running total stats, you want them updated atomically or you get invalid answers. 

I may have had/still have the computation parallelised so no guarantees as to single thread.

If used by a single thread only, JVM will be almost as fast as without... and if you are seeing contention then not safe to remove.

PR #103, opened 2017-02-17
Add reportID to bundle reports in about.md

batmat:
Can sometimes cause some headache to have a simple way to uniquely refer to a bundle when discussing them.
This change aims to resolve this issue.

Generates `about.md` this way now:

```
Jenkins
=======
Report ID: 824d2ab4d1bb1fba69bee1e2cbd894ce-2017-02-17T09:49:22Z-731
...
```

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
@@ -65,7 +65,9 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
+import java.util.Random;
 import java.util.Set;
+import java.util.TimeZone;
 import java.util.TreeMap;
 import java.util.TreeSet;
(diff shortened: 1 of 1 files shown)


kuisathaverat:
LGTM 🐝

jglick:
We already include a timestamp in the filename. Why not include a (possibly truncated) version of the instance ID as well, and then you can simply refer to the filename? That seems a lot more straightforward to look up rather than having to open a series of ZIP files and search for this ID buried inside `about.md`.

christ66:
🐛 I agree with @jglick.

If we're really trying to make each bundle unique and easily readable we could change to a naming scheme to use a "cute" naming scheme like: https://github.com/kohsuke/wordnet-random-name (which I believe someone suggested a long time ago, but don't remember who). It would make it easier to distinguish between different support bundles.

batmat:
Closing as per comments above. And the addition I did some months ago already makes the naming of the generated bundle far more uniquely identifiable.

PR #104, opened 2017-02-21
Delete even if write to bundle fails.

christ66:
If the `writeTo` method fails we still should perform the delete no matter what happens.

@reviewbybees esp. @varmenise


batmat:
And @amuniz

christ66:
Thanks @batmat and @jglick

PR #106, opened 2017-03-01
[JENKINS-42393] Temporary files should not be created in the bundles root directory under JENKINS_HOME

batmat:
[JENKINS-42393](https://issues.jenkins-ci.org/browse/JENKINS-42393)

Realigning File.createTempFile call to others already existing in the support-core plugin codebase. And avoid unnecessary network sync for temporary files in the go.

@reviewbybees esp. @varmenise @cyrille-leclerc @christ66

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
+++ b/src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
@@ -16,8 +16,8 @@ import javax.xml.transform.TransformerException;
 import javax.xml.transform.TransformerFactory;
 import javax.xml.transform.sax.SAXSource;
 import javax.xml.transform.stream.StreamResult;
+import java.io.ByteArrayOutputStream;
 import java.io.File;
-import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.StringReader;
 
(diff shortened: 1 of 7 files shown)


src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
```diff
@@ -78,7 +77,7 @@ public void characters(char[] ch, int start, int length) throws SAXException {
             }
         };
         Source src = new SAXSource(xr, new InputSource(new StringReader(FileUtils.readFileToString(xmlFile))));
-        File patchedFile = File.createTempFile("patched", ".xml", SupportPlugin.getRootDirectory());
+        File patchedFile = File.createTempFile("patched", ".xml");
```
jtnord: I doubt many setups configure the java tmp directory to be private, despite what you are calling out in the JIRA.
These files are created with the default umask - and although we attempt to patch them we may not - so it may leak secrets world readable.
so a :bug: on at least not restricting read access here.

src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
```diff
@@ -77,13 +77,13 @@ public void characters(char[] ch, int start, int length) throws SAXException {
...
-        Result res = new StreamResult(new FileOutputStream(patchedFile));
+        final ByteArrayOutputStream result = new ByteArrayOutputStream();
+        Result res = new StreamResult(result);
         Transformer transformer = TransformerFactory.newInstance().newTransformer();
         //omit xml declaration because of https://bugs.openjdk.java.net/browse/JDK-8035437
         transformer.setOutputProperty(OutputKeys.OMIT_XML_DECLARATION, "yes");
         transformer.transform(src, res);
 
-        return patchedFile;
+        return result.toString("UTF-8");
```
jtnord: is the factory guaranteed to write in `UTF-8`?
I think you need to be explicit by setting [encoding](https://docs.oracle.com/javase/7/docs/api/javax/xml/transform/OutputKeys.html#ENCODING)

alvarolobato:
Isn't it possible to use FileContent and override it to filter the file when writing? Even if you load the whole file in memory, that would allow you to only have one in memory vs having all of them.

batmat:
@alvarolobato I like this idea, probably the best compromise indeed.

batmat:
@alvarolobato just pushed an impl of your idea, thanks again!

@jtnord I think this should address your 1st concern: now no temporary intermediate file is written, only filtered in memory while writing the ZIP. Also addressed your valid concern about encoding  by sharing a constant between the two involved parts. 

Globally, now the 3 components are much simpler. They could most probably be simplified a bit more, but I don't want to mix too many subjects here. If so this will be done/doable in a future PR.

batmat:
@reviewbybees done

batmat:
@christ66 if you don't mind, I plan to release it in the next few hours if you're OK and nobody else complains. Thanks!

batmat:
support-core-2.39 is out

PR #108, opened 2017-03-16
Make Slow Requests an extension point.

christ66:
I would like `SlowRequests` to be an extension point, so we can add other checks in the event a slow request happens. We can add checks like I/O, networking, etc. when there is a slow request.

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/slowrequest/SlowRequest.java
+++ b/src/main/java/com/cloudbees/jenkins/support/slowrequest/SlowRequest.java
@@ -10,7 +10,7 @@ import java.util.List;
 public abstract class SlowRequest implements ExtensionPoint {
     abstract void doRun(InflightRequest req, long totalTime, FileListCap logs) throws IOException;
 
-    public List<SlowRequest> all() {
+    public static List<SlowRequest> all() {
         return ExtensionList.lookup(SlowRequest.class);
     }
 }
--- a/src/main/java/com/cloudbees/jenkins/support/slowrequest/SlowRequest.java
+++ b/src/main/java/com/cloudbees/jenkins/support/slowrequest/SlowRequest.java
@@ -1,0 +1,16 @@
+package com.cloudbees.jenkins.support.slowrequest;
+
+import com.cloudbees.jenkins.support.timer.FileListCap;
+import hudson.ExtensionList;
+import hudson.ExtensionPoint;
+
+import java.io.IOException;
+import java.util.List;
+
+public abstract class SlowRequest implements ExtensionPoint {
+    abstract void doRun(InflightRequest req, long totalTime, FileListCap logs) throws IOException;
+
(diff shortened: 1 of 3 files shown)


JTSweet:
🐝

jglick:
So where is the downstream PR showing this being used?

christ66:
@jglick nothing yet. Looking to implement it as an extension point first, then we'd need to figure out what needs to be executed during slow requests.

src/main/java/com/cloudbees/jenkins/support/slowrequest/SlowRequest.java
```diff
@@ -0,0 +1,16 @@
+package com.cloudbees.jenkins.support.slowrequest;
+
+import com.cloudbees.jenkins.support.timer.FileListCap;
+import hudson.ExtensionList;
+import hudson.ExtensionPoint;
+
+import java.io.IOException;
+import java.util.List;
+
+public abstract class SlowRequest implements ExtensionPoint {
```
oleg-nenashev: Needs to be documented 🐛

src/main/java/com/cloudbees/jenkins/support/slowrequest/SlowRequest.java
```diff
@@ -0,0 +1,16 @@
...
+    abstract void doRun(InflightRequest req, long totalTime, FileListCap logs) throws IOException;
+
+    public static List<SlowRequest> all() {
```
oleg-nenashev: Why not ExtensionList?

src/main/java/com/cloudbees/jenkins/support/slowrequest/SlowRequest.java
oleg-nenashev: I also doubt the method is ideal. Maybe it should be `AsyncMonitoringRequest` or whatever

src/main/java/com/cloudbees/jenkins/support/slowrequest/ThreadDumpSlowRequest.java
```diff
@@ -0,0 +1,69 @@
...
+@Extension
+public class ThreadDumpSlowRequest extends SlowRequest {
+
+    private final SimpleDateFormat format = new SimpleDateFormat("yyyyMMdd-HHmmss.SSS");
+    {
+        format.setTimeZone(TimeZone.getTimeZone("UTC"));
+    }
+
+    @Override
+    public void doRun(InflightRequest req, long totalTime, FileListCap logs) throws IOException {
```
oleg-nenashev: I am not sure what is purpose of moving `totalTime` in such way

jglick:
Definitely needs a downstream PR showing this being used for something.

christ66:
@jglick It is currently being used for `ThreadDumpSlowRequest` class in this pull request. Should I provide a second use case for the extension point?

PR #109, opened 2017-04-03
Suppress PeriodicWorkImpl during functional tests

jglick:
Otherwise, in plugins which depend on this one (like `workflow-cps`), every now and then the bundle generator will start to run while the test is running. Sometimes that is harmless (if a waste of time); sometimes it starts to run while Jenkins is shutting down, resulting in various stupid exceptions and the like being printed to the console. Harmless but confusing.

@reviewbybees


aheritier:
:bee:

PR #112, opened 2017-05-15
[JENKINS-44236] NFS stats related /proc files should also be gathered

batmat:
[JENKINS-44236](https://issues.jenkins-ci.org/browse/JENKINS-44236)

* [x] Harness existing code before changing/factorizing the code => https://github.com/jenkinsci/support-core-plugin/pull/112/commits/089f56902ea8a77b547a618d7bb969697bee87a1
* [x] Refactor `SystemConfiguration` and `JVMProcessSystemMetricsContents` with a base class as they contain a lot of common code => https://github.com/jenkinsci/support-core-plugin/pull/112/commits/9e1e7bdb3641e061d18b4619e100924797fe8c84
* [x] Also gather the nfs related files => https://github.com/jenkinsci/support-core-plugin/pull/112/commits/1021568723083b7cc3c4eb4154a4845d5861ebe6

@reviewbybees


src/main/java/com/cloudbees/jenkins/support/impl/ProcFilesRetriever.java
```diff
@@ -0,0 +1,109 @@
...
+
+        for (Map.Entry<String, String> procDescriptor : getFilesToRetrieve().entrySet()) {
+            container.add(new FileContent("nodes/" + name + "/proc/" + procDescriptor.getValue(),
+                                          new File(procDescriptor.getKey())));
+        }
+
+        addUnixContents(container, node, name);
+    }
+
+    protected void addUnixContents(@NonNull Container container, final @NonNull Node node, String name) {
```
jtnord: erm this is empty?

batmat: @jtnord clarified.

ydubreuil:
Not sure it makes sense to collect `/proc/net/rpc/nfs` and `/proc/net/rpc/nfsd` which are raw counters. They are only useful if collected at a regular interval and fed to a gauge or something like that.

Otherwise, I'm fine with the changes.

batmat:
@reviewbybees done, thanks

batmat:
Going to merge and release this. Thanks everyone!

batmat:
BTW @ydubreuil though definitely agreed this is currently very impractical, over time we will somehow have that regular collect process: using this version in the $JENKINS_HOME/support directory, you could extract a kind of trend from many bundles.

Now, what you said got me thinking that this could actually be improved to:
* parse the files you say should be collected regularly and pushed _somewhere_
* push those to the metrics-plugin

As the metrics are already retrieved through the [`Metrics`](https://github.com/jenkinsci/support-core-plugin/blob/master/src/main/java/com/cloudbees/jenkins/support/impl/Metrics.java) class, we'd then have those NFS stats over time almost for free. (then we'd need something to graph those easily, but that's another matter)

batmat:
Filed https://issues.jenkins-ci.org/browse/JENKINS-44534 as a followup. Will not be able to work on it myself in the short term at least, so anyone feel free to take it if you want/can.

Thanks

PR #113, opened 2017-05-29
JENKINS-42512 - "Slave" is replaced with Agent.

duemir:
Java elements and bundle structure is not touched for compatibility.
UI and comments are updated.

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
+++ b/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
@@ -621,7 +621,7 @@ public class SupportPlugin extends Plugin {
         }
 
         public Void call() {
-            // avoid double installation of the handler. JNLP slaves can reconnect to the master multiple times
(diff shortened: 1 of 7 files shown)


duemir:
@stephenc you mentioned as a maintainer on the wiki. It's minor straightforward change. Can I get a second bee and a merge? :)

stephenc:
CloudBees support took over. Long time since I cut a release. Suggest finding out off @aheritier who to ping

batmat:
Merging as this is a minor change and somehow already pre-agreed per Jenkins 2.x.

@stephenc do you think removing you from the pom.xml/developers would be good so that you don't appear anymore on the wiki macro as a maintainer?

aheritier:
@christ66 is still around if we need. CloudBees support didn't take over it even if Steven was the most active recently. All others contributors are welcomed.

christ66:
I'm still around 😄 .

Maybe we should create a Jenkinsfile which can perform a release after every merge request just in case I'm not around to complete it.

WRT the PR, 👍 .

PR #114, opened 2017-08-11
JENKINS-46132 - Download actual proc info from the Linux agents

duemir:
https://issues.jenkins-ci.org/browse/JENKINS-46132

FileContent which was using Master's files for all the agents is
replaced with FilePathContent which loads actual info via remoting.

Started phasing out the Active Instance Helper, since Jenkins Core
dependency is now greater than 1.609

@reviewbybees


duemir:
This will, unfortunately, disable it for both master and agents. 
I am not ready to lose the info from the master. It is very handy at judging the sizing in performance issue and double checking that arguments were added as suggested.

As a compromise, we can just not capture it from agents at all or split it into another component which will be disabled by default.

duemir:
@stephenc you seem to be an auther of the `FilePathContent` class. Do you mind providing feedback to the Owen's concern?

duemir:
@christ66 can you give your 2c about it?

src/main/java/com/cloudbees/jenkins/support/impl/ProcFilesRetriever.java
```diff
@@ -55,7 +53,7 @@ protected static String getNodeName(Node node) {
 
     @Override
     public void addContents(@NonNull Container container) {
-        Jenkins j = Helper.getActiveInstance();
+        Jenkins j = Jenkins.getActiveInstance();
```
christ66: :NIT: should be part of a different PR (I assume it's safe to use and should just get rid of the helper version and just use the one in the `Jenkins` class.

christ66:
If you would like to do this for `Master` configurations, but make it optional for the `Agent` configuration, my advice is just create 2 different versions of this.

src/main/java/com/cloudbees/jenkins/support/impl/ProcFilesRetriever.java
batmat: +1, and possibly more than just NIT because https://github.com/jenkinsci/jenkins/pull/3195... :)

src/main/java/com/cloudbees/jenkins/support/impl/ProcFilesRetriever.java
```diff
@@ -65,11 +63,11 @@ public void addContents(@NonNull Container container) {
 
     protected void addUnixContents(@NonNull Container container, final @NonNull Node node) {
         Computer c = node.toComputer();
-        if (c == null) {
+        if (c == null || c.isOffline()) {
             return;
         }
         // fast path bailout for Windows
-        if (c instanceof SlaveComputer && !Boolean.TRUE.equals(((SlaveComputer) c).isUnix())) {
+        if (!Boolean.TRUE.equals(c.isUnix())) {
```
batmat: This and above seem unrelated?

batmat:
Also, test failure. Retriggering. If this persists, you'll obviously want to wait until it's fixed before merging.

src/main/java/com/cloudbees/jenkins/support/impl/ProcFilesRetriever.java
jglick: It may be deprecated in later baselines, but it works as advertised and will continue to.

src/main/java/com/cloudbees/jenkins/support/impl/ProcFilesRetriever.java
duemir: You mean any refactoring due to new baseline should be in the separate PR?

duemir:
1. Now there are 4 components instead of 2. Those 2 that are for agents are not selected by default.
2. No more `getInstance` update.

duemir:
3. Rebased on top of the `master`
4. Fixed the test

duemir:
Are you kidding me? I resolved this conflict from your wonderful UI Github, why are you telling me that it's still there!!!

duemir:
All resolved and fixed. Ready for review again.

duemir:
@batmat any luck reviewing this 
@Evildethow @christ66 what do you think about updated version?

src/main/java/com/cloudbees/jenkins/support/api/FilePathContent.java
```diff
@@ -24,11 +24,11 @@
 
 package com.cloudbees.jenkins.support.api;
 
+import com.cloudbees.jenkins.support.SupportLogFormatter;
 import hudson.FilePath;
-import hudson.util.IOException2;
 
-import java.io.IOException;
-import java.io.OutputStream;
+import java.io.*;
```
batmat: hmmm

src/main/java/com/cloudbees/jenkins/support/api/FilePathContent.java
```diff
@@ -49,7 +53,25 @@ public void writeTo(OutputStream os) throws IOException {
         try {
             file.copyTo(os);
         } catch (InterruptedException e) {
-            throw new IOException2(e);
+            throw new IOException(e);
+        } catch (IOException e) {
+            if (isFileNotFound(e) || isFileNotFound(e.getCause())) {
+                OutputStreamWriter osw = new OutputStreamWriter(os, "utf-8");
+                try {
```
batmat: `try-with-resources`?

duemir: the `os` comes from the caller which is supposed to manage its lifecycle. We don't want to close it, even though there are provisions made to prevent `Content` classes from closing this stream.

src/main/java/com/cloudbees/jenkins/support/api/FilePathContent.java
duemir: My IntelliJ settings might be a bit aggressive about wildcard imports. Didn't know if there is some standard regarding this. Is there one?

christ66: We've been advising to use explicit imports since it makes having class name conflicts easier to resolve.

src/main/java/com/cloudbees/jenkins/support/impl/ProcFilesRetriever.java
christ66: Yes. For future PRs.

christ66: It should be refactored when the new baseline is updated.

src/main/java/com/cloudbees/jenkins/support/impl/SystemConfiguration.java
```diff
@@ -46,8 +47,29 @@
...
  */
-@Extension
-public class SystemConfiguration extends ProcFilesRetriever {
+public abstract class SystemConfiguration extends ProcFilesRetriever {
+    @Extension
+    public static class Master extends SystemConfiguration {
+        @Override
+        @NonNull
+        public String getDisplayName() {
+            return "Master system configuration (Linux only)";
```
christ66: Sine this is only available for linux I think we should add the ability to not show it if it's on windows. I.E.

`@Override public boolean isEnabled() { return (Platform.current() == Platform.UNIX); }`

christ66:
Minor :NITS:. I think it can be merged.

duemir:
@christ66 I addressed the imports comment. Would appreciate if you can merge it.

duemir:
5th most commented PR. Just a handful more comments and I will win 😆

duemir:
I don't think I have ever had a PR in review for almost half a year. Setting personal records on all fronts 🏆

PR #116, opened 2017-09-01
[JENKINS-42169] Include task logs in support bundles

rysteboe:
@reviewbybees 

https://issues.jenkins-ci.org/browse/JENKINS-42169

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/JenkinsLogs.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/JenkinsLogs.java
@@ -127,7 +127,7 @@ public class JenkinsLogs extends Component {
             }
         }
         File taskLogs = new File(logs, "tasks");
-        if (taskLogs.isDirectory()) {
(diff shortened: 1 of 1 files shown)


jglick:
Necessitated by https://github.com/jenkinsci/jenkins/pull/2047 IIUC. (Nice to have the backlinks.)

src/main/java/com/cloudbees/jenkins/support/impl/JenkinsLogs.java
```diff
@@ -126,6 +126,15 @@ private void addOtherMasterLogs(Container result) {
                 }
             }
         }
+        File taskLogs = new File(logs, "tasks");
+        if (taskLogs.exists() && taskLogs.isDirectory()) {
```
jglick: Redundant—`isDirectory` covers all relevant cases.

Actually it is not even necessary to check that here, or in the `logs` block above: if the directory does not exist, `listFiles` will return null, which we are already checking for.

PR #118, opened 2017-09-22
[FIX JENKINS-33203] trick to execute 'dmesg --ctime' if it is available if not 'dmesg'

kuisathaverat:
[JENKINS-40077](https://issues.jenkins-ci.org/browse/JENKINS-40077)

@reviewbybees


christ66:
👍

PR #119, opened 2017-10-06
[JENKINS-22791] Record heap histograms

escoem:
Added heap histogram generation component using _com.sun.management:type=DiagnosticCommand_ object and _GC.class_histogram_ command.

Only for master instance currently, the new component generates nodes/master/heap-histogram.txt file. Can be disabled with the option _Master Heap Histogram -histo:live_ 

please @reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/HeapHisto.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/HeapHisto.java
@@ -75,7 +75,7 @@ public class HeapHisto extends Component {
             final StringBuilder bos = new StringBuilder();
 
             bos.append("Master Heap Histogram -histo:live");
-            for (int i=1; i<limit; i++) {
+            for (int i=0; i<limit; i++) {
                 bos.append(lines[i]).append('\n');
             }
 
@@ -91,7 +91,7 @@ public class HeapHisto extends Component {
         private String getRawLiveHistogram() {
             String result;
             try {
-                ObjectName objName = new ObjectName("com.sun.management:type=DiagnosticCommand");;
+                ObjectName objName = new ObjectName("com.sun.management:type=DiagnosticCommand");
                 result = (String) ManagementFactory.getPlatformMBeanServer().invoke(objName, "gcClassHistogram", new Object[] {null}, new String[]{String[].class.getName()});
             }
             catch (InstanceNotFoundException | ReflectionException | MBeanException | MalformedObjectNameException e) {
--- a/src/main/java/com/cloudbees/jenkins/support/impl/HeapHisto.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/HeapHisto.java
@@ -29,7 +29,7 @@ import java.util.logging.Logger;
 @Extension
 public class HeapHisto extends Component {
     private static final int OFFSET = 3;
-    private static final int MAX = 500 + OFFSET;
+    private static final int MAX = 200 + OFFSET;
 
     private static final Logger logger = Logger.getLogger(HeapHisto.class.getName());
     private final WeakHashMap<Node, String> heapHistoCache = new WeakHashMap<Node, String>();
--- a/src/main/java/com/cloudbees/jenkins/support/impl/HeapHisto.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/HeapHisto.java
@@ -11,9 +11,12 @@ import hudson.remoting.Callable;
 import hudson.security.Permission;
 import jenkins.model.Jenkins;
 import org.jenkinsci.remoting.RoleChecker;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
 
 import javax.management.InstanceNotFoundException;
 import javax.management.MBeanException;
+import javax.management.MBeanServer;
 import javax.management.MalformedObjectNameException;
 import javax.management.ObjectName;
 import javax.management.ReflectionException;
@@ -26,7 +29,11 @@ import java.util.WeakHashMap;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+/**
+ * Heap histogram from master node.
+ */
 @Extension
+@Restricted(NoExternalUse.class)
 public class HeapHisto extends Component {
     private static final int OFFSET = 3;
     private static final int MAX = 200 + OFFSET;
@@ -43,7 +50,7 @@ public class HeapHisto extends Component {
     @NonNull
     @Override
     public String getDisplayName() {
-        return "Master Heap Histogram -histo:live";
+        return "Master Heap Histogram";
     }
 
     @Override
@@ -74,7 +81,7 @@ public class HeapHisto extends Component {
 
             final StringBuilder bos = new StringBuilder();
 
-            bos.append("Master Heap Histogram -histo:live");
+            bos.append("Master Heap Histogram");
             for (int i=0; i<limit; i++) {
                 bos.append(lines[i]).append('\n');
             }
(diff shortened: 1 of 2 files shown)


src/main/java/com/cloudbees/jenkins/support/impl/HeapHisto.java
```diff
@@ -0,0 +1,104 @@
...
+    @NonNull
+    @Override
+    public Set<Permission> getRequiredPermissions() {
+        return Collections.singleton(Jenkins.ADMINISTER);
+    }
+
+    @NonNull
+    @Override
+    public String getDisplayName() {
+        return "Master Heap Histogram -histo:live";
```
oleg-nenashev: I a not sure what is this flag doing here. I would rather provide a more extensive documentation in help.jelly (if it is supported by the plugin).

src/main/java/com/cloudbees/jenkins/support/impl/HeapHisto.java
```diff
@@ -0,0 +1,104 @@
...
+
+    private static final class GetLiveHeapHistogram implements Callable<String, RuntimeException> {
+        public String call() {
+            final String raw = getRawLiveHistogram();
+            final String[] lines = raw.split("\n");
+            final int limit = MAX <= lines.length ? MAX : lines.length;
+
+            final StringBuilder bos = new StringBuilder();
+
+            bos.append("Master Heap Histogram -histo:live");
```
oleg-nenashev: same, not sure why you need the flag here

src/main/java/com/cloudbees/jenkins/support/impl/HeapHisto.java
```diff
@@ -0,0 +1,104 @@
...
+import java.io.IOException;
+import java.io.OutputStream;
+import java.lang.management.ManagementFactory;
+import java.util.Collections;
+import java.util.Set;
+import java.util.WeakHashMap;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+@Extension
```
oleg-nenashev: Restrict it or document. Restricting is probably fine in this case

src/main/java/com/cloudbees/jenkins/support/impl/HeapHisto.java
```diff
@@ -0,0 +1,104 @@
...
+        @Override
+        public void checkRoles(RoleChecker checker) throws SecurityException {
+            // TODO: do we have to verify some role?
+        }
+
+        private String getRawLiveHistogram() {
+            String result;
+            try {
+                ObjectName objName = new ObjectName("com.sun.management:type=DiagnosticCommand");
+                result = (String) ManagementFactory.getPlatformMBeanServer().invoke(objName, "gcClassHistogram", new Object[] {null}, new String[]{String[].class.getName()});
```
oleg-nenashev: I'd guess null result is not possible here. Other implementations like [nuxeo-connect-tools](https://github.com/pipeline-as-code-nxmatic/nuxeo-connect-tools/blob/4423de6c0cfa7435d9f198b1b123a2ddc324ed6a/report-core/src/main/java/org/nuxeo/connect/tools/report/management/MxClassHistogramReport.java) do not check it as well

src/main/java/com/cloudbees/jenkins/support/impl/HeapHisto.java
```diff
@@ -0,0 +1,104 @@
...
+            }
+            catch (InstanceNotFoundException | ReflectionException | MBeanException | MalformedObjectNameException e) {
+                logger.log(Level.WARNING,"Could not record heap live histogram.", e);
+                result = "NA";
```
oleg-nenashev: "N/A"

escoem:
hey @oleg-nenashev, it's done

src/main/java/com/cloudbees/jenkins/support/impl/HeapUsageHistogram.java
```diff
@@ -0,0 +1,115 @@
...
+
+    private String getLiveHistogram(Node node) throws IOException {
+        return AsyncResultCache.get(node,
+                heapHistoCache,
+                new GetLiveHeapHistogram(),
+                "heap histogram",
+                "N/A: No connection to node, or no cache.");
+    }
+
+    private static final class GetLiveHeapHistogram implements Callable<String, RuntimeException> {
```
oleg-nenashev: If you use it only on master, it is recommended to avoid use `hudson.remoting.Callable` (if AsyncResultCache allows it). Otherwise maybe it makes sense to use `MasterToSlaveCallable` though it is not a 100% fit in this case

escoem:
Hi @oleg-nenashev , the class is already restricted. IIUC the AsyncResultCache.get requires a Callable. Should I stop using it?

oleg-nenashev:
@reviewbybees done

escoem:
Hi @oleg-nenashev I have no permissions to merge and release the plugin, can you do that for me?

src/main/java/com/cloudbees/jenkins/support/impl/HeapUsageHistogram.java
christ66: I agree. We should probably not create a `Callable` if we are obtaining this information from the master. Seems useless.

src/main/java/com/cloudbees/jenkins/support/impl/HeapUsageHistogram.java
```diff
@@ -0,0 +1,115 @@
...
+                @Override
+                public void writeTo(OutputStream os) throws IOException {
+                    os.write(getLiveHistogram(Jenkins.getInstance()).getBytes("UTF-8"));
+                }
+            }
+        );
+    }
+
+    private String getLiveHistogram(Node node) throws IOException {
+        return AsyncResultCache.get(node,
```
christ66: I hate to sound mean about this, but should we be caching the "live" histogram?

src/main/java/com/cloudbees/jenkins/support/impl/HeapUsageHistogram.java
```diff
@@ -0,0 +1,115 @@
...
+                new GetLiveHeapHistogram(),
+                "heap histogram",
+                "N/A: No connection to node, or no cache.");
+    }
+
+    private static final class GetLiveHeapHistogram implements Callable<String, RuntimeException> {
+        public String call() {
+            final String raw = getRawLiveHistogram();
+            final String[] lines = raw.split("\n");
+            final int limit = MAX <= lines.length ? MAX : lines.length;
```
christ66: I assume this is trying to truncate the output of the file? If so it might be best to use the `TruncatedContent` instead.

src/main/java/com/cloudbees/jenkins/support/impl/HeapUsageHistogram.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.logging.Logger;
+
+/**
+ * Heap histogram from master node.
+ */
+@Extension
+@Restricted(NoExternalUse.class)
+public class HeapUsageHistogram extends Component {
+    private static final int OFFSET = 3;
+    private static final int MAX = 200 + OFFSET;
```
christ66: :nit: Variable could be improved. `MAX` what? I assume it's max size, but best to define it as such. Also I'm not sure what's the reason for the `OFFSET` variable other than to set the `MAX` to 203.

src/main/java/com/cloudbees/jenkins/support/impl/HeapUsageHistogram.java
```diff
@@ -0,0 +1,98 @@
...
+            new Content("nodes/master/heap-histogram.txt") {
+                @Override
+                public void writeTo(OutputStream os) throws IOException {
+                    os.write(getLiveHistogram(Jenkins.getInstance()).getBytes("UTF-8"));
+                }
+            }
+        );
+    }
+
+    private String getLiveHistogram(Node node) throws IOException {
```
jglick: unused parameter

src/main/java/com/cloudbees/jenkins/support/impl/HeapUsageHistogram.java
```diff
@@ -0,0 +1,98 @@
...
+        final String raw = getRawLiveHistogram();
+        final String[] lines = raw.split("\n");
+        final int limit = MAX <= lines.length ? MAX : lines.length;
+
+        final StringBuilder bos = new StringBuilder();
+
+        bos.append("Master Heap Histogram");
```
jglick: redundant given filename, and missing newline anyway

batmat:
Hold on your pants, I'm releasing at the request of @christ66.

batmat:
2.42 is out.

PR #120, opened 2017-10-20
[JENKINS-41653] Support core warning when collecting data

escoem:
https://issues.jenkins-ci.org/browse/JENKINS-41653

malformed xml files are not included. Test validating the issue.

Please @reviewbybees @varmenise @batmat @oleg-nenashev

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
+++ b/src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
@@ -38,6 +38,7 @@ class SecretHandler {
      */
     protected static final String SECRET_MARKER = "#secret#";
     public static final String OUTPUT_ENCODING = "UTF-8";
+    public static final Pattern SECRET_PATTERN = Pattern.compile(">\\{(.*)\\}<|>(.*)\\=<");
 
     /**
      * find the secret in the xml file and replace it with the place holder
@@ -100,7 +101,7 @@ class SecretHandler {
 
 
     private static String findSecretFallback(String xml) {
-        Matcher matcher = Pattern.compile(">\\{(.*)\\}<|>(.*)\\=<").matcher(xml);
+        Matcher matcher = SECRET_PATTERN.matcher(xml);
         while(matcher.find()) {
             String secret = matcher.group();
             if(secret.length() > 1)
--- a/src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
+++ b/src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
@@ -41,6 +41,13 @@ class SecretHandler {
     public static final Pattern SECRET_PATTERN = Pattern.compile(">\\{(.*)\\}<|>(.*)\\=<");
 
     /**
+     * Enabled by default.
+     * FALLBACK will be disable in case you define a system property like -Dsupport-core-plugin.SecretHandler.ENABLE_FALLBACK=false
+     * Otherwise will be enabled.
+     */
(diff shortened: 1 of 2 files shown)


src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
```diff
@@ -77,15 +81,37 @@ public void characters(char[] ch, int start, int length) throws SAXException {
...
         transformer.setOutputProperty(OutputKeys.OMIT_XML_DECLARATION, "yes");
         transformer.setOutputProperty(OutputKeys.ENCODING, OUTPUT_ENCODING);
-        transformer.transform(src, res);
 
-        return result.toString("UTF-8");
+        try {
+            transformer.transform(src, res);
+            return result.toString("UTF-8");
+        } catch (TransformerException e) {
+            return findSecretFallback(str);
```
oleg-nenashev: Probably there should be fine-grain logging to notify about the issue

src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
```diff
@@ -77,15 +81,37 @@ public void characters(char[] ch, int start, int length) throws SAXException {
...
+        }
     }
+
+
+    private static String findSecretFallback(String xml) {
```
oleg-nenashev: This method definitely works as a fallback, but I am bit concerned about corner-cases. What if the file is damaged, e.g. what if the last `}` is missing? Does it add risks to this implementation?

src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
```diff
@@ -77,15 +81,37 @@ public void characters(char[] ch, int start, int length) throws SAXException {
...
+        Matcher matcher = Pattern.compile(">\\{(.*)\\}<|>(.*)\\=<").matcher(xml);
+        while(matcher.find()) {
```
oleg-nenashev: Maybe makes sense to cache pattern and to avoid compiling it every time. Likely NIT since it's a fallback

src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
oleg-nenashev: I would also recommend this routine to be opt-out or opt-in. Skipping broken files is generally the most reasonable approach though you lose info

src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
escoem: it is the same behavior than the current implementation. If it doesn't match with {}, the secret is not overwritten

src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
oleg-nenashev: I would vote for it. better safe than sorry

src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
oleg-nenashev: The current solution needs a valid XML, and it's hard to imagine a case when we get a corrupted secret and a full XML file in the case of the glitch. Anyway, it's rather nitpicking

escoem:
it's done @oleg-nenashev

escoem:
@reviewbybees done cc. @batmat is this PR ready for merging?

PR #121, opened 2017-11-03
[FIXED JENKINS-47779] Could not attach 'plugins/active.txt' to support bundle

escoem:
https://issues.jenkins-ci.org/browse/JENKINS-47779

Since Jenkins 2.86, pluginManager is returning an unmodifiable list each time getPlugins is called. 

Support core plugin was sorting the list by using Collections.sort thus currently is throwing UnsupportedOperationException.

Please, @reviewbybees @christ66 @andresrc @oleg-nenashev

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
@@ -107,15 +107,19 @@ public class AboutJenkins extends Component {
 
     @Override
     public void addContents(@NonNull Container container) {
-        container.add(new AboutContent());
+        List<PluginWrapper> activePlugins = new ArrayList<PluginWrapper>();
+        List<PluginWrapper> disabledPlugins = new ArrayList<PluginWrapper>();
+
+        populatePluginsLists(activePlugins, disabledPlugins);
+
+        container.add(new AboutContent(activePlugins));
         container.add(new ItemsContent());
         container.add(new NodesContent());
-        container.add(new ActivePlugins("plugins/active.txt"));
-        container.add(new DisabledPlugins());
+        container.add(new ActivePlugins(activePlugins));
+        container.add(new DisabledPlugins(disabledPlugins));
         container.add(new FailedPlugins());
 
-        // container.add(new ActivePlugins("docker/plugins.txt"));
-        container.add(new Dockerfile());
+        container.add(new Dockerfile(activePlugins, disabledPlugins));
 
         container.add(new MasterChecksumsContent());
         for (final Node node : Helper.getActiveInstance().getNodes()) {
@@ -540,8 +544,11 @@ public class AboutJenkins extends Component {
     }
 
     private static class AboutContent extends PrintedContent {
-        AboutContent() {
+        private final Iterable<PluginWrapper> plugins;
+
+        AboutContent(Iterable<PluginWrapper> plugins) {
             super("about.md");
+            this.plugins = plugins;
         }
         @Override protected void printTo(PrintWriter out) throws IOException {
             final Jenkins jenkins = Helper.getActiveInstance();
@@ -583,7 +590,7 @@ public class AboutJenkins extends Component {
             out.println("Active Plugins");
             out.println("--------------");
             out.println();
-            Iterable<PluginWrapper> plugins = getPluginsSorted();
+
             for (PluginWrapper w : plugins) {
                 if (w.isActive()) {
                     out.println("  * " + w.getShortName() + ":" + w.getVersion() + (w.hasUpdate()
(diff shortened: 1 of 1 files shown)


src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
```diff
@@ -1010,4 +1007,15 @@ private String getLabelString(Node n) {
...
+     * Fixes JENKINS-47779 caused by JENKINS-47713
+     * Not using SortedSet because of PluginWrapper doesn't implements equals and hashCode.
+     * @param list original list, probably an unmodifiableList
+     * @return new copy of the list sorted
+     */
+    private static Iterable<PluginWrapper> getSorted(List<PluginWrapper> list) {
+        final List<PluginWrapper> sorted = new LinkedList<PluginWrapper>(list);
+        Collections.sort(sorted);
+        return sorted;
+    }
```
andresrc: The sorting can be done in a generic method, and the method with all the comments could also obtain the plugins from the plugin manager, which is also something repeated in all the uses.

escoem: Thanks @andresrc

src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
```diff
@@ -1010,4 +1004,24 @@ private String getLabelString(Node n) {
...
+
+    private static Iterable<PluginWrapper> getPluginsSorted(PluginManager pluginManager) {
+        return listToSortedIterable(pluginManager.getPlugins());
+    }
+
+
+    private static <T extends Comparable<T>> Iterable<T> listToSortedIterable(List<T> list) {
+        final List<T> sorted = new LinkedList<T>(list);
+        Collections.sort(sorted);
```
oleg-nenashev: it's not the most effective way of sorting, but OK.

escoem: It's the same way than before but suggestions are welcome :-)

src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
```diff
@@ -582,9 +583,7 @@ public synchronized int n() {
             out.println("Active Plugins");
             out.println("--------------");
             out.println();
-            PluginManager pluginManager = jenkins.getPluginManager();
-            List<PluginWrapper> plugins = new ArrayList<PluginWrapper>(pluginManager.getPlugins());
-            Collections.sort(plugins);
+            Iterable<PluginWrapper> plugins = getPluginsSorted();
```
batmat: NIT: get rid of the local variable as it's unnecessary and inline it below?

batmat:
@reviewbybees done

Jimilian:
@oleg-nenashev, @escoem I don't know much about this plugin, but Core API provides two lists of plugins: `plugins` (accessible by method `getPlugins`) and `activePlugins` ( plugins that are used by Jenkins itself + it's topological sorted.). So, two comments:
- based on title of this PR I expected to see the usage of `activePlugins`, because currently it dumps not used plugins as well and after that filters them out
- We access same method from three classes: AboutContent, ActivePlugins and DisabledPlugins. If it's possible I think it would be better to pass the sorted list of plugins as parameter to constructors. Otherwise we will be copy/sort same thing multiple times for nothing. Also the content can be different :)

escoem:
@Jimilian why not :-) I have reused the lists as you have suggested.

escoem:
@reviewbybees done @batmat is this PR reading for merging?

src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
```diff
@@ -1010,4 +1003,33 @@ private String getLabelString(Node n) {
...
+        Collections.sort(sorted);
+        return sorted;
+    }
+
+    private static void populatePluginsLists(List<PluginWrapper> activePlugins, List<PluginWrapper> disabledPlugins) {
+        for(PluginWrapper plugin : getPluginsSorted()){
+            if(plugin.isActive()) {
+                activePlugins.add(plugin);
+            } else {
+                disabledPlugins.add(plugin);
```
batmat: You're slightly changing the behaviour here for the Dockerfile, FWIW, from `isEnabled()` to `isActive()`. That means the plugins list will be the one really active currently, not the one that *will* be next time. 

I suppose this is OK and probably what is desired from a support standpoint but well.

batmat:
Moving forward as this is broken with core 2.86+ and the new LTS base has been selected and is 2.89.

So merging and releasing soon to give it some time to soak.

PR #105, opened 2017-02-28
[JENKINS-35341][JENKINS-35343] Delete selectable Bundle

jpthiery:
List Bundle files to display bundle files.
Then allow to Delete them selecting from UI.
Done thanks to help of @alecharp .

PR #122, opened 2017-12-07
[JENKINS-48436] Restore backward compatibility on web endpoints

Vlatombe:
This restores backward compatibility on the endpoint `/support/download`, as well as adding tests checking that the endpoints remain valid.

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/SupportAction.java
+++ b/src/main/java/com/cloudbees/jenkins/support/SupportAction.java
@@ -161,6 +161,18 @@ public class SupportAction implements RootAction {
         rsp.sendRedirect("");
     }
 
+    /**
+     * Generates a support bundle.
+     * @param req The stapler request
+     * @param rsp The stapler response
+     * @throws ServletException
+     * @throws IOException
+     */
+    @RequirePOST
+    public void doDownload(StaplerRequest req, StaplerResponse rsp) throws ServletException, IOException {
+        doGenerateAllBundles(req, rsp);
+    }
+
     @RequirePOST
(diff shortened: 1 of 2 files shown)


jglick:
Cf. [JENKINS-48436](https://issues.jenkins-ci.org/browse/JENKINS-48436) & #105.

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -161,6 +161,18 @@ public void doDeleteBundles(StaplerRequest req, StaplerResponse rsp) throws Serv
...
 
+    /**
+     * Generates a support bundle.
+     * @param req The stapler request
+     * @param rsp The stapler response
+     * @throws ServletException
+     * @throws IOException
+     */
+    @RequirePOST
+    public void doDownload(StaplerRequest req, StaplerResponse rsp) throws ServletException, IOException {
```
jglick: Suggest marking `@Deprecated`.

src/test/java/com/cloudbees/jenkins/support/SupportActionTest.java
```diff
@@ -39,6 +43,24 @@ public void setUp() {
...
+    }
+
+    @Test
+    public void generateAllBundles() throws IOException, SAXException {
+        downloadBundle("/generateAllBundles?json={\"components\":1}");
+    }
+
+    private void downloadBundle(String s) throws IOException, SAXException {
+        JenkinsRule.JSONWebResponse jsonWebResponse = rule.postJSON(root.getUrlName() + s, "");
+        File zipFile = File.createTempFile("test", "zip");
```
jglick: Use

```java
@Rule
public TemporaryFolder tmp = new TemporaryFolder();
```

PR #100, opened 2017-01-05
[JENKINS-40613] Report a blocked Timer

jglick:
Would I think make problems like that which https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/23 purports to solve more immediately apparent.

@reviewbybees

PR #123, opened 2017-12-08
Fixing a ClassCastException

jglick:
Seems that diagnostics in #100 cause an error as of https://github.com/jenkinsci/jenkins/pull/2792; observed in a support bundle a `manifest/errors.txt`:

```
Could not attach 'nodes/master/thread-dump.txt' to support bundle
-----------------------------------------------------------------------

java.lang.ClassCastException: jenkins.security.ImpersonatingScheduledExecutorService cannot be cast to java.util.concurrent.ScheduledThreadPoolExecutor
	at com.cloudbees.jenkins.support.impl.ThreadDumps$1.writeTo(ThreadDumps.java:90)
	at com.cloudbees.jenkins.support.SupportPlugin.writeBundle(SupportPlugin.java:359)
	at com.cloudbees.jenkins.support.SupportPlugin.writeBundle(SupportPlugin.java:273)
	at com.cloudbees.jenkins.support.SupportPlugin$PeriodicWorkImpl$1.run(SupportPlugin.java:761)
```

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/ThreadDumps.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/ThreadDumps.java
@@ -36,7 +36,6 @@ import java.util.Collections;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.logging.Level;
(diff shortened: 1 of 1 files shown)


jglick:
_Could_ attempt to restore the former diagnostics, by calling `InterceptingExecutorService.delegate` reflectively, but this seems like overkill. Any problem with the `Timer` thread pool is likely to be visible from the thread dump anyway. (If we let that be generated—due to this bug, `nodes/master/thread-dump.txt` is empty!)

jglick:
> using `instanceof` to keep the original diagnostics

Would only be useful in 2.50-; not worth bothering with.

PR #124, opened 2017-12-15
POM & baseline update

jglick:
According to [stats](http://stats.jenkins.io/pluginversions/support-core.html) there are almost no users on pre-2.60.x of 2.42 or newer versions, and that was released several weeks ago, so I think few people will be left behind.

@reviewbybees


jglick:
@reviewbybees done

PR #125, opened 2017-12-18
Never needed to refer to RoleChecker

jglick:
At least some of the incorrect usages seem to have been introduced in #53, and maybe people just copied it after that.

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/AsyncResultCache.java
+++ b/src/main/java/com/cloudbees/jenkins/support/AsyncResultCache.java
@@ -26,14 +26,14 @@ public class AsyncResultCache<T> implements Runnable {
     private final Node node;
     private final String name;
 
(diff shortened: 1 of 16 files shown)


src/main/java/com/cloudbees/jenkins/support/AsyncResultCache.java
```diff
@@ -26,14 +26,14 @@
     private final Node node;
     private final String name;
 
-    public static <V, T extends java.lang.Throwable> V get(Node node, WeakHashMap<Node, V> cache, Callable<V,T> operation, String name, V defaultIfNull)
+    public static <V, T extends java.lang.Throwable> V get(Node node, WeakHashMap<Node, V> cache, /*MasterToSlave*/Callable<V,T> operation, String name, V defaultIfNull)
```
oleg-nenashev: Uh oh, maybe introducing a new API is reasonable in this case

jglick: Seems like overkill I think—there are no known external usages.

PR #129, opened 2018-01-05
Switch to RewindableFileOutputStream

jglick:
Picks up https://github.com/jenkinsci/jenkins/pull/2490. I saw a bundle whose `nodes/master/file-descriptors.txt` included a lot of entries like

    /var/lib/jenkins/support/cache/…/all_2017-…04.log

and maybe this is why.

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/JenkinsLogs.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/JenkinsLogs.java
@@ -15,8 +15,8 @@ import hudson.logging.LogRecorder;
 import hudson.model.PeriodicWork;
 import hudson.security.Permission;
 import hudson.util.CopyOnWriteList;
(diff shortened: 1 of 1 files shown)


src/main/java/com/cloudbees/jenkins/support/impl/JenkinsLogs.java
```diff
@@ -15,8 +15,8 @@
 import hudson.model.PeriodicWork;
 import hudson.security.Permission;
 import hudson.util.CopyOnWriteList;
-import hudson.util.io.ReopenableFileOutputStream;
-import hudson.util.io.ReopenableRotatingFileOutputStream;
+import hudson.util.io.RewindableFileOutputStream;
```
batmat: OK, gotcha `ReopenableFileOutputStream` is actually deprecated.

PR #131, opened 2018-01-05
Sort thread dumps by name rather than ID

jglick:
Otherwise it is cumbersome to find threads of a particular kind, like all `Executor`s. The ID is still recorded in the file; this just changes the display order to something I think will be more helpful for humans.

@reviewbybees


src/main/java/com/cloudbees/jenkins/support/impl/ThreadDumps.java
```diff
@@ -240,8 +242,14 @@ public static void threadDumpModern(OutputStream out) throws UnsupportedEncoding
             threads = new ThreadInfo[0];
         }
 
-        for (int ti = threads.length - 1; ti >= 0; ti--) {
-            printThreadInfo(writer, threads[ti], mbean);
+        Arrays.sort(threads, new Comparator<ThreadInfo>() {
+            @Override
+            public int compare(ThreadInfo t1, ThreadInfo t2) {
+                return t1.getThreadName().compareTo(t2.getThreadName());
```
jglick: nicer in Java 8 BTW

PR #132, opened 2018-01-09
[JENKINS-21670] Add option to anonymize jobs, views, nodes, etc

rysteboe:
https://issues.jenkins-ci.org/browse/JENKINS-21670

@reviewbybees 

Everything about the UX side of this needs work, but it is functional from what testing I could do locally.  Needs to be tested on a larger scale instance and discussions about the UX need to happen.

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/AnonymizedItems.java
+++ b/src/main/java/com/cloudbees/jenkins/support/AnonymizedItems.java
@@ -26,6 +26,8 @@ package com.cloudbees.jenkins.support;
 import com.cloudbees.jenkins.support.util.Anonymizer;
 import hudson.Extension;
 import hudson.model.ManagementLink;
+import hudson.util.FormApply;
+import net.sf.json.JSONObject;
 import org.kohsuke.stapler.StaplerRequest;
 import org.kohsuke.stapler.StaplerResponse;
 import org.kohsuke.stapler.interceptor.RequirePOST;
@@ -40,11 +42,17 @@ import java.util.Map;
  */
 @Extension
 public class AnonymizedItems extends ManagementLink {
+    private SupportPlugin plugin;
+    private SupportPlugin.AnonymizationSettings settings;
+
+    public AnonymizedItems() {
+        plugin = SupportPlugin.getInstance();
+        settings = plugin.getAnonymizationSettings();
+    }
 
     @CheckForNull
     @Override
     public String getIconFileName() {
-        // TODO:  New icon?
         return "/plugin/support-core/images/24x24/support.png";
     }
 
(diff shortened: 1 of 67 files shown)


src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -821,4 +846,81 @@ public boolean configure(StaplerRequest req, JSONObject json) throws FormExcepti
...
+            excludedWords.add("all");
+            excludedWords.add("master");
+            excludedWords.add("jenkins");
+            excludedWords.add("node");
+            excludedWords.add("computer");
+            excludedWords.add("item");
+            excludedWords.add("label");
+            excludedWords.add("view");
+            excludedWords.add("user");
+            excludedWords.add("");
```
dwnusbaum: Some of these seem weird to me (e.g., "a", "the", "all") and there are some words that seem to be related to what's here but are not included (e.g. "agent", "slave", "job", "build"). Are these defaults a just-in-case defense, or are they necessary for the feature to work correctly?

dwnusbaum: I guess the issue is that if someone created a job called "computer", then all uses of "computer" would get anonymized, even where it isn't being used as the name of a job?

src/main/java/com/cloudbees/jenkins/support/AnonymizedItems.java
```diff
@@ -0,0 +1,148 @@
...
+        return settings.isAnonymizeUsers();
+    }
+
+    public void setAnonymizeUsers(boolean anonymizeUsers) throws IOException {
+        settings.setAnonymizeUsers(anonymizeUsers);
+    }
+
+    @RequirePOST
+    public void doUpdateNow(StaplerRequest request, StaplerResponse response) throws ServletException, IOException {
+        Anonymizer.refresh();
```
dwnusbaum: Is there an issue if this is called while a support bundle is being generated? I think it is fine as long as the config hasn't changed, but if it has changed would it be possible to end up with invalid mappings?

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
rysteboe: Yeah, exactly.  I added some words that I know are going to be used a lot and that are extremely unlikely to pose any sort of risk to not be anonymized.  

Unfortunately, I couldn't figure out a better way to do this.  I don't think this is good, I just don't know what the other options are short of rewriting how the logs are written in the first place (which is rather unfeasible).

src/main/java/com/cloudbees/jenkins/support/AnonymizedItems.java
rysteboe: I don't think so - I think that what would happen is there would be inconsistency within the support bundle.

That said, I honestly have a real lack of experience with dealing with multiple threads, so I'd greatly appreciate someone looking at the potential flaws in `Anonymizer` wrt multithreading.

src/main/java/com/cloudbees/jenkins/support/util/Anonymizer.java
```diff
@@ -0,0 +1,207 @@
...
+            }
+        }
+
+        save();
+    }
+
+    public static String anonymize(String input) {
+        String anonymized = input;
+        for (String key : ORDER) {
+            anonymized = anonymized.replaceAll("(?i)\\b(" + key + ")\\b", ANON_MAP.get(key));
```
dwnusbaum: We need to escape `key` here so that Regex control characters are treated as literals. Right now having a Job with a weird name such as `J1-test_test2{hmm}+nice` throws a `PatternSyntaxException` in `replaceAll`.

dwnusbaum: TIL: https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html#quote-java.lang.String-

src/main/java/com/cloudbees/jenkins/support/util/Anonymizer.java
```diff
@@ -0,0 +1,207 @@
...
+        return new XmlFile(Jenkins.XSTREAM, new File(Jenkins.get().getRootDir(), "secrets/anonymized-names.xml"));
+    }
+
+    @SuppressWarnings("unchecked")
+    private static synchronized void load() {
+        XmlFile config = getConfigXml();
+        if (config.exists()) {
+            try {
+                ANON_MAP.clear();
+                ANON_MAP.putAll((Map<? extends String, ? extends String>) config.read());
```
dwnusbaum: We need to also load the other fields, otherwise the UI and anonymization don't work after a restart because `DISPLAY` and `ORDER` stay empty. We could derive them from ANON_MAP like the following, but it might make more sense to store them independently so that original and escaped names don't end up in the display.

    DISPLAY.putAll(ANON_MAP);
    ORDER.addAll(ANON_MAP.keySet());

src/main/java/com/cloudbees/jenkins/support/util/Anonymizer.java
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    @SuppressWarnings("unchecked")
+    public static void refresh() {
+        LOGGER.log(Level.FINE, "Refreshing anonymized items");
+        Jenkins instance = Jenkins.getInstance();
+        load();
+
+        SupportPlugin.AnonymizationSettings settings = SupportPlugin.getInstance().getAnonymizationSettings();
+        if (settings.isAnonymizeLabels()) {
```
dwnusbaum: We should probably refresh the labels after nodes, because otherwise all nodes end up with the `label` prefix since there is always an implicit label with the name of the agent.

src/main/java/com/cloudbees/jenkins/support/AnonymizedItems.java
jvz: This could be addressed by making `Anonymizer::refresh` a `synchronized` method, though that may be overkill depending on where it's used.

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
jvz: The common terminology for these are _stop words_. [Example list](https://kb.yoast.com/kb/list-stop-words/).

dwnusbaum: > A more appropriate place might be a text file resource of some sort. That way end users could potentially add more things if necessary locally.

I think that what we have here should cover most cases (I would probably leave out the normal English words and just focus on the Jenkins model objects), but to be more exhaustive we could also add some dynamic entries based on iterating through the result of `ExtensionList.lookup(AbstractItem.class)` and adding `AbstractItem#getPronoun` and `AbstractItem#getTaskNoun`.

jvz: Definitely like that idea. I'm currently trying to make the set of strings configurable in the settings as well.

PR #134, opened 2018-03-06
[FIXED JENKINS-49931] Heap Histogram Collection Destabilizes Masters

escoem:
https://issues.jenkins-ci.org/browse/JENKINS-49931

@svanoort @reviewbybees please review.

Now the histogram is disabled by default, thus you will need to enable it either adding a system property _-Dcom.cloudbees.jenkins.support.impl.HeapUsageHistogram.DISABLED=false_ or running  from Script Console the line _com.cloudbees.jenkins.support.impl.HeapUsageHistogram.DISABLED=false_

Yes, the system property and the direct assignment have the same path/name.

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/HeapUsageHistogram.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/HeapUsageHistogram.java
@@ -33,6 +33,10 @@ public class HeapUsageHistogram extends Component {
     // first 200 classes so 203 lines required because of the header
     private static final int MAX = 203;
 
+    // disabled by default because of JENKINS-49931
+    // to be reviewed in the future.
+    private static /*final*/ boolean DISABLED = Boolean.parseBoolean(System.getProperty(HeapUsageHistogram.class.getCanonicalName() + ".DISABLED", "true"));
+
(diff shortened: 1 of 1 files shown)


src/main/java/com/cloudbees/jenkins/support/impl/HeapUsageHistogram.java
```diff
@@ -33,6 +33,10 @@
     // first 200 classes so 203 lines required because of the header
     private static final int MAX = 203;
 
+    // disabled by default because of JENKINS-49931
+    // to be reviewed in the future.
+    private static /*final*/ boolean DISABLED = Boolean.parseBoolean(System.getProperty(HeapUsageHistogram.class.getCanonicalName() + ".DISABLED", "true"));
```
jglick: BTW use of the negative sense is discouraged. Rather use `ENABLED` and default to false.

Also use `SystemProperties` since we are on Jenkins 2.x now.

escoem: @jglick thanks! I will use ENABLE instead of DISABLE in the future. I have created the https://issues.jenkins-ci.org/browse/JENKINS-50010 in order to let admins enable/disable components through UI, and also a common way to enforce the disable of components as a system property.

PR #135, opened 2018-03-06
[FIXED JENKINS-49931] Heap Histogram Collection Destabilizes Masters

escoem:
https://issues.jenkins-ci.org/browse/JENKINS-49931

@svanoort @dwnusbaum @jglick  @reviewbybees please review.

Used -all parameter in order to avoid the fullgc. By default, the gcClassHistogram is invoked using -all

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/HeapUsageHistogram.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/HeapUsageHistogram.java
@@ -34,8 +34,7 @@ public class HeapUsageHistogram extends Component {
     private static final int MAX = 203;
 
     // disabled by default because of JENKINS-49931
-    // to be reviewed in the future.
-    private static /*final*/ boolean DISABLED = Boolean.parseBoolean(System.getProperty(HeapUsageHistogram.class.getCanonicalName() + ".DISABLED", "true"));
+    private static /*final*/ boolean FULLGC = Boolean.getBoolean(HeapUsageHistogram.class.getCanonicalName() + ".FULLGC");
 
     private static final Logger logger = Logger.getLogger(HeapUsageHistogram.class.getName());
 
@@ -84,15 +83,6 @@ public class HeapUsageHistogram extends Component {
     }
 
     private String getRawLiveHistogram() {
-        if (DISABLED) {
-            return new StringBuilder().append('\n')
-                    .append("Histogram generation is disabled. If you want to enable it, do either:")
-                    .append('\n')
-                    .append("* Add the system property: -Dcom.cloudbees.jenkins.support.impl.HeapUsageHistogram.DISABLED=false")
-                    .append('\n')
-                    .append("* Run from Script Console the line: com.cloudbees.jenkins.support.impl.HeapUsageHistogram.DISABLED=false")
-                    .toString();
-        }
         String result;
         try {
             ObjectName objName = new ObjectName("com.sun.management:type=DiagnosticCommand");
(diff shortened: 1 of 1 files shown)


src/main/java/com/cloudbees/jenkins/support/impl/HeapUsageHistogram.java
```diff
@@ -84,23 +83,16 @@ private String getLiveHistogram() throws IOException {
...
-        }
         String result;
         try {
             ObjectName objName = new ObjectName("com.sun.management:type=DiagnosticCommand");
             MBeanServer platform = ManagementFactory.getPlatformMBeanServer();
             if (platform == null) {
                 return "N/A";
             }
-            result = (String) platform.invoke(objName, "gcClassHistogram", new Object[] {null}, new String[]{String[].class.getName()});
+            String[] params = FULLGC ? null : new String[] {"-all"};
```
svanoort: I don't see us getting much value from the histogram if it's capturing dead objects too.  Better to be live-only or off.

aheritier:
@svanoort I don't know. I understand that the histogram without the `-all` is less useful but when a memory issue happens I'm afraid that we cannot activate the histogram without restarting and thus at the best we loose some time (and perhaps we won't reproduce the issue). At least with the non `-all` we'll have something to look at even if not perfect.

But for sure in the other side I don't want that we take some risks in term of stability. This problem with
 the histogram generation just proved again that we don't have the capacity to test our changes in real conditions and our changes, even if they have a good objective, could impact our customers.

@ryancampbell @nazrul007 WDYT ?

svanoort:
> typically most of the heap will not be collectible, for the kind of workloads in Jenkins

@jglick This is not unfortunately not true, from my extensive analyses of the GC behavior in Jenkins.  Jenkins generates *tons* of object  garbage.  Object creation rates on busy large-scale masters exceed 300 MB/s on average with far higher burst rates, and 150 MB/s would be normal for a quieter/smaller master.   With healthy GC, generally ~0.1-1% of that object content is promoted to old-gen (promotion rates of <1 MB/s, usually a few hundred kB/S).   Generally with G1 GC we end up using over half the heap for young-gen, most of which is collectible. 

Using non-live histograms is really only useful if the master is nearly idle, and in most other cases will generate a very poor signal to noise ratio.

svanoort:
@aheritier You can toggle it off of DISABLED via the Script console without restarting the master.

If we could set it up to collect a heap histogram only when explicitly requested by a user (rather than periodically) that would be even better but alas is not an option at this time from what I've been told.

My explicit concern here is that we'll get a lot of people chasing red-herring memory issues that are due to `WeakReference`/`SoftReference`-cached objects and the buffers and CPS objects created by Pipeline at runtime.   See above for some ballpark figures on memory use and why I think garbage volume will generate false positives. 

I've done a fair bit of profiling for what kinds of object allocation we see for Pipeline in general and have a solid general idea for what constitutes norma memoryl patterns for a production master (although obviously we don't get the luxury of attaching a profiler).

escoem:
I will add DISABLED feature today @svanoort

svanoort:
@escoem I think you misunderstand -- I think the already-merged PR you did is great and really does everything we need.

svanoort:
Specifically, I'd like to see a release cut ASAP with https://github.com/jenkinsci/support-core-plugin/pull/134/ because I think this was the most perfect version we could hope for without adding major changes to support-core.   It's straightforward to disable/enable heap histograms -- we might consider adding a UI option to toggle the variable if we wanted, but it does what it needs to. 

It also means we don't need to worry about potential stability/performance impacts from collecting Live histograms regularly (something I can't rule out and would like to avoid).

escoem:
I know what has been done @svanoort, thanks. Closing this PR in favor of https://github.com/jenkinsci/support-core-plugin/pull/136

PR #136, opened 2018-03-08
[FIXED JENKINS-50008] Heap Histogram (all instances by default)

escoem:
https://issues.jenkins-ci.org/browse/JENKINS-50008

Because of live histogram requires a fullgc, by default now is generated for all instances (not only live one).

You could easily configure it by using the system property -Dcom.cloudbees.jenkins.support.impl. HeapUsageHistogram.FULLGC=true (permanent) or by running from Script Console com.cloudbees.jenkins.support.impl. HeapUsageHistogram.FULLGC=true (temporal) if you want to grab live instances.

The histogram component could be easily disabled by using the system property -Dcom.cloudbees.jenkins.support.impl. HeapUsageHistogram.DISABLE=true (permanent) or by running from Script Console com.cloudbees.jenkins.support.impl. HeapUsageHistogram.DISABLE=true (temporal).

I prefer to have something (all instances) than nothing (disable) by default.

Please @jglick @reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/HeapUsageHistogram.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/HeapUsageHistogram.java
@@ -33,9 +33,8 @@ public class HeapUsageHistogram extends Component {
     // first 200 classes so 203 lines required because of the header
     private static final int MAX = 203;
 
-    // disabled by default because of JENKINS-49931
-    // to be reviewed in the future.
-    private static /*final*/ boolean DISABLED = Boolean.parseBoolean(System.getProperty(HeapUsageHistogram.class.getCanonicalName() + ".DISABLED", "true"));
+    private static /*final*/ boolean FULLGC = Boolean.getBoolean(HeapUsageHistogram.class.getCanonicalName() + ".FULLGC");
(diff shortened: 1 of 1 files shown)


svanoort:
@escoem I still don't see how this addresses my objections in https://github.com/jenkinsci/support-core-plugin/pull/135#issuecomment-370869466 and https://github.com/jenkinsci/support-core-plugin/pull/135#issuecomment-370873737?  

If see any value from the (moderately useless) live heap histograms, there's always the command-line option via JMap.

recampbell:
Perhaps the feature that we want is to optionally collect the heap histogram when generating the bundle?

I think we want a way to do this easily from the support bundle instead of going through the process of getting the user to successfully execute jmap which is always a trial and especially so when the user is under the stress of an outage.

christ66:
>Perhaps the feature that we want is to optionally collect the heap histogram when generating the bundle?

I think the issue is if I don't have access to the UI, then it's impossible to generate a bundle. This seems like a good opportunity to make the `SlowRequests` an extension point to allow for this type of diagnostic information available when the UI isn't accessible.

> If we see any value from garbage-filled (literally) heap histograms, there's always the command-line option via JMap to collect them.

True, but not everyone installs with the JDK and some install with the JRE (iirc JRE doesn't come with the jmap command).

varyvol:
It seems there's not real progress here so closing. Please reopen if you consider this can still move forward.

PR #137, opened 2018-03-08
added to keep backward compatibility

escoem:
please @reviewbybees UPDATE -> we found abnormal private usages of this API, but for temporary backward compat let's reintroduce it to give some more time to adapt, and then actually remove.

**UPDATE (@batmat): Helper removed in https://github.com/jenkinsci/support-core-plugin/pull/127**

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/util/Helper.java
+++ b/src/main/java/com/cloudbees/jenkins/support/util/Helper.java
@@ -1,0 +1,27 @@
+package com.cloudbees.jenkins.support.util;
+
+import jenkins.model.Jenkins;
+import javax.annotation.Nonnull;
+
+/**
+ * Simple helper so we don't have to check {@code Jenkins.getInstance() != null} everywhere.
+ *
+ * TODO: replace with Jenkins.getActiveInstance() when on core {@literal >=} 1.609
+ *
+ */
+@Deprecated
+public final class Helper {
+    /** Not instantiable. */
+    private Helper() {
+        throw new AssertionError("Not instantiable");
+    }
+
+    @Nonnull
+    public static Jenkins getActiveInstance() {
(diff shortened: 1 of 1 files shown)


src/main/java/com/cloudbees/jenkins/support/util/Helper.java
```diff
@@ -0,0 +1,27 @@
+package com.cloudbees.jenkins.support.util;
+
+import jenkins.model.Jenkins;
+import javax.annotation.Nonnull;
+
+/**
+ * Simple helper so we don't have to check {@code Jenkins.getInstance() != null} everywhere.
+ *
+ * TODO: replace with Jenkins.getActiveInstance() when on core {@literal >=} 1.609
```
dwnusbaum: What? Already on 2.46 in pom.xml: https://github.com/jenkinsci/support-core-plugin/blob/70684fc409f53a10db0c0d31c58fc3dbdf82d65f/pom.xml#L72

src/main/java/com/cloudbees/jenkins/support/util/Helper.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+@Deprecated
+public final class Helper {
+    /** Not instantiable. */
+    private Helper() {
+        throw new AssertionError("Not instantiable");
+    }
+
+    @Nonnull
+    public static Jenkins getActiveInstance() {
```
batmat: Add `Thread.sleep(2000);` here as you're deprecating. I'm actually serious :).

Then release, then make it from `2000` to `10000` and so on. A great way to push users to not use this, and finally be able to remove it.

(note: not my idea, I saw that on Twitter some time ago and found this is an awesome idea, and I'd like to see how/if we can put in the hopefully upcoming _JEP-DEPRECATED` :-))

PR #138, opened 2018-03-09
removed deprecated Helper class

escoem:
Restoring situation after #137 
Class was removed in #127


batmat:
Should we really do that this soon? I mean, it was not _supposed_ to be used before already. 

So given the past experience, I'd rather have an intermediate step using `Thread.sleep(10000)` or more as described in https://github.com/jenkinsci/support-core-plugin/pull/137/files#r173302451 so that it can show up where possibly still in use?

Not feeling strongly in any case: the previous usage was really wrong, and before it was `restricted` already anyway, so it should have failed the build from the ground up...

aheritier:
1 year is enough @batmat @MRamonLeon ?

varyvol:
@aheritier @batmat I think this should be safe by now.

batmat:
How about we file a separate PR to mark this class `@Restricted(NoExternalUse.class)` and release an intermediate version, test it on various testing pipelines we have and tools that integrate with `support-core`, and then merge & release this one?

Sorry to be picky, but `support-core` is quite critical in the end :)? (and also because we did the breakage once already :))

batmat:
And/or do what I wrote in https://github.com/jenkinsci/support-core-plugin/pull/138#issuecomment-371785552 with the `Thread.sleep`...

varyvol:
@escoem I know it was a long time ago, but maybe you could tackle @batmat 's suggestion?

varyvol:
Closing due to inactivity. Please feel free to reopen if you work on this again.

PR #140, opened 2018-03-28
[JENKINS-50428] Take into account the now configurable location for (tasks) logging

batmat:
Downstream to https://github.com/jenkinsci/jenkins/pull/3367

https://issues.jenkins-ci.org/browse/JENKINS-50428

@reviewbybees


src/main/java/com/cloudbees/jenkins/support/impl/JenkinsLogs.java
```diff
@@ -128,6 +128,22 @@ private void addOtherMasterLogs(Container result) {
         }
     }
 
+    /**
+     * Returns the root directory for logs (historically always found under <code>$JENKINS_HOME/logs</code>.
+     * Configurable since Jenkins 2.114.
```
escoem: currently support-core depends on 2.107, can it be a problem?

batmat: Nope, the code is just looking for a sysprop. 

So, yes if someone sets this on Jenkins < 2.114, not sure we should take that in account though (?). The docs clearly states this property is only available >= 2.114: https://wiki.jenkins.io/display/JENKINS/Features+controlled+by+system+properties

batmat: Had offline feedback from @escoem, this ^ addresses his concern. So going to merge.

PR #141, opened 2018-04-12
[JENKINS-50765] Defend SecretHandler from XXE attacks

dwnusbaum:
See [JENKINS-50765](https://issues.jenkins-ci.org/browse/JENKINS-50765).

This does not seem to be an issue in practice, but hardening the code won't hurt.

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
+++ b/src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
@@ -9,6 +9,7 @@ import org.xml.sax.InputSource;
 import org.xml.sax.SAXException;
 import org.xml.sax.XMLReader;
 
+import javax.xml.XMLConstants;
 import javax.xml.transform.OutputKeys;
 import javax.xml.transform.Result;
 import javax.xml.transform.Source;
@@ -16,6 +17,7 @@ import javax.xml.transform.Transformer;
 import javax.xml.transform.TransformerException;
 import javax.xml.transform.TransformerFactory;
 import javax.xml.transform.sax.SAXSource;
+import javax.xml.transform.sax.SAXTransformerFactory;
 import javax.xml.transform.stream.StreamResult;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
@@ -45,7 +47,7 @@ class SecretHandler {
      * FALLBACK will be disable in case you define a system property like -Dsupport-core-plugin.SecretHandler.ENABLE_FALLBACK=false
      * Otherwise will be enabled.
      */
-    private static boolean ENABLE_FALLBACK = !StringUtils.equalsIgnoreCase(System.getProperty("support-core-plugin.SecretHandler.ENABLE_FALLBACK", "TRUE"), "FALSE");
+    /* package */ static boolean ENABLE_FALLBACK = !StringUtils.equalsIgnoreCase(System.getProperty("support-core-plugin.SecretHandler.ENABLE_FALLBACK", "TRUE"), "FALSE");
 
     /**
      * find the secret in the xml file and replace it with the place holder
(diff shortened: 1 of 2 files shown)


src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
```diff
@@ -90,15 +92,17 @@ public void characters(char[] ch, int start, int length) throws SAXException {
...
-        Transformer transformer = TransformerFactory.newInstance().newTransformer();
+        TransformerFactory factory = TransformerFactory.newInstance();
+        factory.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
+        Transformer transformer = factory.newTransformer();
         //omit xml declaration because of https://bugs.openjdk.java.net/browse/JDK-8035437
         transformer.setOutputProperty(OutputKeys.OMIT_XML_DECLARATION, "yes");
         transformer.setOutputProperty(OutputKeys.ENCODING, OUTPUT_ENCODING);
 
         try {
+            Source src = createSafeSource(xr, new InputSource(new StringReader(str)));
```
dwnusbaum: Should we move this out of the try/catch so that if `createSafeSource` throws a `TransformerException` (should only happen if the underlying XML lib doesn't support the `http://xml.org/sax/features/external-*-entities` features and the file has an XXE) and is unable to remove the external entities it won't be added to the bundle? Currently the file would be added with the XXE intact but unexecuted.

dwnusbaum:
The first CI failure was a relevant test failure, which should be fixed in 89a01e0. The latest CI build is timing out while checking out SCM on Windows, so I'll try to rerun it in a few hours in case it is infra-related.

src/test/java/com/cloudbees/jenkins/support/configfiles/SecretHandlerTest.java
```diff
@@ -88,4 +91,28 @@ public void shouldPutAPlaceHolderInsteadOfSecret() throws Exception {
...
+        FileUtils.writeStringToFile(file, xxeXml);
+        boolean fallbackEnabled = SecretHandler.ENABLE_FALLBACK;
+        try {
+            // Disable the fallback so TransformerExceptions are always thrown.
+            SecretHandler.ENABLE_FALLBACK = false;
+            String redactedXxeXml = SecretHandler.findSecrets(file);
+            // If there is no exception than the XML library should have removed the entity without processing it.
+            assertThat(redactedXxeXml, containsString("<xxx/>"));
+        } catch (TransformerException e) {
+            assertThat(e.getMessage(), containsString("Refusing to resolve entity"));
```
Wadeck: The expected execution flow is not easily understandable and so the comment becomes mandatory to be read.

WDYT about this approach that explicitly show the flow ? (if/else against try/catch)

```
String redactedXxeXml = null;
TransformerException exceptionIfNotProcessable = null;
try {
    // Disable the fallback so TransformerExceptions are always thrown.
    SecretHandler.ENABLE_FALLBACK = false;
    redactedXxeXml = SecretHandler.findSecrets(file);
} catch (TransformerException e) {
    exceptionIfNotProcessable = e;
} finally {
    SecretHandler.ENABLE_FALLBACK = fallbackEnabled;
}

if(redactedXxeXml == null){
    // the parser supports the secure features
    assertNotNull(exceptionIfNotProcessable);
    assertThat(exceptionIfNotProcessable.getMessage(), containsString("Refusing to resolve entity"));
}else{
    // If there is no exception than the XML library should have removed the entity without processing it.
    assertNull(exceptionIfNotProcessable);
    assertThat(redactedXxeXml, containsString("<xxx/>"));
}
```

With that approach we tell the reader that both path are possible. With the try/catch approach, you need to read the comment to understand that actually it's not the last assert that can throw the exception but the findSecrets.

It's just a matter of perspective, do not feel forced to follow my subjective PoV.

src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
Wadeck: Ideally we want to give the support a bundle with as much information as possible, especially if there are some potential attacks in place. But OTOH we do not want some automatic parsing tool from the support to trigger the attack and so become vulnerable to support bundle. From my PoV the best approach if we know there is a XXE in a XML, to put inside the bundle but either with its content commented out or the extension changed to something like `.dangerous_xml`.

Otherwise, not inserting the file seems a safer approach. We do not know for what usage this bundle could be and so better safe than sorry.

src/test/java/com/cloudbees/jenkins/support/configfiles/SecretHandlerTest.java
dwnusbaum: I agree that it is confusing to keep track of what is going on here. I will refactor it similarly to your recommendation to make it easier to understand the various cases.

dwnusbaum:
@Wadeck Combining your suggestions in 0f0f64b makes the logic a lot simpler. Either the XML library removes the XXE, or we replace it with a placeholder in the custom EntityResolver (instead of throwing a TransformerException).

src/test/java/com/cloudbees/jenkins/support/configfiles/SecretHandlerTest.java
```diff
@@ -88,4 +91,22 @@ public void shouldPutAPlaceHolderInsteadOfSecret() throws Exception {
         FileUtils.writeStringToFile(file, xml);
         String patchedXml = SecretHandler.findSecrets(file);
         assertEquals(expectedXml, patchedXml);
-    }}
+    }
+
+    @Test
+    public void shouldNotResolveExternalEntities() throws Exception {
```
christ66: Annotate with jira issue: `@Issue("JENKINS-50765")`

dwnusbaum: Fixed by 1de266c

PR #142, opened 2018-05-01
[JENKINS-21670] Add support for anonymizing support bundle output

jvz:
See [JENKINS-21670](https://issues.jenkins-ci.org/browse/JENKINS-21670)

This adds the extension points `ContentFilter` and `SensitiveNameProvider` along with sensible default implementations. The SupportPlugin has been updated to filter all non-binary files through these filters when enabled.

Proposed changelog entries:

* Adds support for anonymizing predictable support bundle output contents. This allows for a support bundle to be submitted without exposing internal names while still allowing for a way to debug the output.
* _Internal:_ Adds extension points `ContentFilter` and `SensitiveNameProvider` for implementing the above mentioned feature.
* _Internal:_ Adds an [editorconfig](http://editorconfig.org/) file to help automate cross-editor style settings such as files ending in a newline, charset encoding, tabs/spaces settings, line ending formats, and more.

@reviewbybees @dwnusbaum


dwnusbaum:
AIUI this is a different approach to #132 that should hopefully cut down on the number of files that need to be changed and will work without other plugins needing to be updated.

src/main/java/com/cloudbees/jenkins/support/filter/FilteredOutputStream.java
```diff
@@ -0,0 +1,192 @@
...
+/**
+ * Wraps an OutputStream by filtering written lines using a provided ContentFilter.
+ * Data written to an instance of this will be decoded on the fly using the provided charset, and each line
+ * written is filtered.
+ *
+ * @author Matt Sicker
+ * @see ContentFilter
+ * @since 2.48
+ */
+public class FilteredOutputStream extends FilterOutputStream {
```
Wadeck: On one hand, I think this class has too much complexity for something "simple" but OTOH, I didn't find any existing class doing that and do not have a better approach.

Perhaps separate into multiple classes, one that convert bytes to characters and the second that takes characters and separate them by line and finally the one that filter whole line using the ContentFilter. (Or perhaps just merge the two last). Just an idea to reduce the complexity by using composition, but not sure it's 100% applicable here.

src/main/java/com/cloudbees/jenkins/support/filter/ContentFilter.java
```diff
@@ -0,0 +1,55 @@
...
+
+import hudson.ExtensionPoint;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Provides a strategy to filter support bundle written contents. This is primarily useful to anonymize data written
+ * to the bundle, though more complex filtering can be achieved.
+ *
+ * @author Matt Sicker
```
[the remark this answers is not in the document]
dwnusbaum: Yeah I think generally it is discouraged because git blame does almost the same thing but is more likely to be helpful.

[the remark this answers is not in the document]
dwnusbaum: Well I'm not sure if that is the "official" reason but it seems like a good reason to me 😄

src/main/java/com/cloudbees/jenkins/support/filter/DefaultContentFilter.java
```diff
@@ -0,0 +1,270 @@
...
+import hudson.model.Saveable;
+import hudson.security.Permission;
+import jenkins.model.Jenkins;
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.kohsuke.randname.RandomNameGenerator;
+import org.kohsuke.stapler.DataBoundSetter;
+import org.kohsuke.stapler.StaplerRequest;
+import org.kohsuke.stapler.StaplerResponse;
+import org.kohsuke.stapler.interceptor.RequirePOST;
+import org.slf4j.Logger;
```
dwnusbaum: I think you should use `java.util.logging` for consistency.

src/main/java/com/cloudbees/jenkins/support/filter/DefaultContentFilter.java
```diff
@@ -0,0 +1,270 @@
...
+
+/**
+ * Filters contents based on names provided by all {@linkplain SensitiveNameProvider known sources}.
+ *
+ * @author Matt Sicker
+ * @see SensitiveNameProvider
+ * @since 2.48
+ */
+@Extension
+public class DefaultContentFilter extends ManagementLink implements ContentFilter, Saveable {
```
dwnusbaum: For new public classes which aren't intended to be part of the public API of the plugin you should add `@Restricted(NoExternalUse.class)`.

src/main/java/com/cloudbees/jenkins/support/filter/DefaultSensitiveNameProviders.java
```diff
@@ -0,0 +1,148 @@
...
+        }
+
+        @Nonnull
+        @Override
+        public String prefix() {
+            return "user";
+        }
+    }
+
+    @Extension(ordinal = -100)
```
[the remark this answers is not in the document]
jvz: Great point, thanks. I did have such a comment elsewhere but forgot to migrate it when refactoring to this solution.

src/main/java/com/cloudbees/jenkins/support/filter/DefaultSensitiveNameProviders.java
```diff
@@ -0,0 +1,148 @@
...
+
+import javax.annotation.Nonnull;
+import java.util.Arrays;
+import java.util.stream.Stream;
+
+/**
+ * @author Matt Sicker
+ * @since 2.48
+ */
+public final class DefaultSensitiveNameProviders {
```
dwnusbaum: nit: Maybe cleaner to put the inner classes directly in SensitiveNameProvider and delete this file.

src/main/java/com/cloudbees/jenkins/support/filter/FilteredOutputStream.java
jvz: I think what I initially tried was essentially implementing this as a `Writer`, then using a chain like `new WriterOutputStream(new AnonymizedWriter(new OutputStreamWriter(out)))` which started to get silly.

src/main/resources/com/cloudbees/jenkins/support/filter/DefaultContentFilter/index.jelly
```diff
@@ -0,0 +1,60 @@
...
+            <h2>Mappings</h2>
+            <table class="pane sortable bigtable">
+                <tr>
+                    <th class="pane-header">Actual Name</th>
+                    <th class="pane-header" initialSortDir="up">Anonymized Name</th>
+                </tr>
+                <j:forEach items="${it.mappings}" var="item">
+                    <tr>
+                        <td class="pane"><st:out value="${item.key}"/></td>
+                        <td class="pane"><st:out value="${item.value}"/></td>
```
Wadeck: FYI the `<st:out` tag is not required here as jelly is escaping the variables (due to the first line)

But take care, OTOH the `<j:out` cannot be used here because in such case you open the way for injection. That's a bit confusion between `st:out` and `j:out` and just wanted to highlight that point.

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -428,29 +428,27 @@ public static void threadDumpStartup() throws Exception {
             @Override
             public void run() {
                 while (true) {
-                    final Jenkins jenkins =Jenkins.getInstance();
+                    final Jenkins jenkins = Jenkins.getInstanceOrNull();
                     if (jenkins == null || jenkins.getInitLevel() != InitMilestone.COMPLETED) {
```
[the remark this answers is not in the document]
Wadeck: Yeah concerning the `get()` and `getInstanceOrNull` that's perfectly fine 👍

src/main/java/com/cloudbees/jenkins/support/filter/ContentFilter.java
```diff
@@ -0,0 +1,55 @@
...
+ * @since 2.48
```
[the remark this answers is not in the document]
Wadeck: For the moment it's not automatic but done manually by the person that release, sometime you have some remaining `@since TODO`, so that's not perfect but sufficiently nice

src/main/java/com/cloudbees/jenkins/support/filter/MasterContentFilter.java
```diff
@@ -0,0 +1,54 @@
...
+
+import javax.annotation.Nonnull;
+
+/**
+ * Provides a ContentFilter that aggregates and applies all registered filters.
+ *
+ * @author Matt Sicker
+ * @since 2.48
+ */
+public enum MasterContentFilter implements ContentFilter {
```
dwnusbaum: I would just make these static methods on `ContentFilter` and delete this class.

src/test/resources/test-input.txt
```diff
@@ -0,0 +1,1000 @@
+Line 0
+Line 1
+Line 2
+Line 3
```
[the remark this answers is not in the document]
dwnusbaum: Seems like it would be fine to just create the file dynamically in the test.

src/main/java/com/cloudbees/jenkins/support/filter/DefaultContentFilter.java
```diff
@@ -0,0 +1,270 @@
...
+        return map;
+    }
+
+    public Set<String> getStopWords() {
+        return stopWords;
+    }
+
+    @CheckForNull
+    @Override
+    public Permission getRequiredPermission() {
```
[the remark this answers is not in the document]
Wadeck: TL;DR: `@Nullable` for a value that could be null under specific circumstances compared to `@CheckForNull` that allows the value to be null in any circonstance.

---

The meaning is different and always confusing, already had discussion about that with Daniel and Devin.

The conclusion was: `@CheckForNull` implies you need to check the value before using, and `@Nullable` means the `null` value is totally fine. In general with `@Nullable` you have either an implicit way to determine the case the value is null depending on the arguments of the method or the "rule" is documented. 
Imagine you have a method `UserUtil.getFirstName(User user)`, if you specify that the firstName is always non-null when the user is non-null, so you can put `@Nullable` on the method, and in your usage, you check for the user to be null, otherwise you know the firstName will be non-null.

I read also that the `@Nullable` was just created to override a method that was `@Nullable` in its parent / interface, but that seems against OOP.

PS: `@Nullable` is not taken into consideration by FindBugs compared to `@CheckForNull` as it cannot understand the rule.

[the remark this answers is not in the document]
dwnusbaum: I think FindBugs requires null checks for things annotated with `@CheckForNull` but not for `@Nullable`, so `@CheckForNull` is preferred. I'm not 100% sure though.

src/test/resources/test-input.txt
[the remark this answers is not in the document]
Wadeck: Or just put a comment in the test to explicitely say you want to check for large input.

src/main/java/com/cloudbees/jenkins/support/filter/FilteredOutputStream.java
Wadeck: If you can do that within 5-10 lines it's a lot clearer than implementing the things at the byte level from my PoV. Just don't loose hours on that :)

src/main/java/com/cloudbees/jenkins/support/filter/DefaultContentFilter.java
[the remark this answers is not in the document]
jvz: Oh interesting, I think I understand.

src/test/java/com/cloudbees/jenkins/support/filter/DefaultContentFilterTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+        String actual = filter.filter(name);
+        assertEquals(name, actual);
+
+        filter.setEnabled(true);
+
+        actual = filter.filter(name);
+        assertNotEquals(name, actual);
+    }
+}
```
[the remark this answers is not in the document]
jvz: I blame IntelliJ! I'll add an editorconfig file to help automate that.

src/main/java/com/cloudbees/jenkins/support/filter/FilteredOutputStream.java
jvz: Playing with charset encoders/decoders and byte buffers is like the majority of the work log4j spends its CPU time doing. Too much practice. ;)

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -335,26 +341,23 @@ public void add(@CheckForNull Content content) {
...
+                    final String name = filter.filter(c.getName());
+                    final ZipArchiveEntry entry = new ZipArchiveEntry(name);
+                    entry.setTime(c.getTime());
                     try {
-                        ZipEntry entry = new ZipEntry(name);
-                        entry.setTime(c.getTime());
-                        zip.putNextEntry(entry);
-                        c.writeTo(bos);
+                        zip.putArchiveEntry(entry);
+                        if (name.endsWith(".png")) {
```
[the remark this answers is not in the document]
jvz: The png files are graphs. There are associated csv files. Also, text files include .log, .log.1, .log.2, etc., along with some other extensions. Seemed easier to make a blacklist.

src/test/java/com/cloudbees/jenkins/support/filter/DefaultContentFilterTest.java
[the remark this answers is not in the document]
Wadeck: If you find a way to add that, share ^^
Otherwise, just looking at your PR for the red symbol at end of file

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -335,26 +341,23 @@ public void add(@CheckForNull Content content) {
         }
         toProcess.add(new StringContent("manifest.md", manifest.toString()));
         try {
-            ZipOutputStream zip = new ZipOutputStream(outputStream);
-            try {
-                BufferedOutputStream bos = new BufferedOutputStream(zip, 16384) {
-                    @Override
-                    public void close() throws IOException {
```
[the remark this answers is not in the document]
jvz: Good catch, but I moved that logic to AnonymizedOutputStream as it was logically necessary there anyways due to internal buffering.

src/main/java/com/cloudbees/jenkins/support/filter/DefaultSensitiveNameProviders.java
jvz: Heh, if only Java let you put multiple top level classes in a single file like Scala and Kotlin do.

src/main/java/com/cloudbees/jenkins/support/filter/DefaultContentFilter.java
jvz: I'd prefer to use log4j-api, but it's not a dependency yet, and I'd rather discuss a larger JEP or whatever to address a mass migration. As for this use case, well, I'll figure out how to refactor to that.

src/main/resources/com/cloudbees/jenkins/support/filter/DefaultContentFilter/index.jelly
Wadeck: Concerning that, in general the escape tag is placed after the license, to keep all the code "close", to avoid such situation where you forgot that you have that line.

src/main/java/com/cloudbees/jenkins/support/filter/DefaultContentFilter.java
jvz: To me, the user value will be that Jenkins and plugins themselves will be able to more easily log things (with more contextual info) that can make troubleshooting and debugging easier. However, I think I'll need to get more familiar with Jenkins at large to devise a proper proposal.

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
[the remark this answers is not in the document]
dwnusbaum: I wonder if a decent way to avoid issues would be to check the first X bytes (maybe 10) and if there are ascii control characters other than tab, line feed, vertical tab, form feed, and carriage return (`0x09`-`0x0D`) don't filter the file. On a somewhat related note I am worried about mangling files using encodings other than UTF-8.

src/test/java/com/cloudbees/jenkins/support/filter/FilteredOutputStreamTest.java
```diff
@@ -0,0 +1,78 @@
...
+    public void shouldModifyStream() throws IOException {
+        assertNotNull(testInput);
+        ContentFilter filter = s -> s.replace("Line", "Network");
+        FilteredOutputStream out = new FilteredOutputStream(testOutput, filter);
+
+        IOUtils.copy(testInput, out);
+        String contents = new String(testOutput.toByteArray(), UTF_8);
+
+        assertNotEquals(0, contents.length());
+        String[] lines = contents.split("\n");
```
dwnusbaum: I think changing this to `contents.split("\r?\n")` will fix the build failure on Windows.

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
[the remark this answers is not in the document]
dwnusbaum: Yes exactly, all of this will go wrong pretty quickly if things aren't UTF-8. Maybe a good idea to look at the components that use FileContent to see if any of their encodings can be predicted based on other information (e.g. will agent logs always use `Charset#defaultEncoding` from the agent? If so we could use that encoding for that file)

.editorconfig
```diff
@@ -0,0 +1,6 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
```
Wadeck: Perhaps this addition requires a changelog entry to provide more information to other developers, like link to some popular IDE support of EditorConfig.

src/main/java/com/cloudbees/jenkins/support/filter/ContentFilter.java
```diff
@@ -32,19 +33,45 @@
...
+ * @since TODO
  */
 public interface ContentFilter extends ExtensionPoint {
 
+    /**
+     * @return all ContentFilter extensions
+     */
+    static ExtensionList<ContentFilter> all() {
+        return ExtensionList.lookup(ContentFilter.class);
+    }
```
Wadeck: Not used outside of the interface, perhaps better to put the `lookup` method directly inside the `ofAll`
Or wait Java 9 for the private method inside interface ;)

jvz: In Scala, I would just make the companion class implement the trait/interface as well, but Java doesn't offer that. However, I did make this `all()` method to follow the standard extension point pattern.

.editorconfig
jvz: Ah, good point. I'll update the changelog.

src/main/java/com/cloudbees/jenkins/support/filter/DefaultContentFilter.java
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    @DataBoundSetter
+    public void setEnabled(boolean enabled) {
+        this.enabled = enabled;
+        save();
+    }
+
+    public Map<String, String> getMappings() {
+        Map<String, String> map = new HashMap<>(mappings.size());
```
kuisathaverat: maybe it can return an unmodifiable Map

src/main/java/com/cloudbees/jenkins/support/filter/DefaultContentFilter.java
```diff
@@ -0,0 +1,277 @@
...
+        for (Map.Entry<String, Replacer> entry : mappings.entrySet()) {
+            map.put(entry.getKey(), entry.getValue().replacement);
+        }
+        return map;
+    }
+
+    public Set<String> getStopWords() {
+        return stopWords;
```
kuisathaverat: Is it need to add stopWords from outside? if not it would be an unmodifiable Set

jvz: I changed it to be unmodifiable. I initially made this modifiable because it would allow a power user to change the value in the console, but now you can just edit the config file itself, so it's not really necessary.

src/main/java/com/cloudbees/jenkins/support/filter/DefaultContentFilter.java
```diff
@@ -0,0 +1,277 @@
...
+        proxy.setMappings(mappings);
+        proxy.setStopWords(new HashSet<>(stopWords));
+        try {
+            file.write(proxy);
+        } catch (IOException e) {
+            logWarning(e, "Could not save file %s", file.getFile().getName());
+        }
+    }
+
+    private synchronized void load() {
```
kuisathaverat: it is only called from the constructor and is private, Should it be synchronized?

jvz: I think a better question I had here was whether or not I should make this method public. It was public at one point, but I minimized access once I wasn't using it outside the constructor. The synchronization would be technically unnecessary in the constructor though, indeed.

src/main/java/com/cloudbees/jenkins/support/filter/DefaultContentFilter.java
```diff
@@ -185,26 +179,31 @@ private synchronized void load() {
...
             }
         } catch (IOException e) {
-            LOGGER.warn("Could not load file {}", file, e);
+            logWarning(e, "Could not load file %s", file.getFile().getName());
         }
         reload();
     }
 
-    @CheckForNull
+    private static void logWarning(@CheckForNull Throwable e, @Nonnull String format, Object... params) {
```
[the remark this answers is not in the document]
jvz: Oh wow, that was why I was using a real logging api before. Will fix.

src/main/java/com/cloudbees/jenkins/support/filter/DefaultContentFilter.java
```diff
@@ -185,26 +179,31 @@ private synchronized void load() {
...
+        LogRecord record = new LogRecord(Level.WARNING, String.format(format, params));
+        if (e != null) {
+            record.setThrown(e);
+        }
+        LOGGER.log(record);
+    }
+
     @Override
-    public String getIconFileName() {
+    public @CheckForNull String getIconFileName() {
```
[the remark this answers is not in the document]
jvz: I think I misunderstood the annotation usage here. Thanks for the explanation.

jvz:
I'm intrigued by the stream of patterns idea to support IP addresses. This may need some refactoring from string matching to patterns, though. I'll see how feasible it is.

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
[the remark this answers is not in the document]
jvz: Perhaps it may be worth refactoring a bit to use Writers instead of OutputStreams where possible?

jvz:
I've updated everything, though I can't remove the wip label.

src/main/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilter.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @Override
+    public synchronized void reload() {
+        ContentMappings mappings = ContentMappings.get();
+        Set<String> stopWords = mappings.getStopWords();
+        for (NameProvider provider : NameProvider.all()) {
+            provider.names()
+                    .filter(name -> !stopWords.contains(name.toLowerCase(Locale.ENGLISH)))
+                    .forEach(name -> mappings.getMappingOrCreate(name, original -> ContentMapping.of(original, provider.generateFake())));
```
dwnusbaum: I think you want to use a `BulkChange` on `ContentMappings` around this loop so that the first time through we don't save the file for each item/user in Jenkins.

jvz: Great point! I knew I forgot to do this somewhere. I was thinking of refactoring the writeBundle methods into its own class but got distracted by not wanting to balloon this PR any larger. Will fix.

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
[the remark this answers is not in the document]
jvz: This is making me think that some cleanups in SupportPlugin would be worth it just for the better charset usage.

src/test/java/com/cloudbees/jenkins/support/filter/InetAddressContentFilterTest.java
```diff
@@ -0,0 +1,82 @@
...
+    @Parameters(name = "{0}")
+    public static Object[][] data() throws UnknownHostException {
+        return new Object[][] {
+                {"127.0.0.1", true},
+                {"0.0.0.0", true},
+                {"256.0.0.1", false},
+                {"255.255.255.255", true},
+                {"http://192.168.10.100/", true},
+                {"192:168::1", true},
+                {"ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff", true},
```
Wadeck: Due the complexity of the regex, could you have "lots" more samples to cover the regex ?

From my PoV it seems way too complicated for "just" finding ipv6 and not "validating correct" ones. WDYT ?

dwnusbaum: I agree here that it seems simpler to just include anything that looks like an IP address ignoring validity if it makes things significantly easier to read and maybe increases matching performance (since we will run these patterns against every log line). For example, `\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}` matches all IPv4 addresses and a bunch of invalid addresses like 999.999.999.999, but is much simpler than the current regex.

jvz: I started writing my own regex, but then I realized that matching number ranges outside 0-9 are a pain, so I looked up existing patterns for IPv4 and IPv6 from Stack Overflow and selected the more reasonable ones.

I can certainly improve the test coverage of these patterns, though I'd likely change this test to use property testing instead (a bit too complex to attempt a formal proof hehe).

jvz:
@dwnusbaum those changes are part of an overall audit I'm doing of charset usage to ensure no text files are being written with improper character sets. After reviewing what the commit looks like, though, you have a great point. I'll revert it from this branch and move further cleanups like that to a separate branch.

src/test/java/com/cloudbees/jenkins/support/filter/InetAddressContentFilterTest.java
Wadeck: > from Stack Overflow

I dunno if it's a general good practice or just my habits, but in such case, I add a comment with the source (post, comment, blog post). This could help future maintainer (or future myself) if they encounter some troubles with the code, because potentially someone has replied to the initial source article with some bugs / corrections / etc.

jvz: Oops, I usually remember to add a link to SO posts where applicable. I'll go find them again and add comments, thanks for the reminder.

src/main/java/com/cloudbees/jenkins/support/filter/FilteredWriter.java
```diff
@@ -0,0 +1,138 @@
...
+
+    @Override
+    public synchronized void write(@Nonnull char[] cbuf, int off, int len) throws IOException {
+        while (len > 0) {
+            if (!buf.hasRemaining()) filterFlushLines();
+            if (!buf.hasRemaining()) {
+                buf = CharBuffer.allocate(buf.capacity() + DEFAULT_BUFFER_CAPACITY).put(buf);
+            }
+            int toCopy = Math.min(buf.remaining(), len);
+            if (toCopy == 0) throw new IllegalStateException();
```
dwnusbaum: This looks impossible but a message in the exception would be great for anyone (including ourselves) reading the code in the future.

src/main/java/com/cloudbees/jenkins/support/filter/FilteredOutputStream.java
```diff
@@ -0,0 +1,202 @@
...
+    @Override
+    public synchronized void flush() throws IOException {
+        if (buf.position() > 0) {
+            buf.flip();
+            String contents = buf.toString();
+            String filtered = contentFilter.filter(contents);
+            out.write(filtered.getBytes(charset));
+            buf.clear();
+            if (buf.capacity() > DEFAULT_DECODER_CAPACITY) {
+                buf = CharBuffer.allocate(DEFAULT_DECODER_CAPACITY);
```
dwnusbaum: I would not resize the buffer here, see my earlier comment.

src/main/java/com/cloudbees/jenkins/support/filter/FilteredOutputStream.java
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    /**
+     * Flushes the current buffered contents and filters them as is.
+     */
+    @Override
+    public synchronized void flush() throws IOException {
+        if (buf.position() > 0) {
+            buf.flip();
+            String contents = buf.toString();
```
dwnusbaum: Couldn't this be incorrect if a component calls flush when it has not written a full line? I would use `filterFlushLines` but leave any partially buffered lines buffered.

src/main/java/com/cloudbees/jenkins/support/filter/FilteredWriter.java
```diff
@@ -0,0 +1,138 @@
...
+    @Override
+    public synchronized void flush() throws IOException {
+        if (buf.position() > 0) {
+            buf.flip();
+            String original = buf.toString();
+            String filtered = contentFilter.filter(original);
+            out.write(filtered);
+            buf.clear();
+            if (buf.capacity() > DEFAULT_BUFFER_CAPACITY) {
+                buf = CharBuffer.allocate(DEFAULT_BUFFER_CAPACITY);
```
dwnusbaum: I would not resize the buffer here, see my earlier comment. Also, if you reused the buffer from the parent class couldn't you use its implementation as well?

src/main/java/com/cloudbees/jenkins/support/filter/NameProvider.java
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * Provides the names of labels. Note that this extension is given a lower priority than the others to avoid
+     * naming conflicts between labels and nodes/computers.
+     */
+    public static final @Extension(ordinal = -100) NameProvider LABELS = new NameProvider(
+            () -> Jenkins.get()
+                    .getLabels().stream()
+                    .map(Label::getDisplayName),
+            DataFaker.get().apply(faker -> "label_" + faker.space().star()));
```
dwnusbaum: I don't have a strong opinion about the classes of fake names, but it might be better to use the same class here as you do for nodes and computers since they are used in similar contexts in Jenkins.

src/main/java/com/cloudbees/jenkins/support/util/Persistence.java
```diff
@@ -0,0 +1,78 @@
...
+ * @since TODO
+ */
+@Restricted(NoExternalUse.class)
+public final class Persistence {
+
+    /**
+     * @return the default config file for the given class
+     */
+    public static @Nonnull <T extends Saveable> XmlFile getConfigFile(@Nonnull Class<T> clazz) {
+        return new XmlFile(new File(Jenkins.get().getRootDir(), clazz.getCanonicalName() + ".xml"));
```
dwnusbaum: nit: I generally prefer `Class.getName` since it is more useful for inner classes.

src/main/java/com/cloudbees/jenkins/support/util/Persistence.java
```diff
@@ -0,0 +1,78 @@
...
+        }
+    }
+
+    /**
+     * Loads a Saveable object from its default location or returns {@code null} when the file doesn't exist.
+     */
+    @SuppressWarnings("unchecked")
+    public static @CheckForNull <T extends Saveable> T load(@Nonnull Class<T> clazz) throws IOException {
+        XmlFile file = getConfigFile(clazz);
+        return file.exists() ? (T) file.read() : null;
```
dwnusbaum: Use `clazz.cast` instead of `(T)` so you can remove the unchecked cast.

src/main/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilter.java
jvz: I put the bulk change in `writeBundle` in the two relevant areas where mappings can get updated (during `filter.reload()` and while writing files to the zip stream).

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -367,23 +370,20 @@ public void close() throws IOException {
...
-                    zip.flush();
                 }
                 errorWriter.close();
                 String errorContent = errors.toString();
-                if (!StringUtils.isBlank(errorContent)) {
-                    try {
-                        zip.putNextEntry(new ZipEntry("manifest/errors.txt"));
-                        zip.write(errorContent.getBytes("utf-8"));
-                    } catch (IOException e) {
-                        // ignore
```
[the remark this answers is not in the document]
jvz: Looks like I forgot to add that while refactoring.

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -53,18 +57,17 @@
...
-import hudson.util.TimeUnit2;
 import jenkins.metrics.impl.JenkinsMetricProviderImpl;
 import jenkins.model.GlobalConfiguration;
 import jenkins.model.Jenkins;
+import jenkins.security.MasterToSlaveCallable;
 import net.sf.json.JSONObject;
 import org.acegisecurity.Authentication;
 import org.acegisecurity.context.SecurityContext;
 import org.acegisecurity.context.SecurityContextHolder;
+import org.apache.commons.compress.archivers.zip.ZipArchiveEntry;
```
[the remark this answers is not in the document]
jvz: Nothing really to be careful about. It extends the existing ZIP API and adds more zip file features that the Java API never added support for.

src/main/java/com/cloudbees/jenkins/support/filter/FilteredOutputStream.java
```diff
@@ -0,0 +1,202 @@
...
+        while (true) {
+            CoderResult result = decoder.decode(in, buf, endOfInput);
+            if (result.isUnderflow()) {
+                // keep accepting; can't decode this yet
+                break;
+            } else if (result.isOverflow()) {
+                if (!filterFlushLines()) {
+                    // unable to make space, need to resize
+                    buf.flip();
+                    buf = CharBuffer.allocate(buf.capacity() + DEFAULT_DECODER_CAPACITY).put(buf);
```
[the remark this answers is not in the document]
jvz: Makes sense to me. If lines start coming in longer than about a thousand characters, they might just be humongous lines.

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -260,136 +262,153 @@ public static void setLogLevel(Level level) {
...
-                        zip.putNextEntry(new ZipEntry("manifest/errors.txt"));
-                        zip.write(errorContent.getBytes("utf-8"));
-                    } catch (IOException e) {
-                        // ignore
+                        zip.putArchiveEntry(new ZipArchiveEntry("manifest/errors.txt"));
+                        filteredOut.write(errorContent.getBytes(StandardCharsets.UTF_8));
+                        filteredOut.flush();
+                        zip.closeArchiveEntry();
+                    } catch (IOException ignored) {
                     }
```
dwnusbaum: Please log the exception so we will at least see the failure in the logs.

src/main/java/com/cloudbees/jenkins/support/filter/FilteredOutputStream.java
```diff
@@ -0,0 +1,202 @@
...
+                int end = matcher.end();
+                String line = buf.subSequence(start, end).toString();
+                String filtered = contentFilter.filter(line);
+                out.write(filtered.getBytes(charset));
+                start = end;
+                flushed = true;
+            }
+            buf.position(start);
+            if (buf.capacity() - buf.remaining() > DEFAULT_DECODER_CAPACITY) {
+                buf = CharBuffer.allocate(DEFAULT_DECODER_CAPACITY).put(buf);
```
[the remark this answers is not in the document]
jvz: I added the resizing to avoid memory leaks. Moving it to reset makes sense, though.

src/main/java/com/cloudbees/jenkins/support/filter/FilteredWriter.java
```diff
@@ -0,0 +1,138 @@
...
+ * @since TODO
+ */
+public class FilteredWriter extends FilterWriter {
+
+    static final Pattern EOL = Pattern.compile("\r?\n");
+    static final int DEFAULT_BUFFER_CAPACITY = 1024;
+
+    private final ContentFilter contentFilter;
+    @GuardedBy("this")
+    private CharBuffer buf = CharBuffer.allocate(DEFAULT_BUFFER_CAPACITY);
```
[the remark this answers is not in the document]
jvz: An interesting idea! Since this class is somewhat of a view of the filtered output stream, it would make sense to share an underlying buffer.

src/main/java/com/cloudbees/jenkins/support/filter/FilteredOutputStream.java
jvz: The point here is to force a filtering when a flush is called. This could be in files that don't end a new line, for example.

src/main/java/com/cloudbees/jenkins/support/util/Persistence.java
jvz: But then you get file names with dollar signs in them. The canonical name uses a dot instead.

src/main/java/com/cloudbees/jenkins/support/filter/FilteredOutputStream.java
dwnusbaum: Yes, but the case where files don't end in a new line is always handled by `#close`, right?

src/main/java/com/cloudbees/jenkins/support/filter/NameProvider.java
jvz: Makes sense to me. I didn't think of that at the time.

src/main/java/com/cloudbees/jenkins/support/filter/FilteredOutputStream.java
jvz: Files don't necessarily get closed while writing them to the zip file. That would need to be forced I suppose.

dwnusbaum: Yes I was confused here thinking that we call `FilterOutputStream#close` after `content.writeTo(filteredOut)` in SupportPlugin, but that's not the case.

src/main/java/com/cloudbees/jenkins/support/configfiles/OtherConfigFilesComponent.java
```diff
@@ -23,8 +23,11 @@
 public class OtherConfigFilesComponent extends Component {
 
     private static final List<String> BLACKLISTED_FILENAMES = Arrays.asList(
+            // contains anonymized content mappings
             ContentMappings.class.getName() + ".xml",
+            // credentials.xml is handled by XmlRedactedSecretFileContent
```
dwnusbaum: credentials.xml is like your content mappings, it is never included in the support bundle because it includes sensitive info that isn't necessary for support

src/main/resources/com/cloudbees/jenkins/support/Messages.properties
```diff
@@ -28,7 +28,11 @@ SupportAction_DefaultActionBlurb=\
   In order to assist the Jenkins community to provide you with a response to your issues please consider generating \
   a bundle of the commonly requested information and consider attaching this bundle to the corresponding issue in \
   the Jenkins community issue tracker. Be mindful that attachments to the Jenkins community issue tracker are public \
-  so be extra careful not to include sensitive information.
+  so be extra careful not to include sensitive information. Note that certain metadata can be automatically \
+  anonymized when support bundle anonymization is enabled through the global configuration settings. \
+  When anonymization is enabled, the following information is replaced with automatically generated anonymized names: \
+  nodes, computers, labels, users, items, views, and IP addresses. Original names can be looked up via the \
```
dwnusbaum: I might reword this to `the names of nodes, computers, labels, users, items (including jobs), and views, and IP addresses`.

src/main/resources/com/cloudbees/jenkins/support/Messages.properties
```diff
@@ -28,7 +28,11 @@ SupportAction_DefaultActionBlurb=\
...
+  support bundle anonymization management page.
```
dwnusbaum: Would be good to add a link to that page.

jvz:
Recreating this pull request because there are too many comments and GitHub is loading slowly.

PR #143, opened 2018-05-11
[JENKINS-30468] Add API for item-specific components

dwnusbaum:
See [JENKINS-30468](https://issues.jenkins-ci.org/browse/JENKINS-30468). 

Adds an API for item-specific support bundle components, an implementation that includes the config.xml of AbstractItems, and an implementation that includes recent builds folders for Jobs.

To do:
- [ ] Decide if it makes sense for the `ItemComponent` classes to be configurable. If not, then the API could be simplified and the UI would just need to show checkboxes for the item-specific components (which would be consistent with the current global component UI).
- [ ] Add `ItemComponent` implementation for `Job` that gives a summary of the `Job` and its build history.
- [ ] Hide/disable/filter `ItemComponent`s from the UI when the user does not have the required permission.
- [ ] Allow multiple items to be added to a single support bundle each with their own set of item-specific components. Should work from whether the action is accessed from the root or a specific item.

Current screenshot:

screen shot 2018-05-11 at 15 36 50

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/SupportAction.java
+++ b/src/main/java/com/cloudbees/jenkins/support/SupportAction.java
@@ -25,23 +25,34 @@
 package com.cloudbees.jenkins.support;
 
 import com.cloudbees.jenkins.support.api.Component;
+import com.cloudbees.jenkins.support.api.ItemComponent;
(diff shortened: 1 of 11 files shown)


dwnusbaum:
I have no plans to work on this further.

PR #146, opened 2018-06-25
[WIP] [JENKINS-52168] Dockerfile is using the wrong image

alecharp:
Docker repository jenkins is deprecated for a long time.
Having the pinned notation on the plugins breaks the official scripts
to install the plugins.

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
@@ -117,7 +117,7 @@ public class AboutJenkins extends Component {
         container.add(new DisabledPlugins(disabledPlugins));
         container.add(new FailedPlugins());
 
-        container.add(new Dockerfile(activePlugins, disabledPlugins));
+        container.add(new Dockerfile());
 
         container.add(new MasterChecksumsContent());
         for (final Node node : Jenkins.getInstance().getNodes()) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
```diff
@@ -732,64 +732,27 @@ protected void printTo(PrintWriter out) throws IOException {
...
-                    if (disabledIT.hasNext()) {
-                        out.println(" \\");
-                    }
-                }
-                out.println();
-            }
-
             out.println();
 
+            out.println("COPY plugins/active.txt /plugins.txt");
```
duemir: Dockerfile is still in the `docker` subdirectory so this like most likely doesn't work since plugins is a sibling not a child.

alecharp: `docker build -t foobar -f docker/Dockerfile .` does work. Maybe I can add this as a comment in the file

duemir:
@alecharp BTW, did you see #61 ? It is in the same area.

alecharp:
ok, but I cannot contribute to #61. I can, however, change my PR to include suggestion from Jesse.

src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
duemir: OK. This means the whole bundle is going to be used as a context. Given how big some bundles can become the `.dockerignore` should be added as well in such case.

Also, how to use `Dockerfile` is documented on the plugins page https://plugins.jenkins.io/support-core and needs to be updated if we were to move forward with this new approach.

PR #147, opened 2018-07-18
[JENKINS-52639] Remove snakeyaml dependency by switching to wordnet-random-name

dwnusbaum:
Using snakeyaml without shading it is a little precarious in the current Jenkins ecosystem, see the comments in [JENKINS-50202](https://issues.jenkins-ci.org/browse/JENKINS-50202) for details.

This PR replaces the javafaker library with [wordnet-random-name](https://github.com/kohsuke/wordnet-random-name). It doesn't allow us to use different themes for different kinds of data (which might be a blocker), but it has no compile-scope dependencies.

We could also downgrade to snakeyaml 1.17 to avoid API compatibility issues for now (assuming javafaker 0.15 is compatible with snakeyaml 1.17), or we could shade it and the javafaker library to avoid conflicts with other plugins, but getting rid of it seemed like the easiest long-term solution.

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/filter/DataFaker.java
+++ b/src/main/java/com/cloudbees/jenkins/support/filter/DataFaker.java
@@ -37,7 +37,7 @@ import java.util.function.Supplier;
 import org.kohsuke.randname.RandomNameGenerator;
 
 /**
- * Provides a way to bind generate random names.
+ * Provides a way to generate random names.
  *
  * @since TODO
  */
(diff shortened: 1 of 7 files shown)


src/main/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilter.java
```diff
@@ -26,13 +26,13 @@
...
 import hudson.Extension;
 import hudson.ExtensionList;
-import org.apache.commons.lang3.StringUtils;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
 
 import javax.annotation.Nonnull;
 import java.util.Locale;
 import java.util.Set;
+import org.apache.commons.lang.StringUtils;
```
jvz: Yuck, I like using the current versions of Commons libraries. :)

dwnusbaum: Yeah, commons-lang has a better compatibility story than snakeyaml IIRC, so it would likely be fine to add a commons-lang3 dependency, but since commons-lang 2.6 is already bundled in Jenkins core it seems easiest to just use that to avoid any issues.

src/main/java/com/cloudbees/jenkins/support/filter/DataFaker.java
```diff
@@ -52,14 +52,14 @@ public static DataFaker get() {
         return ExtensionList.lookupSingleton(DataFaker.class);
     }
 
-    private final Faker faker = new Faker(Locale.ENGLISH);
+    private final RandomNameGenerator faker = new RandomNameGenerator();
```
varyvol: A little picky, but the fact that this variable is still called faker was a little confusing to me at the beginning.

PR #150, opened 2018-10-15
Deleting long-obsolete SlaveLogFetcher hack

jglick:
Rendered obsolete as of https://github.com/jenkinsci/jenkins/commit/a16c59bfb8b79738cde4f54783cfc432d027d7c3. I guess this plugin used to have a really old baseline.


jglick:
> Also qualifies as Java 11 compat fix (setAccesible(true))

Not really; that call was from dead code.

PR #151, opened 2018-11-18
JENKINS-54687 - GCLogs should be excluded by default

Evildethow:


Evildethow:
@duemir Yes, you are right. I've updated the definition in https://issues.jenkins-ci.org/browse/JENKINS-54687

PR #152, opened 2018-11-27
[JENKINS-54688] Operating System informations are anonymized

fcojfernandez:
See [JENKINS-54688](https://issues.jenkins-ci.org/browse/JENKINS-54688)

Due to [JENKINS-21670](https://issues.jenkins-ci.org/browse/JENKINS-21670) the Operating System information from the master node in _about.md_ or any nodes in _nodes.md_ could be anonymised if they are used as label (a classical practice), causing confusion and making difficult to get deeper information in the support bundle. This PR tries to be a quick win so adding the common operating systems' names as stop words avoid the anonymisation.

@reviewbybees spec. @varyvol @jvz @aheritier

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
+++ b/src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
@@ -95,10 +95,13 @@ public class ContentMappings extends ManagementLink implements Saveable, Iterabl
             stopWords.add(Jenkins.VERSION);
         }
 
+        // JENKINS-54688
+        //stopWords.addAll(getAllowedOSName());
+
         mappings = proxy.mappings == null
                 ? new ConcurrentSkipListMap<>(COMPARATOR)
                 : proxy.mappings.stream()
-                    .filter(mapping -> !stopWords.contains(mapping.getOriginal()))
+                    .filter(mapping -> !stopWords.contains(mapping.getOriginal().toLowerCase(Locale.ENGLISH)))
                     .collect(toConcurrentMap(ContentMapping::getOriginal, Function.identity(), (a, b) -> {throw new IllegalArgumentException();}, () -> new ConcurrentSkipListMap<>(COMPARATOR)));
     }
 
@@ -109,6 +112,13 @@ public class ContentMappings extends ManagementLink implements Saveable, Iterabl
                 "user", "anonymous", "authenticated",
                 "everyone", "system", "admin",
                 Jenkins.VERSION
+        ));
+    }
+
+    private static Set<String> getAllowedOSName() {
+        return new HashSet<>(Arrays.asList(
+                "linux", "windows", "win", "mac", "macos", "macosx",
+                "mac os x", "ubuntu", "debian", "fedora"
         ));
     }
 
--- a/src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
+++ b/src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
@@ -118,7 +118,8 @@ public class ContentMappings extends ManagementLink implements Saveable, Iterabl
     private static Set<String> getAllowedOSName() {
         return new HashSet<>(Arrays.asList(
                 "linux", "windows", "win", "mac", "macos", "macosx",
-                "mac os x", "ubuntu", "debian", "fedora"
+                "mac os x", "ubuntu", "debian", "fedora", "red hat",
+                "sunos", "freebsd"
         ));
     }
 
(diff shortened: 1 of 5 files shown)


alecharp:
what about `x64`, `arm`, etc used as label? Would it be the same issue?

fcojfernandez:
As discussed offline with @aheritier the main issue is the operating system, so they're the only considered stop words.

src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
```diff
@@ -112,6 +115,13 @@ private ContentMappings(@Nonnull XmlProxy proxy) {
         ));
     }
 
+    private static Set<String> getAllowedOSName() {
+        return new HashSet<>(Arrays.asList(
+                "linux", "windows", "win", "mac", "macos", "macosx",
+                "mac os x", "ubuntu", "debian", "fedora"
```
MRamonLeon: Could you add **SunOS** and **FreeBSD**?

src/test/resources/com/cloudbees/jenkins/support/filter/ContentMappingsTest/operatingSystemIncludedAsStopWord/com.cloudbees.jenkins.support.filter.ContentMappings.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version='1.1' encoding='UTF-8'?>
+<com.cloudbees.jenkins.support.filter.ContentMappings resolves-to="com.cloudbees.jenkins.support.filter.ContentMappings$XmlProxy">
+  <stopWords>
+    <string>authenticated</string>
+    <string>admin</string>
+    <string>mac</string>
+    <string>unknown</string>
+    <string>computer</string>
+    <string>view</string>
+    <string>linux</string>
```
varyvol: The new ones shouldn't be here so that the test proves that even when the stop words are loaded from disk, the new stop words are taken into account.

fcojfernandez:
I've added "red hat", "sunos" and "freebsd" as stop words. However, the scope of this PR is to have a quick fix, so I won't add others. Having a full coverage for all the existing combinations of OS name and distribution is imposible. It's something that have to be improved, but it's necessary a deeper analyse to figure out how to deal with the anonimisation. Let's have a quick fix for the moment!

src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
```diff
@@ -112,6 +115,14 @@ private ContentMappings(@Nonnull XmlProxy proxy) {
         ));
     }
 
+    private static Set<String> getAllowedOSName() {
+        return new HashSet<>(Arrays.asList(
+                "linux", "windows", "win", "mac", "macos", "macosx",
+                "mac os x", "ubuntu", "debian", "fedora", "red hat",
```
jvz: While technically I think using a stop word with whitespace will work, you can add the individual words.

src/test/resources/com/cloudbees/jenkins/support/filter/ContentMappingsTest/operatingSystemIncludedAsStopWord/com.cloudbees.jenkins.support.filter.ContentMappings.xml
jvz: Maybe add a comment?

src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
```diff
@@ -112,6 +115,14 @@ private ContentMappings(@Nonnull XmlProxy proxy) {
         ));
     }
 
+    private static Set<String> getAllowedOSName() {
```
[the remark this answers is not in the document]
fcojfernandez: I thought of doing that, but in fact I would like this to be a quick/temporary fix. This plugin has some performance problems and something should be improved in this way. If we analyse properly the performance issue, maybe this fix is not needed anymore, so having a separated method we can remove easily if needed.

src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
fcojfernandez: I had a look on how the stop words were invoked and I saw it's by executing the `contains` method, so I'm not sure if having the different words would work. May I be missing something?

jvz: Oh, nevermind then.

PR #153, opened 2018-11-27
Change title of existing bundles

jsoref:


varyvol:
@jsoref Given you are proposing changes that seem to have more depth, I've merged this and generated 2.52 (including another changes) so that we can start from a clean ground with all the other changes.

PR #158, opened 2018-12-04
[JENKINS-54999] Performance issue due to the bundle anonymization feature of the support-core plugin

fcojfernandez:
See [JENKINS-54999](https://issues.jenkins-ci.org/browse/JENKINS-54999)

The CPU usage is peaking when the filtering is enabled.

The CPU usage is high even in `support-core-2.47`, which is the version previous to anonymization since it's moving huge files. Anonymization implies to process huge files such as logs. The difference with previous versions is the time to process needed to anonymize those files. **That process is performed line per line and applies some filter to each line of each log and file**, so the anonymization takes a long time. This heavy process during such a long time makes the instances unresponsive. 

**Where can we touch to improve the performance?**
1. [InetAddressContentFilter](https://github.com/jenkinsci/support-core-plugin/blob/86b92654ccf5717792225714c2c7f07c43f1754d/src/main/java/com/cloudbees/jenkins/support/filter/InetAddressContentFilter.java#L72) search for any String that is susceptible to be an IP address and create an new mapping object. Even if the mapping existed previously, the mapping file is saved all the times by the [ContentMappings class](https://github.com/jenkinsci/support-core-plugin/blob/86b92654ccf5717792225714c2c7f07c43f1754d/src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java#L146) and not only when the mapping is a new one. That's a bunch of unnecessary disk accesses on write mode.
1. Both [InetAddressContentFilter](https://github.com/jenkinsci/support-core-plugin/blob/86b92654ccf5717792225714c2c7f07c43f1754d/src/main/java/com/cloudbees/jenkins/support/filter/InetAddressContentFilter.java#L77) and [SensitiveContentFilter](https://github.com/jenkinsci/support-core-plugin/blob/86b92654ccf5717792225714c2c7f07c43f1754d/src/main/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilter.java#L56) are using the method [ContentMapping#filter](https://github.com/jenkinsci/support-core-plugin/blob/master/src/main/java/com/cloudbees/jenkins/support/filter/ContentMapping.java#L95-L97) which executes the _Pattern#matcher_ and therefore compiles the pattern every time the filter is executed. This is a heavy operation and as the filter is executed line per line the process needs a long time to be executed. Since the mapping has already got the replacement String, we can substitute the Pattern objects with _StringUtils_ invocations.
1. As the logs are showing, the filtering is taking place inside the OutputStream ([FilteredOutputStream](https://github.com/jenkinsci/support-core-plugin/blob/master/src/main/java/com/cloudbees/jenkins/support/filter/FilteredOutputStream.java) and [FilteredWriter](https://github.com/jenkinsci/support-core-plugin/blob/master/src/main/java/com/cloudbees/jenkins/support/filter/FilteredWriter.java)) objects that are writing the content of the files in the final zip file. It's less expensive (from the PoV of the process time) to filter the content previously to write it on the zip file. The more Stream are needed, the more expensive is the performance.

**Comparative numbers**
Some checks using 2x85MB and 1x105Mb logs and files.
- Version 2.47: 10 sec.
- Master with anonymisation disabled: 90 sec.
- Master with anonymisation enabled: 638 sec.
- Current PR with anonymisation disabled: 90 sec.
- Current PR with anonymisation enabled: 169 sec.

@reviewbybees @varyvol @jvz @aheritier

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/api/FileContent.java
+++ b/src/main/java/com/cloudbees/jenkins/support/api/FileContent.java
@@ -27,6 +27,7 @@ package com.cloudbees.jenkins.support.api;
 import com.cloudbees.jenkins.support.SupportLogFormatter;
 import com.cloudbees.jenkins.support.filter.ContentFilter;
 import com.cloudbees.jenkins.support.filter.PrefilteredContent;
+import com.cloudbees.jenkins.support.util.StreamUtils;
 import org.apache.commons.io.IOUtils;
 
 import java.io.BufferedReader;
@@ -151,26 +152,26 @@ public class FileContent extends PrefilteredContent {
 
     // Check if the file is binary or not
     private boolean isBinary() {
-        char[] c = new char[3];
-        try (InputStreamReader reader = new InputStreamReader(new FileInputStream(file), ENCODING)) {
-            int data = reader.read();
-
-            for (int i = 0; i < 3; i++) {
-                c[i] = (char) data;
-                data = reader.read();
+        try (InputStream in = getInputStream()) {
+            long size = Files.size(file.toPath());
+            if (size == 0) {
+                // Empty file, so no need to check
+                return true;
             }
+
+            byte[] b = new byte[( size < StreamUtils.DEFAULT_PROBE_SIZE ? (int)size : StreamUtils.DEFAULT_PROBE_SIZE)];
+            int read = in.read(b);
+            if (read != b.length) {
+                // Something went wrong, so better not to read line by line
+                return true;
+            }
+
+            return StreamUtils.isNonWhitespaceControlCharacter(b);
         } catch (IOException e) {
             // If cannot be checked, then considered as binary, so we do not
             // read line by line
             return true;
         }
-
-        StringBuffer type = new StringBuffer(Character.toString(c[0]));
-        for (int i = 1; i < 2; i++) {
-            type.append(c[i]);
-        }
-
-        return (!type.toString().matches("[_a-zA-Z0-9\\-\\.]*"));
     }
 
 
--- a/src/main/java/com/cloudbees/jenkins/support/api/FileContent.java
+++ b/src/main/java/com/cloudbees/jenkins/support/api/FileContent.java
@@ -26,7 +26,7 @@ package com.cloudbees.jenkins.support.api;
 
 import com.cloudbees.jenkins.support.SupportLogFormatter;
 import com.cloudbees.jenkins.support.filter.ContentFilter;
-import com.cloudbees.jenkins.support.filter.FilteredContent;
+import com.cloudbees.jenkins.support.filter.PrefilteredContent;
 import org.apache.commons.io.IOUtils;
 
 import java.io.BufferedReader;
@@ -48,7 +48,7 @@ import java.nio.file.NoSuchFileException;
  *
  * @author Stephen Connolly
  */
-public class FileContent extends FilteredContent {
+public class FileContent extends PrefilteredContent {
 
     protected final File file;
     private final long maxSize;
--- a/src/main/java/com/cloudbees/jenkins/support/api/FileContent.java
+++ b/src/main/java/com/cloudbees/jenkins/support/api/FileContent.java
@@ -25,27 +25,36 @@
 package com.cloudbees.jenkins.support.api;
 
 import com.cloudbees.jenkins.support.SupportLogFormatter;
+import com.cloudbees.jenkins.support.filter.ContentFilter;
+import com.cloudbees.jenkins.support.filter.FilteredContent;
 import org.apache.commons.io.IOUtils;
 
+import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.FilterInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.io.OutputStreamWriter;
 import java.io.PrintWriter;
+import java.nio.file.Files;
+import java.nio.file.NoSuchFileException;
 
 /**
  * Content that is stored as a file on disk.
  *
  * @author Stephen Connolly
  */
-public class FileContent extends Content {
+public class FileContent extends FilteredContent {
 
     protected final File file;
     private final long maxSize;
+    private final boolean isBinary;
+
+    private final static String ENCODING = "UTF-8";
 
     public FileContent(String name, File file) {
         this(name, file, -1);
@@ -55,6 +64,7 @@ public class FileContent extends Content {
         super(name);
         this.file = file;
         this.maxSize = maxSize;
+        this.isBinary = isBinary();
     }
 
     @Override
@@ -71,7 +81,45 @@ public class FileContent extends Content {
                 }
             }
         } catch (FileNotFoundException e) {
-            OutputStreamWriter osw = new OutputStreamWriter(os, "utf-8");
+            OutputStreamWriter osw = new OutputStreamWriter(os, ENCODING);
+            try {
+                PrintWriter pw = new PrintWriter(osw, true);
+                try {
+                    pw.println("--- WARNING: Could not attach " + file + " as it cannot currently be found ---");
+                    pw.println();
+                    SupportLogFormatter.printStackTrace(e, pw);
+                } finally {
+                    pw.flush();
+                }
+            } finally {
+                osw.flush();
+            }
+        }
+    }
+
+    @Override
+    public void writeTo(OutputStream os, ContentFilter filter) throws IOException {
+        if (isBinary || filter == null) {
+            writeTo(os);
+        }
+
+        try {
+            if (maxSize == -1) {
+                for (String s : Files.readAllLines(file.toPath())) {
+                    String filtered = filter.filter(s);
+                    IOUtils.write(filtered, os);
+                }
+            } else {
+                try (TruncatedFileReader reader = new TruncatedFileReader(file, maxSize)) {
+                    String s;
+                    while ((s = reader.readLine()) != null) {
+                        String filtered = filter.filter(s);
+                        IOUtils.write(filtered, os);
+                    }
+                }
+            }
+        } catch (FileNotFoundException | NoSuchFileException e ) {
+            OutputStreamWriter osw = new OutputStreamWriter(os, ENCODING);
             try {
                 PrintWriter pw = new PrintWriter(osw, true);
                 try {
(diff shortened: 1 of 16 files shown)


aheritier:
Excepted the test failure the change looks good to me @fcojfernandez

fcojfernandez:
I'm closing and reopening, so the PR-builder can be re-launched

src/main/java/com/cloudbees/jenkins/support/filter/InetAddressContentFilter.java
```diff
@@ -63,19 +64,23 @@ public static InetAddressContentFilter get() {
...
+                ContentMapping map = mappings.getMappingOrCreate(ip, InetAddressContentFilter::newMapping);
+                searchList.add(ip);
+                replacementList.add(map.getReplacement());
             }
         }
         String filtered = input;
-        for (ContentMapping mapping : found) {
-            filtered = mapping.filter(filtered);
+        if (!searchList.isEmpty()) {
+            filtered = StringUtils.replaceEach(input, searchList.toArray(new String[]{}), replacementList.toArray(new String[]{}));
```
jvz:
```suggestion
            filtered = StringUtils.replaceEach(input, searchList.toArray(new String[0]), replacementList.toArray(new String[0]));
```

src/main/java/com/cloudbees/jenkins/support/api/FileContent.java
```diff
@@ -97,10 +145,35 @@ protected InputStream getInputStream() throws IOException {
     }
 
     @Override
-    public long getTime()  throws IOException {
+    public long getTime() throws IOException {
         return file.lastModified();
     }
 
+    // Check if the file is binary or not
```
jvz: This is already handled in another class: https://github.com/jenkinsci/support-core-plugin/blob/master/src/main/java/com/cloudbees/jenkins/support/util/OutputStreamSelector.java

This logic should be added there to avoid double-detection.

src/main/java/com/cloudbees/jenkins/support/filter/FilteredContent.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.cloudbees.jenkins.support.api.Content;
+
+import java.io.IOException;
+import java.io.OutputStream;
+
+/**
+ * Represents some filtered content in a support bundle.
+ */
+public abstract class FilteredContent extends Content {
```
jvz: To avoid ambiguity, perhaps this should be called something like PrefilteredContent? Because it's somewhat confusing to see this named FilteredContent followed by shouldBeFiltered() returning false.

src/main/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilter.java
```diff
@@ -52,9 +56,17 @@ public static SensitiveContentFilter get() {
...
+        List<String> searchList = new ArrayList<>();
+        List<String> replacementList = new ArrayList<>();
+
         for (ContentMapping mapping : mappings) {
-            filtered = mapping.filter(filtered);
+            searchList.add(mapping.getOriginal());
+            replacementList.add(mapping.getReplacement());
+        }
+        if (!searchList.isEmpty()) {
+            filtered = StringUtils.replaceEach(input, searchList.toArray(new String[]{}), replacementList.toArray(new String[]{}));
```
jvz: Same fix as the other array usage.

src/main/java/com/cloudbees/jenkins/support/impl/SlaveLogs.java
```diff
@@ -98,31 +93,22 @@ public void addContents(@NonNull Container container) {
...
                 container.add(
-                        new PrintedContent("nodes/slave/" + node.getNodeName() + "/jenkins.log") {
+                        new LogRecordContent("nodes/slave/" + node.getNodeName() + "/jenkins.log") {
                             @Override
-                            protected void printTo(PrintWriter out) throws IOException {
+                            public Iterable<LogRecord> getLogRecords() throws IOException {
                                 Computer computer = node.toComputer();
                                 if (computer == null) {
-                                    out.println("N/A");
+                                    return Lists.reverse(Collections.emptyList());
```
jvz:
```suggestion
                                    return Collections.emptyList();
```

src/main/java/com/cloudbees/jenkins/support/filter/FilteredContent.java
fcojfernandez: Sure thing!

src/main/java/com/cloudbees/jenkins/support/api/FileContent.java
fcojfernandez: There is no double-detection since `FileContent` now extends `PrefilteredContent` and `shouldBeFiltered` will return `false`. That makes that the OutputStream will be `unfilteredOut`: https://github.com/jenkinsci/support-core-plugin/blob/128417c63460002be5bfa7411a3056fe6ab82478/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java#L304 
`unfilteredOut` is directly the binary stream: https://github.com/jenkinsci/support-core-plugin/blob/128417c63460002be5bfa7411a3056fe6ab82478/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java#L292
My intention here is to detect if the file to filter is binary or text, so in case it's binary it will not be filtered, and in case it's a text file it will be read and filtered line by line.
After having a look into `OutputStreamSelector`, I think how it's checking if the stream is binary or not is better than the one proposed here, so I will take it out from the class and put it into a utility class, so that check can be invoked from both places.

src/main/java/com/cloudbees/jenkins/support/api/FileContent.java
```diff
@@ -25,27 +25,37 @@
...
  * @author Stephen Connolly
  */
-public class FileContent extends Content {
+public class FileContent extends PrefilteredContent {
 
     protected final File file;
     private final long maxSize;
+    private final boolean isBinary;
+
+    private final static String ENCODING = "UTF-8";
```
varyvol: I assume having this upper or lower case is the same right?

src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
```diff
@@ -141,9 +141,12 @@ private ContentMappings(@Nonnull XmlProxy proxy) {
      * Looks up or creates a new ContentMapping for the given original string and a ContentMapping generator.
      */
     public @Nonnull ContentMapping getMappingOrCreate(@Nonnull String original, @Nonnull Function<String, ContentMapping> generator) {
+        boolean isNew = !mappings.containsKey(original);
         ContentMapping mapping = mappings.computeIfAbsent(original, generator);
```
varyvol: Maybe you could use a new generator to avoid having to inspect for the keys. If I am not wrong, the generator is called only if necessary, and you could create onw that would: set isNew to true and then call the real generator.

src/main/java/com/cloudbees/jenkins/support/filter/PrefilteredContent.java
```diff
@@ -0,0 +1,53 @@
...
+
+package com.cloudbees.jenkins.support.filter;
+
+import com.cloudbees.jenkins.support.api.Content;
+
+import java.io.IOException;
+import java.io.OutputStream;
+
+/**
+ * Represents some filtered content in a support bundle.
```
varyvol: Maybe a little more information about the difference between a normal content and a prefiltered one?

src/main/java/com/cloudbees/jenkins/support/filter/PrefilteredContent.java
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * Write the component in the bundle filtering the content
+     * @param os OutputStream where write the content
+     * @param filter ContentFilter to apply
+     * @throws IOException
+     */
+    public abstract void writeTo(OutputStream os, ContentFilter filter) throws IOException;
+
+    @Override
+    public boolean shouldBeFiltered() {
```
varyvol: Could this be declared final?

src/main/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilter.java
```diff
@@ -52,9 +56,17 @@ public static SensitiveContentFilter get() {
     public @Nonnull String filter(@Nonnull String input) {
         ContentMappings mappings = ContentMappings.get();
         String filtered = input;
+        List<String> searchList = new ArrayList<>();
+        List<String> replacementList = new ArrayList<>();
+
         for (ContentMapping mapping : mappings) {
-            filtered = mapping.filter(filtered);
```
varyvol: Is the filter() method in mappings used anymore?

src/main/java/com/cloudbees/jenkins/support/api/FileContent.java
```diff
@@ -97,10 +146,35 @@ protected InputStream getInputStream() throws IOException {
     }
 
     @Override
-    public long getTime()  throws IOException {
+    public long getTime() throws IOException {
         return file.lastModified();
     }
 
+    // Check if the file is binary or not
+    private boolean isBinary() {
```
varyvol: Would moving this to StreamUtils make sense?

src/main/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilter.java
fcojfernandez: I don't think so, but there are some lambda expressions where I cannot be 100% sure if it's used or not. That's why I'm not removing it.

src/main/java/com/cloudbees/jenkins/support/api/FileContent.java
fcojfernandez: Yes. Also `Charset` class waits for "UTF-8"

src/main/java/com/cloudbees/jenkins/support/api/FileContent.java
fcojfernandez: Not in my opinion. This method is processing something too specific.

src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
fcojfernandez: It's not possible because the _boolean_ should be final and its value could not be updated

src/main/java/com/cloudbees/jenkins/support/filter/PrefilteredContent.java
fcojfernandez: Sure thing!

src/main/java/com/cloudbees/jenkins/support/filter/PrefilteredContent.java
fcojfernandez: Makes sense

PR #159, opened 2019-01-09
[JENKINS-55493] Upgrade parent pom and solve JTH html API usage 

alecharp:
[JENKINS-55493](https://issues.jenkins-ci.org/browse/JENKINS-55493)

This is required to validate the plugin with PCT against Java 11 support. It will require to upgrade `metrics` plugin version requirement to include the fix from https://github.com/jenkinsci/metrics-plugin/pull/42 once released but that can be done in a follow-up PR.

@jenkinsci/java11-support

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,7 +29,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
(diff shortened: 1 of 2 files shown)


batmat:
Going to merge this later today or tomorrow if nobody objects, or does it, in the meantime :-).

Thanks!

PR #160, opened 2019-01-29
[JENKINS-55843] Tests failing when file names are anonymized

MRamonLeon:
See [[JENKINS-55843] Folders missed due to anonymization in the bundle](https://issues.jenkins-ci.org/browse/JENKINS-55843)

Created 3 tests that fail because the anonymization change the file names in the zip file. Even losing the folders in the zip.

In the zip, you should have:
jenkins-root-configuration-files **/** build-failure-analyzer.xml

But you get:
jenkins-root-configuration-files **user_steep_lung** build-failure-analyzer.xml


MRamonLeon:
Tests failing:
screenshot from 2019-01-29 22-45-55

PR #161, opened 2019-01-29
[JENKINS-55843] Folders missed due to anonymization in the bundle

MRamonLeon:
See [[JENKINS-55843] Folders missed due to anonymization in the bundle](https://issues.jenkins-ci.org/browse/JENKINS-55843) 

When a bundle is anonymized and there are objects with tricky names, the zip could become corrupt.

For example, when you create labels like '/', the folders in the zip are not created and instead files acre created. Other tricky object naes are:  '.', ':', '\', or even 'a/b'.

Instead of:
jenkins-root-configuration-files **/** build-failure-analyzer.xml 
you get:
jenkins-root-configuration-files **user_steep_lung** build-failure-analyzer.xml

The first commit add three tests to probe the failure.

**ADVISORY**: As of this PR *Content* class has a specific constructor to filter the name of the files in the bundle. Any implementor needs to use this constructor to leave it clear what needs to be filtered (dynamic names most likely) and  what doesn't. It's a change in the current behavior where the complete name is filtered and due to that, the file/folder names in the zip may become currupt. I took this approach because it's the safest one and It's not likely to be many extra implementors so far.

The other approach is to keep the former constructor filtering all the file name and take the risk to have corrupt bundles. It's just a change on a line: https://github.com/jenkinsci/support-core-plugin/pull/161/files#diff-79928e08589e1f90af352ccce49d8eb0R360

@reviewbybees @jvz @varyvol

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
+++ b/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
@@ -33,8 +33,8 @@ import com.cloudbees.jenkins.support.api.SupportProviderDescriptor;
 import com.cloudbees.jenkins.support.filter.ContentFilter;
 import com.cloudbees.jenkins.support.filter.ContentFilters;
 import com.cloudbees.jenkins.support.filter.ContentMappings;
-import com.cloudbees.jenkins.support.filter.PrefilteredContent;
 import com.cloudbees.jenkins.support.filter.FilteredOutputStream;
+import com.cloudbees.jenkins.support.filter.PrefilteredContent;
 import com.cloudbees.jenkins.support.impl.ThreadDumps;
 import com.cloudbees.jenkins.support.util.IgnoreCloseOutputStream;
 import com.cloudbees.jenkins.support.util.OutputStreamSelector;
@@ -88,17 +88,24 @@ import java.io.UnsupportedEncodingException;
 import java.nio.charset.StandardCharsets;
 import java.text.MessageFormat;
 import java.text.SimpleDateFormat;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Date;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.TimeZone;
+import java.util.TreeSet;
+import java.util.WeakHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicLong;
-import java.util.function.Function;
 import java.util.logging.Handler;
 import java.util.logging.Level;
 import java.util.logging.LogRecord;
 import java.util.logging.Logger;
-import java.util.stream.Stream;
 
 /**
  * Main entry point for the support plugin.
--- a/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
+++ b/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
@@ -342,7 +342,7 @@ public class SupportPlugin extends Plugin {
      * Filter the name of a content depending on the tokens in the name that need to be replaced.
      * @param maybeFilter an Optional with a {@link ContentFilter} or not
      * @param name the name of the content to be filtered
-     * @param tokens tokens in the name to be filtered. If null, the whole name is filtered
+     * @param tokens tokens in the name to be filtered. If null, no filter takes place to avoid corruption
      * @return the name filtered
      */
     private static String getNameFiltered(Optional<ContentFilter> maybeFilter, String name, String[] tokens) {
--- a/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
+++ b/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
@@ -86,24 +86,19 @@ import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.io.UnsupportedEncodingException;
 import java.nio.charset.StandardCharsets;
+import java.text.MessageFormat;
 import java.text.SimpleDateFormat;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Date;
-import java.util.List;
-import java.util.Optional;
-import java.util.Set;
-import java.util.TimeZone;
-import java.util.TreeSet;
-import java.util.WeakHashMap;
+import java.util.*;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Function;
 import java.util.logging.Handler;
 import java.util.logging.Level;
 import java.util.logging.LogRecord;
 import java.util.logging.Logger;
+import java.util.stream.Stream;
 
 /**
  * Main entry point for the support plugin.
@@ -296,7 +291,7 @@ public class SupportPlugin extends Plugin {
                     if (content == null) {
                         continue;
                     }
-                    final String name = maybeFilter.map(filter -> filter.filter(content.getName())).orElseGet(content::getName);
+                    final String name = getNameFiltered(maybeFilter, content.getName(), content.getTokens());
                     final ZipArchiveEntry entry = new ZipArchiveEntry(name);
                     entry.setTime(content.getTime());
                     try {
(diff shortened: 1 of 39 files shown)


MRamonLeon:
Failures of the three new tests:
screenshot from 2019-01-30 09-04-46

Error:
screenshot from 2019-01-30 09-04-57

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -343,7 +338,33 @@ public static void writeBundle(OutputStream outputStream, final List<Component>
...
+
+            // Replace each placeholder {0}, {1} in the name, with the replaced token
+            filteredName = MessageFormat.format(name, replacedTokens);
+        } else {
+            // Previous behavior was filter all the name, but it could end up in having a corrupted bundle. So we expect
+            // implementors to use the appropriate constructor of Content.
+            //filteredName = maybeFilter.map(filter -> filter.filter(name)).orElse(name);
+            filteredName = name;
+        }
+        System.out.format("%s is replaced by %s%n", name, filteredName);
```
Wadeck: Not intented to be kept in production code right? :)

src/test/java/com/cloudbees/jenkins/support/filter/ContentMappingsTest.java
```diff
@@ -165,4 +172,22 @@ public void clear() {
...
+
+        File zipFile = File.createTempFile("test", "zip");
+        ((File) zipFile).deleteOnExit();
+        FileOutputStream fileOutputStream = new FileOutputStream(tempLogFile);
+
+        ClearCaseChangeLogSet.saveToChangeLog(fileOutputStream, history);
+        fileOutputStream.close();
+
+        comprobar que el zip tiene las entradas bien
+    }*/
```
Wadeck: Dead code

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
MRamonLeon: No, sure

src/test/java/com/cloudbees/jenkins/support/filter/ContentMappingsTest.java
MRamonLeon: Why are you bringing out all my playground?

MRamonLeon:
@Wadeck The / test already cover the case of a/b. I mean, a/b is a *subtest*, meaning it covers less cases, than the / one. It happens the same with the **jenkins-root-configuration-files/build-failure-analyzer.xml** test. It's a *subtest* of the more generic /

I don't know if I'm elaborating it well enough. :man_shrugging:

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -86,24 +86,19 @@
...
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Date;
-import java.util.List;
-import java.util.Optional;
-import java.util.Set;
-import java.util.TimeZone;
-import java.util.TreeSet;
-import java.util.WeakHashMap;
+import java.util.*;
```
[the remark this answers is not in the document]
MRamonLeon: I know, I didn't realize. I fixed my IDE to avoid it in the future.

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -343,7 +338,33 @@ public static void writeBundle(OutputStream outputStream, final List<Component>
...
+     * Filter the name of a content depending on the tokens in the name that need to be replaced.
+     * @param maybeFilter an Optional with a {@link ContentFilter} or not
+     * @param name the name of the content to be filtered
+     * @param tokens tokens in the name to be filtered. If null, no filter takes place to avoid corruption
+     * @return the name filtered
+     */
+    private static String getNameFiltered(Optional<ContentFilter> maybeFilter, String name, String[] tokens) {
+        String filteredName;
+
+        if (tokens != null) {
```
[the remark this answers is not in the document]
MRamonLeon: I think it's not needed, the `MessageFormat.format` works fine in this case, so it do the job. The *user developer* should be deliberately set a new `String[]{}` to do that, so a few probabilities it happens and even so, it works.

src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
```diff
@@ -557,7 +557,7 @@ public synchronized int n() {
                 out.println("  * Mode:    WAR");
             }
             final JenkinsLocationConfiguration jlc = JenkinsLocationConfiguration.get();
-            out.println("  * Url:     " + (jlc != null ? jlc.getUrl() : "No JenkinsLocationConfiguration available"));
+            out.println("  * Url:     " + jlc.getUrl());
```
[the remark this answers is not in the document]
MRamonLeon: It was due to the change in the jenkins file, reverted

src/main/java/com/cloudbees/jenkins/support/impl/LoadStats.java
```diff
@@ -163,9 +168,31 @@ public void writeTo(OutputStream os) throws IOException {
...
+            for (Field f : FIELDS) {
+                try {
+                    MultiStageTimeSeries ts = (MultiStageTimeSeries) f.get(stats);
+                    if (ts != null) {
+                        TimeSeries series = ts.pick(scale);
+                        if (series != null) {
+                            data.put(camelCaseToSentenceCase(f.getName()), series.getHistory());
+                        }
+                    }
+                } catch (IllegalAccessException e) {
```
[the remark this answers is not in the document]
MRamonLeon: It's a copy/paste from the other constructor, so the lack of documentation was already there. I did the copy/paste because the fields are `final` and I can't use an auxiliary method.

src/main/java/com/cloudbees/jenkins/support/api/Content.java
```diff
@@ -36,14 +40,49 @@
 
     private final String name;
 
+    /**
+     * Parts of the name to be filtered
+     */
+    private String[] tokens;
```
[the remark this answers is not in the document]
MRamonLeon: What do you mean exactly @oleg-nenashev ? `@CheckForNull` perhaps?

src/test/java/com/cloudbees/jenkins/support/filter/ContentMappingsTest.java
Wadeck: I leave the same stuff in my last PR, I know at which point it's easy to forget about those stuff :D

src/main/java/com/cloudbees/jenkins/support/api/Content.java
```diff
@@ -36,14 +40,49 @@
...
+
+
+    /**
+     * Create a Content with this name. The name is not filtered so this constructor should be used exclusively when
+     * the name is not prone to have sensitive information. If the name of this content is someway dinamically generated
```
jvz:
```suggestion
     * the name is not prone to have sensitive information. If the name of this content is dynamically generated
```

src/main/java/com/cloudbees/jenkins/support/api/Content.java
jvz: Might make more sense to call this `filterableParameters` or something like that. Calling this `tokens` is confusing since you're not really tokenizing anything.

src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
```diff
@@ -110,11 +105,28 @@ private ContentMappings(@Nonnull XmlProxy proxy) {
...
+     * avoid replacing one single characteres (ascii codes actually). A one single character in other languages could
+     * have a meaning, so we remain replacing them. Example: 日 (Sun)
+     * @return Set of characters in ascii code chart
+     */
+    private static Set<String> getAllAsciiCharacters() {
+        final int SPACE = ' '; //20
+        final int TILDE = '~'; //126
+        Set<String> singleChars = new HashSet<>(TILDE - SPACE + 1);
+
+        for (int i = SPACE; i <= TILDE; i++) {
```
jvz:
```suggestion
        for (char i = SPACE; i <= TILDE; i++) {
```

Can you do that? Can remove need for casting below, and can make SPACE and TILDE chars.

src/test/java/com/cloudbees/jenkins/support/filter/ContentMappingsTest.java
```diff
@@ -25,17 +25,24 @@
 
 import hudson.model.FreeStyleProject;
 import jenkins.model.Jenkins;
+import org.apache.commons.io.IOUtils;
```
jvz: Do you need any of these imports anymore?

src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
```diff
@@ -110,11 +105,28 @@ private ContentMappings(@Nonnull XmlProxy proxy) {
...
+            singleChars.add(Character.toString((char) i));
```
MRamonLeon:
```suggestion
            singleChars.add(Character.toString(i));
```

src/test/java/com/cloudbees/jenkins/support/filter/ContentMappingsTest.java
MRamonLeon: No, I don't, Thanks Matt.

src/main/java/com/cloudbees/jenkins/support/api/Content.java
MRamonLeon: I initially called it something like that, but renamed it to something shorter. But as I'm a big fan of meaningful object names and it looks that you agree, I'll change them to your proposal. Thank you.

src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
```diff
@@ -35,15 +35,7 @@
...
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Set;
-import java.util.Spliterator;
+import java.util.*;
```
varyvol: Better not doing this in production code, especially for such package.

MRamonLeon: Definitively my IDE is not fixed yet :-/

MRamonLeon:
@varyvol This PR is about the file name in the zip only, the anonymization of the content remains as always. This PR avoids corrupting the file names in the zip due to the anonymization by allowing the implementors specify the part of the file that is anonymizable.

varyvol:
@MRamonLeon my comment still applies, you need to use the new constructor to get the new functionality so implementers won't get it by default.

MRamonLeon:
Yes, but because
a) currently the zip is becoming corrupt when Jenkins instance has too wide object names, like A or / and
b) most likely there aren't too many implementors so far
I took this approach.

In any case, it's a matter of changing one line if we want the former constructor to keep it's previous behavior, anonymizing but likely corrupting the content of the files or even the zip.

MRamonLeon:
I was thinking on alternatives, but even if we avoid replacing things like / to avoid loosing the folders in the zip, we always have the problem that the file names in the zip can be renamed due to the object names in the instance, so we need to avoid replacing them explicitly. That's the reason to add an extra constructor, to allow implementors to say what should be anonymized.

PR #163, opened 2019-02-06
[JENKINS-55866] Anonymize by words and some other fixes

MRamonLeon:
See [JENKINS-55866](https://issues.jenkins-ci.org/browse/JENKINS-55866).

When anonymizing, replace the content by words to avoid replace inside the words. Example:
- Replace: _a_
- By: _-a-_
- Input: _This is a sample_
- Actual behavior: _This is -a- s-a-mple_
- With this PR: _This is -a- sample_

Also avoid replace single characters to avoid having a bundle unreadable. Only ascii characters because  in some cultures, a single character has a meaning. Example: 日 (Sun in chinese).

Fix name on the manifest. Now they get printed filtered.

Fix slow-requests files. They were written with \0 instead of the character filtered.

Fix new line in the logs files. They were missed.

Fix replace ignoring the case and with all the flavors previously defined (/ ->  »,  \n -> <br>, ...)

Add a Chrono class to measure times for performance tests. Only on test environment.

Add a test that fails proving the bug that when you remove an object from the instance, the mapping is not removed from the persisted data.

@reviewbybees specially @varyvol @fcojfernandez @jvz


src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -430,18 +446,29 @@ private static void appendManifestHeader(StringBuilder manifest) {
...
         private final Set<String> names = new TreeSet<>();
 
+        //The filter to return the names filtered
+        private final Optional<ContentFilter> maybeFilter;
+
+        /**
+         * We need the filter to be able to filter the contents written to the manifest
+         * @param maybeFilter filter to use when writing the name of the contents
+         */
+        public ContentContainer(Optional<ContentFilter> maybeFilter) {
```
MRamonLeon: The new constructor (private class) to filter the names every time a content is added to this container. Weird architecture BTW

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -430,18 +446,29 @@ private static void appendManifestHeader(StringBuilder manifest) {
...
+            this.maybeFilter = maybeFilter;
+        }
+
         @Override
         public void add(Content content) {
             if (content != null) {
                 contents.add(content);
-                names.add(content.getName());
+                names.add(getNameFiltered(maybeFilter, content.getName(), content.getFilterableParameters()));
```
MRamonLeon: Filter the name

src/main/java/com/cloudbees/jenkins/support/api/FileContent.java
```diff
@@ -119,14 +119,18 @@ public void writeTo(OutputStream os, ContentFilter filter) throws IOException {
             if (maxSize == -1) {
                 for (String s : Files.readAllLines(file.toPath())) {
                     String filtered = filter.filter(s);
-                    IOUtils.write(filtered, os);
+                    IOUtils.write(filtered, os, ENCODING);
+                    // The new line
+                    IOUtils.write("\n", os, ENCODING);
```
MRamonLeon: Fix all the logs in a single line

src/main/java/com/cloudbees/jenkins/support/api/FileContent.java
```diff
@@ -260,7 +264,7 @@ public String readLine() throws IOException {
             len -= length;
 
             byte[] dest = new byte[toRead];
-            System.arraycopy(line.getBytes(ENCODING), 0, new byte[toRead], 0, toRead);
+            System.arraycopy(line.getBytes(ENCODING), 0, dest, 0, toRead);
```
MRamonLeon: Fix all slow-requests logs with \0 instead of the real content

src/main/java/com/cloudbees/jenkins/support/filter/ContentMapping.java
```diff
@@ -49,32 +48,45 @@
...
     private final String original;
-    private final Pattern pattern;
     private final String replacement;
     private final int hashCode;
 
-    private ContentMapping(@Nonnull String original, @Nonnull Pattern pattern, @Nonnull String replacement) {
+    private final String[] originals;
+    private final String[] replacements;
+
+    private ContentMapping(@Nonnull String original, @Nonnull String replacement) {
```
MRamonLeon: Switch to custom replacer, instead of RegExp, to get some more performance, so no need to use the pattern (private)

src/main/java/com/cloudbees/jenkins/support/filter/ContentMapping.java
```diff
@@ -49,32 +48,45 @@
...
+        addIfNotExist(originalsList, slashChangedInOriginal);
+        addIfNotExist(originalsList, Functions.escape(slashChangedInOriginal));
+        originals = originalsList.toArray(new String[0]);
+
+        // create the replacement array with the same length as the resulting originals
+        replacements = new String[originals.length];
+        for(int i = 0; i < replacements.length; i++) {
+            replacements[i] = replacement;
+        }
+
```
MRamonLeon: Add all the flavors we had with RegExp: / -> ALT_SEPARATOR, < -> &lt;, \n -> <br>, ...It was missed in the previous version.

src/main/java/com/cloudbees/jenkins/support/filter/ContentMapping.java
```diff
@@ -93,7 +105,7 @@ private static Pattern generatePattern(String original) {
 
     @Override
     public @Nonnull String filter(@Nonnull String input) {
-        return pattern.matcher(input).replaceAll(replacement);
+        return WordReplacer.replaceWordsIgnoreCase(input, originals, replacements);
```
MRamonLeon: The change in how we replace

src/main/java/com/cloudbees/jenkins/support/util/WordReplacer.java
```diff
@@ -0,0 +1,168 @@
+package com.cloudbees.jenkins.support.util;
```
MRamonLeon: The new custom class to replace by words without loosing performance. There is a set of public methods for comfort.

src/main/java/com/cloudbees/jenkins/support/util/WordReplacer.java
```diff
@@ -0,0 +1,168 @@
...
+    private static void replaceWord(StringBuilder input, String word, String replace, boolean ignoreCase) {
+        // The string where we look for
+        StringBuilder workInput;
+        // The string to find (lower case if ignoreCase)
+        String workWord = word;
+
+        // Use the lowercase versions to compare, but replace in the original input
+        if (ignoreCase) {
+            workInput = new StringBuilder(input.toString().toLowerCase(Locale.ENGLISH));
+            workWord = word.toLowerCase(Locale.ENGLISH);
```
MRamonLeon: Replacing in some languages is not so trivial, the string can change its length and we are looking backward and forward for the word separator characters.

src/test/java/com/cloudbees/jenkins/support/filter/ContentMappingsTest.java
```diff
@@ -110,6 +111,49 @@ public void contentMappingsSurviveSerializationRoundTrip() {
         });
     }
 
+    @Test
+    @Ignore("Bug to be resolved. Elements removed aren't removed from the persisted mapping (ContentMappings.xml")
+    public void contentMappingsRemovedSerialized() {
```
MRamonLeon: The test failing ignored until it is fixed

src/test/java/com/cloudbees/jenkins/support/measures/ChronoTest.java
```diff
@@ -0,0 +1,35 @@
+package com.cloudbees.jenkins.support.measures;
+
+import com.cloudbees.jenkins.support.util.Chrono;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class ChronoTest {
```
MRamonLeon: Testing the chrono class

src/test/java/com/cloudbees/jenkins/support/util/Chrono.java
```diff
@@ -0,0 +1,225 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Stream;
+
+import static java.util.concurrent.TimeUnit.DAYS;
+import static java.util.concurrent.TimeUnit.HOURS;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.util.concurrent.TimeUnit.MINUTES;
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+public class Chrono {
```
MRamonLeon: A class used for performance tests, used only in test environment.

src/test/java/com/cloudbees/jenkins/support/util/WordReplacerTest.java
```diff
@@ -0,0 +1,154 @@
...
+import org.junit.Assert;
+import org.junit.Ignore;
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class WordReplacerTest {
```
MRamonLeon: Testing the new replacer with some interesting strings

src/main/java/com/cloudbees/jenkins/support/filter/ContentMapping.java
```diff
@@ -49,32 +48,45 @@
...
+        // create the replacement array with the same length as the resulting originals
+        replacements = new String[originals.length];
+        for(int i = 0; i < replacements.length; i++) {
+            replacements[i] = replacement;
+        }
+
         this.hashCode = original.hashCode();
     }
 
+    private void addIfNotExist(List<String> list, String element) {
```
jvz: What if you used a LinkedHashMap? That would retain its insert order while also avoiding duplicates.

src/test/java/com/cloudbees/jenkins/support/measures/ChronoTest.java
```diff
@@ -0,0 +1,35 @@
...
+        halt(400);
+        c.markFromPrevious("After 400");
+        Assert.assertTrue(c.getMeasure("After 400") >= 400);
+        Assert.assertEquals(c.getMeasure("After 400"), c.getMeasures("After 400").get("After 3s from beginning"));
+
+        System.out.println(c.printMeasures());
+
+    }
+
+    private void halt(long millis) throws InterruptedException {
```
jvz: Sleep might make more sense. Halt means to stop entirely like shutting down.

src/test/java/com/cloudbees/jenkins/support/util/WordReplacerTest.java
```diff
@@ -0,0 +1,154 @@
...
+
+            // Both filtered strings should be the same
+            Assert.assertEquals("The string replaced should be the same using RegExp or WordReplacer", resultCM, resultWordReplacer);
+        }
+    }
+
+   @Ignore("It was useful to make the decision to move out of Reg Exp. As the implementation of the Content Mapping is" +
+            "now WordReplacer, it has no sense to test. We keep it here for future cases.")
+    @Test
+    public void performanceTest() {
```
jvz: This would be better handled using something like [JMH](https://openjdk.java.net/projects/code-tools/jmh/).

src/main/java/com/cloudbees/jenkins/support/filter/ContentMapping.java
```diff
@@ -49,32 +48,45 @@
     private static final String ALT_SEPARATOR = " » ";
 
     private final String original;
-    private final Pattern pattern;
```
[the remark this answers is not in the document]
MRamonLeon: Could you elaborate? I can't understand clearly what you say

src/main/java/com/cloudbees/jenkins/support/util/WordReplacer.java
```diff
@@ -0,0 +1,168 @@
...
+     * @param input the text where the replacements take place
+     * @param words the words to look for and replace
+     * @param replaces the new words to use
+     */
+    public static void replaceWordsIgnoreCase(StringBuilder input, String[] words, String[] replaces) {
+        replaceWordsIgnoreCase(input, words, replaces, true);
+    }
+
+    // To avoid repeat this code above
+    private static void replaceWordsIgnoreCase(StringBuilder input, String[] words, String[] replaces, boolean ignoreCase) {
```
varyvol: The "ignoreCase" parameter is never used?

src/main/java/com/cloudbees/jenkins/support/filter/ContentMapping.java
MRamonLeon: I didn't want to use a map, as key are not important. But yes, definitively a Set is a better interface for such purpose

src/test/java/com/cloudbees/jenkins/support/measures/ChronoTest.java
MRamonLeon: I'll remove the method and just use the `Thread.sleep` call. The method is not worth it

src/main/java/com/cloudbees/jenkins/support/util/WordReplacer.java
```diff
@@ -0,0 +1,168 @@
...
+            if (nextPos < workInput.length()) {
+                char nextChar = workInput.charAt(nextPos);
+                if (Character.isLetterOrDigit(nextChar)) {
+                    pos = pos + workWord.length();
+                    continue;
+                }
+            }
+
+            // replace this word but in the original input (without changing the case)
+            input.replace(pos, pos + workWord.length(), replace);
```
varyvol: Isn't the replacement assigned anywhere?

src/test/java/com/cloudbees/jenkins/support/util/WordReplacerTest.java
MRamonLeon: I saw it but all examples measure a full method, with annotations and I wanted to measure specific pieces of code inside methods, so I ended up doing it by myself

src/test/java/com/cloudbees/jenkins/support/util/Chrono.java
varyvol: Where is the class used?

MRamonLeon: I used this to measure inside the _writebundle_ method and other places what is taking the most of the time. I left it there for future use, as it's in test scope it doesn't go to production.

src/main/java/com/cloudbees/jenkins/support/filter/ContentMapping.java
MRamonLeon: Used a set

src/main/java/com/cloudbees/jenkins/support/util/WordReplacer.java
MRamonLeon: Right, I'm fixed some bugs in the class. Some methods are not used, I created them just in case I need them. I've also added a complete test to check all methods.

src/main/java/com/cloudbees/jenkins/support/util/WordReplacer.java
MRamonLeon: input is a **StringBuilder**, so it's changed itself with the replace method.

src/main/java/com/cloudbees/jenkins/support/filter/ContentMapping.java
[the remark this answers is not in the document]
varyvol: I think @jvz refers to the fact that these content mappings are stored into disk. If you change the format now, you should make them work with the previous format too.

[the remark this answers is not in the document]
jvz: Exactly. These are persisted, partially so that it can be configured. Now I'm sort of wondering how important it is to persist these.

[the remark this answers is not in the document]
MRamonLeon: AFAIK the `ContentMapping` is not persisted. The one persisted is `com.cloudbees.jenkins.support.filter.ContentMappings.xml`. It only persists the original and replacement strings. Perhaps I'm wrong, you've worked much more than me on this plugin.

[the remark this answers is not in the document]
varyvol: Not exactly. The ContentMappings in the end contains instances of ContentMapping, but when saving/reading from/to disk only the original and replacement fields are taken into account.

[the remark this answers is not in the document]
MRamonLeon: As agreed on the channel, it's fine because the ContentMapping already has the SerializationProxy class with the `readResolve` method used in the `writeReplace` method.

MRamonLeon:
Just for the record and to make it easy to know the status of the PR, as said before, the serialization is already covered by the _ContentMapping_  class, by implementing the proxy.

fcojfernandez:
Re- :bee:

PR #164, opened 2019-02-14
[JENKINS-55866] Fix replacement with empty strings

MRamonLeon:
See [JENKINS-55866](https://issues.jenkins-ci.org/browse/JENKINS-55866)

Previous PR was: https://github.com/jenkinsci/support-core-plugin/pull/163

This is a follow up ticket to fix the replacements with empty strings, a test and avoid flaky test that happened to me during testing.

@reviewbybees Specially: @varyvol @fcojfernandez @jvz

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/util/WordReplacer.java
+++ b/src/main/java/com/cloudbees/jenkins/support/util/WordReplacer.java
@@ -60,7 +60,7 @@ public class WordReplacer {
 
         // the same number of word to replace and replaces to use
         if (words.length != replaces.length) {
-            throw new IllegalArgumentException(String.format("Words (%d) and replaces (%d) lenghts should be equals", words.length, replaces.length));
+            throw new IllegalArgumentException(String.format("Words (%d) and replaces (%d) lengths should be equals", words.length, replaces.length));
         }
 
         for(int i = 0; i < words.length; i++) {
@@ -120,6 +120,14 @@ public class WordReplacer {
         // The string to find (lower case if ignoreCase)
         String workWord;
 
+        if (input == null || word == null || input.length() == 0 || word.length() == 0) {
+            return;
+        }
+
+        if (replace == null) {
+            replace = "";
+        }
+
         // Use the lowercase versions to compare, but replace in the original input
         if (ignoreCase) {
             workInput = new StringBuilder(input.toString().toLowerCase(Locale.ENGLISH));
@@ -161,11 +169,14 @@ public class WordReplacer {
                 }
             }
 
-            // replace this word but in the original input (without changing the case)
-            input.replace(pos, pos + workWord.length(), replace);
+            if (pos >= 0 && pos < input.length()) {
+                // replace this word but in the original input (without changing the case)
+                input.replace(pos, pos + workWord.length(), replace);
+                workInput.replace(pos, pos + workWord.length(), replace);
+            }
 
             // move to the next character after the replace word
             pos = pos + replace.length();
-        } while (pos > -1);
+        } while (pos > -1 && pos < workInput.length());
     }
 }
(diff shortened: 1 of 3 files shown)


src/test/java/com/cloudbees/jenkins/support/util/WordReplacerTest.java
```diff
@@ -153,6 +153,20 @@ public void caseReplacementsTest() {
         Assert.assertEquals("1 1 b,B.c:C abc ABC ignored", inputSB.toString());
     }
 
+    @Test
+    public void indexOutOfBoundsExceptionTest() {
+        String input = "input one input";
+        String[] words = new String[]{   "input",   "one"};
+        String[] replaces = new String[]{"", ""};
```
varyvol: How is this situation reached? I mean when would we want to replace by an empty string?

src/test/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilterTest.java
```diff
@@ -45,21 +45,23 @@
     @Test
     public void anonymizeSlavesAndLabels() throws Exception {
         SensitiveContentFilter filter = SensitiveContentFilter.get();
-        jenkins.createSlave("foo", "bar", null);
-        jenkins.createSlave("jar", "war", null);
+        // using foo, bar, jar and war could raise flaky test failures. It happened to me when
+        // bar was changed by label_barrier :-O So we use stranger words to avoid this test to be flaky
+        jenkins.createSlave("foostrange", "barstrange", null);
```
varyvol: What does flaky mean here? That the string was not replaced properly?

src/main/java/com/cloudbees/jenkins/support/util/WordReplacer.java
```diff
@@ -161,11 +169,14 @@ private static void replaceWord(StringBuilder input, String word, String replace
                 }
             }
 
-            // replace this word but in the original input (without changing the case)
-            input.replace(pos, pos + workWord.length(), replace);
+            if (pos >= 0 && pos < input.length()) {
```
varyvol: Probably out of scope of this PR, but I'm starting to think it would be interesting to have some utility functions to improve the readability here.

src/test/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilterTest.java
MRamonLeon: It was replaced correctly, but the new text also contains the word *bar* (label_barrier). So the test failed because the assert said it shouldn't, that's the word replaced. :laughing: random texts

src/main/java/com/cloudbees/jenkins/support/util/WordReplacer.java
MRamonLeon: It could happen if the replacement word has less length

src/test/java/com/cloudbees/jenkins/support/util/WordReplacerTest.java
MRamonLeon: As the class WordReplacer is a sort of utility class, it should work in any case, as this one. Specifically, in the use case of this plugin this test is a concrete case of a replacement by a word with less length, which is a valid use case.

src/test/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilterTest.java
jeffret-b: Yes, I'm curious as well. Is it just the test that's flaky? Or is the test reflecting flakiness in the behavior?

MRamonLeon: when the test replaces "bar" by "label_aword" the test is successful. But:
when the test replaces "bar" by "label_barrier" the test fails because bar is in the word _again_ :laughing: 

It's a random word, so sometimes it returns the same word or a word with this substring (bar).

jeffret-b: Ah, so yeah, a flaky test, but not an issue in the implementation.

PR #165, opened 2019-02-19
JENKINS-56190 - Only include the proc files that has to be included.

duemir:
"Master JVM porcess system metrics" and "Agent system configuration"
components were only split in label. So all the files were included
twice. The node list to extract proc info from is abstracted. The
components are update to provide only the nodes from which data must
be collected.

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/ProcFilesRetriever.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/ProcFilesRetriever.java
@@ -14,6 +14,7 @@ import jenkins.model.Jenkins;
 
 import java.io.IOException;
 import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.WeakHashMap;
@@ -45,6 +46,8 @@ public abstract class ProcFilesRetriever extends Component {
      */
     abstract public Map<String, String> getFilesToRetrieve();
 
+    abstract protected List<Node> getNodes();
+
     @NonNull
     @Override
     public Set<Permission> getRequiredPermissions() {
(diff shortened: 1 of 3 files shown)


src/main/java/com/cloudbees/jenkins/support/impl/ProcFilesRetriever.java
```diff
@@ -45,6 +46,8 @@ protected static String getNodeName(Node node) {
      */
     abstract public Map<String, String> getFilesToRetrieve();
 
+    abstract protected List<Node> getNodes();
```
MRamonLeon: My only concern is that now all existing implementors of ProcFilesRetriever will fail as they don't implement `getNodes`, but I suppose it's not very used out there. A mention on the **Changelog** talking about this loss of compatibility is a must-have, I think. cc: @varyvol

aheritier: makes sense @MRamonLeon (limited risk due to the number of existing extensions)

duemir: I missed the compatibility concern completely because I was just trying to fix asap and continue with other work :(
Do you think there is a way to make it backward compatible? I can probably make it concrete and return a list including the master and all the nodes.

varyvol: Good point @MRamonLeon I totally missed it.

@duemir probably you can make the method in the ProcFilesRetriever class not be abstract but return both the Jenkins instance and the nodes (previous behaviour) and then you override it in the subclasses?

duemir: Done. Please let me know what you think.

PR #166, opened 2019-02-20
[JENKINS-50297] WinSW logs not being cleaned up for cloud-provisioned (ephemeral) agents

dariver:
https://issues.jenkins-ci.org/browse/JENKINS-50297

The issue was not only impacting over `winsw` agents logs but in every agent log. 

`SmartLogFetcher` is not removing the cache entries created by agents that are no longer attached to Jenkins. Sometimes these directories can take a big amount of disk space may cause a performance issue in the instance.

In this approach, `SmartLogCleaner` provides a way to clean up the directories which belong to nodes that are not attached anymore to Jenkins.

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/SmartLogCleaner.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/SmartLogCleaner.java
@@ -18,6 +18,7 @@ import com.cloudbees.jenkins.support.SupportPlugin;
  * <p>
  * Iterate over the cache entries stored by the {@link SmartLogFetcher} and remove those which 
  * belong to agents that are no longer attached.
+ * </p>
  */
 class SmartLogCleaner {
 
--- a/src/main/java/com/cloudbees/jenkins/support/impl/SmartLogCleaner.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/SmartLogCleaner.java
@@ -1,10 +1,8 @@
 package com.cloudbees.jenkins.support.impl;
 
 import java.io.File;
-import java.io.FilenameFilter;
+import java.io.FileFilter;
 import java.io.IOException;
-import java.util.HashSet;
-import java.util.List;
 import java.util.Set;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
@@ -12,12 +10,9 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import org.apache.commons.io.FileUtils;
-import org.apache.commons.lang.StringUtils;
 
 import com.cloudbees.jenkins.support.SupportPlugin;
 
-import hudson.Util;
-import hudson.model.Node;
 
 /**
  * <p>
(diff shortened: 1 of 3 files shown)


src/main/java/com/cloudbees/jenkins/support/impl/SmartLogCleaner.java
```diff
@@ -0,0 +1,94 @@
...
+     * @param nodes
+     *      Used to generate the cache keys to match within the target directory.
+     */
+    SmartLogCleaner(final String id, final List<Node> nodes) {
+        this.rootCacheDir = new File(SupportPlugin.getRootDirectory(), id);
+        this.cacheKeys = getActiveCacheKeys(nodes);
+    }
+
+    void execute() {
+        File[] cacheKeyDirs = rootCacheDir.listFiles(new FilenameFilter() {
```
varyvol: I think you can use a FileFilter and the code will be simpler.

src/main/java/com/cloudbees/jenkins/support/impl/SmartLogCleaner.java
```diff
@@ -0,0 +1,94 @@
...
+        if (cacheKeyDirs == null || cacheKeyDirs.length == 0) {
+            LOGGER.log(Level.FINE, "cacheKeys directory '{0]' is empty, nothing to clean up", rootCacheDir);
+        } else {
+            ExecutorService executor = Executors.newSingleThreadExecutor();
+            try {
+                executor.execute(new Runnable() {
+                    @Override
+                    public void run() {
+                        for (File dir: cacheKeyDirs) {
+                            if (dir.isDirectory() && dir.exists() && cacheKeys.contains(dir.getName())) {
```
varyvol: You don't need to check here if it's a directory given the previous filtering. I think the same applies for exists().

src/main/java/com/cloudbees/jenkins/support/impl/SmartLogCleaner.java
```diff
@@ -0,0 +1,94 @@
...
+                });
+            } catch (RuntimeException e) {
+                LOGGER.log(Level.WARNING, "The clean up task has ended with errors", e);
+            } finally {
+                executor.shutdown();
+            }
+        }
+    }
+
+    private Set<String> getActiveCacheKeys(final List<Node> nodes) {
```
varyvol: How heavy is this operation? Does it take much time? Because it's done twice (one for each cleaner and maybe that could be improved.

dariver: It depends on how many agents are attached in the Jenkins instance. I guess it's better moving the operation to `SlaveLogs`As a result of this, the test has also been changed.

dariver:
@varyvol I've addressed your comments

src/main/java/com/cloudbees/jenkins/support/impl/SmartLogCleaner.java
```diff
@@ -0,0 +1,94 @@
...
+import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang.StringUtils;
+
+import com.cloudbees.jenkins.support.SupportPlugin;
+
+import hudson.Util;
+import hudson.model.Node;
+
+/**
+ * <p>
```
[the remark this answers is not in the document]
dariver: done

PR #167, opened 2019-04-04
[JENKINS-33628] Separate ItemsContent from AboutJenkins

fbelzunc:
This PR intends to fix [JENKINS-33628](https://issues.jenkins-ci.org/browse/JENKINS-33628).

In an enterprise environment with thousands of jobs/builds, there is a big performance degradation produced by the computation of counting jobs/builds. When Jenkins is inside a Load Balancer, usually the HTTP session finished before the computation is done and this finish with the support-bundle being corrupted.

Support bundle is critical for Jenkins support organizations.

CC @escoem @aheritier

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/ItemsContent.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/ItemsContent.java
@@ -1,0 +1,198 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2019, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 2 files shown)


varyvol:
@fbelzunc if I am understanding this correctly, you want to move out the items data recollection so that it can be disabled on large instances and thus avoid the performance issues?

src/main/java/com/cloudbees/jenkins/support/impl/ItemsContent.java
```diff
@@ -0,0 +1,202 @@
...
+                    out.println("======================");
+                    out.println();
+                    out.println("  * Number of jobs: " + jobTotal.n());
+                    out.println("  * Number of builds per job: " + jobTotal);
+                }
+            });
+        }
+    }
+
+    private static class Stats {
```
varyvol: I'm not a big fan of duplicating this class. Do you think it makes sense to move it to an external class to be then used anywhere necessary?

fbelzunc: Yes, I will submit the changes

fbelzunc: Sorry, the class is not used in `AboutJenkins` anymore, so I just removed it from there. No need to create a new class only for this.

fbelzunc:
@varyvol Would you mind to re-review, please? I think it should be ready to merge now :-)

fbelzunc:
@varyvol Answering also your previous comment, you are right, the idea is to separate this from the AboutJenkins to avoid the performance issue, which usually produced also corrupted support-bundle when Jenkins is behind a load balancer - as the load balancer timeout is usually reached out.

PR #169, opened 2019-05-27
JENKINS-57708 (CE-2748) - Check if X-Forwarded-For is present

Evildethow:


Evildethow:
> It doesn't look right. If I understand correctly the `X-Forwarded` should never be in the response to the client from the reverse proxy. They are added to the Request headers from reverse proxy to Jenkins. You can get a current request from Stapler if the bundle is generated from the UI but there needs to be a fallback for the bundle generated in the background.
> 
> For reference - https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/jenkins/model/Jenkins.java#L2313-L2352

@duemir 

The component now looks up the request header using the current request. 

I'm not sure how to do this for bundles that aren't generated from the UI. Open to suggestions.

aheritier:
I don't think we can solve the problem for bundles generated in the background
Like for the browser we won't have the info : https://github.com/jenkinsci/support-core-plugin/blob/master/src/main/java/com/cloudbees/jenkins/support/impl/AboutBrowser.java

PR #174, opened 2019-06-17
[JENKINS-57990] Add filtering per content and avoid filtering some contents

MRamonLeon:
**See [[JENKINS-57990] ](https://issues.jenkins-ci.org/browse/JENKINS-57990)**

Now the components can filter the contents. With this approach, the structure of the files is not affected, like markdowns, getting more readability.

Some contents are not filtered because they are not expected to have private information. For example Garbage Collector logs.

Added a full test that checks a lot of components and its expected behavior regarding the filtering.

@reviewbybees @varyvol @fcojfernandez @oleg-nenashev @jvz


aheritier:
@Dohbedoh @Evildethow @PierreBtz could you review this one please ?

aheritier:
@MRamonLeon instead of adding a lot of `filter == null ? ..` in the code couldn't we just have a transparent filter which is doing nothing by default ? Thus we always call `filter.filter()` and we remove all conditions. Not sure in term of performance if it could be really different to have a condition vs `String filter.filter(String value){return value}`

src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
```diff
@@ -477,14 +471,14 @@ private static String humanReadableSize(long size) {
         return format.format(number) + " " + measure + " (" + size + ")";
     }
 
-    private static class AboutContent extends PrintedContent {
+    private static class AboutContent extends PrefilteredPrintedContent {
```
MRamonLeon: We just filter the url of the Jenkins to avoid IPs

src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
```diff
@@ -707,29 +701,29 @@ private String getLabelString(Node n) {
             out.println();
             out.println("  * master (Jenkins)");
             out.println("      - Description:    _" +
-                    Markdown.escapeUnderscore(Util.fixNull(jenkins.getNodeDescription())) + "_");
+                    Markdown.escapeUnderscore(Util.fixNull(filter != null ? filter.filter(jenkins.getNodeDescription()) : jenkins.getNodeDescription())) + "_");
```
MRamonLeon: We filter the pieces of information that may include some private information and avoid filtering the structure of the markdown

src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
```diff
@@ -846,6 +840,12 @@ private String getLabelString(Node n) {
                 }
             }
         }
+
+        @Override
+        public boolean shouldBeFiltered() {
+            // The information of this content is not sensible, so it doesn't need to be filtered.
+            return false;
```
MRamonLeon: It's for the checksums.md, it only has the checksums of all plugins and war

src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
```diff
@@ -866,6 +866,12 @@ private String getLabelString(Node n) {
                         "Could not compute checksums on slave " + node.getNodeName(), e);
             }
         }
+
+        @Override
+        public boolean shouldBeFiltered() {
+            // The information of this content is not sensible, so it doesn't need to be filtered.
+            return false;
```
MRamonLeon: It's for the checksums of the nodes

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -287,13 +287,19 @@ public static void writeBundle(OutputStream outputStream, final List<Component>
         try {
             try (BulkChange change = new BulkChange(ContentMappings.get());
                  ZipArchiveOutputStream binaryOut = new ZipArchiveOutputStream(new BufferedOutputStream(outputStream, 16384))) {
+                // Get the filter to be used
                 Optional<ContentFilter> maybeFilter = getContentFilter();
 
+                // Recalculate the mappings and stop words and save it to disk
+                if(maybeFilter.isPresent()) {
```
varyvol: Why is this necessary now?

src/main/java/com/cloudbees/jenkins/support/impl/AdministrativeMonitors.java
```diff
@@ -98,6 +98,12 @@
                         }
                     }
                 }
+
+                @Override
+                public boolean shouldBeFiltered() {
+                    // The information of this content is not sensible, so it doesn't need to be filtered.
+                    return false;
```
MRamonLeon: It only shows the active monitors.

src/main/java/com/cloudbees/jenkins/support/api/BaseCommandOutputContent.java
```diff
@@ -0,0 +1,91 @@
...
+    // an object of this class. As no external use is expected, it's been moved here.
+    private static class CommandLauncher extends MasterToSlaveCallable<String, RuntimeException> {
+        final String[] command;
+
+        private CommandLauncher(String... command) {
+            this.command = command;
+        }
+
+        public String call() {
+            StringWriter bos = new StringWriter();
```
varyvol: Closing?

src/main/java/com/cloudbees/jenkins/support/impl/AdvancedProcFilesRetriever.java
```diff
@@ -0,0 +1,122 @@
+package com.cloudbees.jenkins.support.impl;
```
MRamonLeon: A new class for getting info from proc files allowing you to select which ones should be filtered and which ones keep unfiltered. The ```ProcFilesRetriever``` uses a ```FilePathContent``` which is always filtered at the end.

src/main/java/com/cloudbees/jenkins/support/api/UnFilteredCommandOutputContent.java
```diff
@@ -0,0 +1,33 @@
+package com.cloudbees.jenkins.support.api;
+
+import hudson.model.Node;
+
+import java.util.WeakHashMap;
+
+/**
+ * Content of a command output. You can only instantiate this content with
+ * a builder method which launch the command immediately. It doesn't pre filter the result.
```
varyvol:
```suggestion
 * a builder method which launches the command immediately. It doesn't pre filter the result.
```

src/main/java/com/cloudbees/jenkins/support/impl/AdvancedProcFilesRetriever.java
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        for (ProcFile procDescriptor : getProcFilesToRetrieve()) {
+            container.add(new FilePathContent("nodes/{0}/proc/{1}", new String[]{name, procDescriptor.getName()},
+                    new FilePath(c.getChannel(), procDescriptor.getFile())) {
+
+                        @Override
+                        public boolean shouldBeFiltered() {
+                            // Whether this specific file should be filtered or not
+                            return procDescriptor.isFiltered();
```
MRamonLeon: here you can decide whether filtering or not

src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
```diff
@@ -500,7 +494,7 @@ private static String humanReadableSize(long size) {
                 out.println("  * Mode:    WAR");
             }
             final JenkinsLocationConfiguration jlc = JenkinsLocationConfiguration.get();
-            out.println("  * Url:     " + (jlc != null ? jlc.getUrl() : "No JenkinsLocationConfiguration available"));
+            out.println("  * Url:     " + (jlc != null ? (filter == null ? jlc.getUrl() : filter.filter(jlc.getUrl())) : "No JenkinsLocationConfiguration available"));
```
varyvol: This would be the other way actually?

src/main/java/com/cloudbees/jenkins/support/impl/AdvancedProcFilesRetriever.java
```diff
@@ -0,0 +1,122 @@
...
+public abstract class AdvancedProcFilesRetriever extends ProcFilesRetriever {
+    private static final Logger LOGGER = Logger.getLogger(AdvancedProcFilesRetriever.class.getName());
+
+    /**
+     * If you want to use this method, it's best to use {@link ProcFilesRetriever}.
+     */
+    @Override
+    @Restricted(DoNotUse.class)
+    public Map<String, String> getFilesToRetrieve() {
+        LOGGER.warning("No proc files will be included to the bundle. You should use the class ProcFilesRetriever or the method AdvancedProdFilesRetriever#getProcFilesToRetrieve");
```
MRamonLeon: You shouldn't use this method with this class, this class is intended to be used with the new ```getProcFilesToRetrieve``` where you can define which files are filtered or not filtered.

src/main/java/com/cloudbees/jenkins/support/impl/JVMProcessSystemMetricsContents.java
```diff
@@ -47,22 +48,25 @@ public boolean isSelectedByDefault() {
         }
     }
 
-    static final Map<String, String> UNIX_PROC_CONTENTS;
+    static final Set<ProcFile> UNIX_PROC_CONTENTS;
```
varyvol: I don't think it has compatibility implications but why this change?

src/main/java/com/cloudbees/jenkins/support/impl/AdvancedProcFilesRetriever.java
```diff
@@ -0,0 +1,122 @@
...
+            });
+        }
+
+        afterAddUnixContents(container, node, name);
+    }
+
+    /**
+     * A class to define a file with a name in the bundle and also whether its content should be filtered.
+     */
+    public static class ProcFile {
```
MRamonLeon: A new class to store if the file needs to be filtered. Instead of a Map<String, String>, we have a Set<ProcFile>

src/main/java/com/cloudbees/jenkins/support/impl/BuildQueue.java
```diff
@@ -69,21 +66,20 @@ public String getDisplayName() {
 
   @Override
   public void addContents(@NonNull Container container) {
-    container.add(new Content("buildqueue.md") {
+    container.add(new PrefilteredPrintedContent("buildqueue.md") {
```
MRamonLeon: Only the private information is filtered to avoid destroying the markdown structure

src/main/java/com/cloudbees/jenkins/support/impl/FileDescriptorLimit.java
```diff
@@ -98,6 +98,12 @@ public void writeTo(OutputStream os) throws IOException {
                                 out.flush();
                             }
                         }
+
+                        @Override
+                        public boolean shouldBeFiltered() {
+                            // The information of this content is not sensible, so it doesn't need to be filtered.
+                            return false;
```
MRamonLeon: No private info here

src/main/java/com/cloudbees/jenkins/support/impl/GCLogs.java
```diff
@@ -83,7 +83,7 @@ public void addContents(@NonNull Container result) {
                         "but file '" + gcLogFileLocation + "' not found");
                 return;
             }
-            result.add(new FileContent(GCLOGS_BUNDLE_ROOT + "gc.log", file));
+            result.add(new UnFilteredFileContent(GCLOGS_BUNDLE_ROOT + "gc.log", file));
```
MRamonLeon: Without filtering, not private info here, avoid destroying the file and make it parseable by tools.

src/main/java/com/cloudbees/jenkins/support/impl/HeapUsageHistogram.java
```diff
@@ -65,6 +65,12 @@ public void addContents(@NonNull Container result) {
                 public void writeTo(OutputStream os) throws IOException {
                     os.write(getLiveHistogram().getBytes("UTF-8"));
                 }
+
+                @Override
+                public boolean shouldBeFiltered() {
+                    // The information of this content is not sensible, so it doesn't need to be filtered.
+                    return false;
```
MRamonLeon: No private info

src/main/java/com/cloudbees/jenkins/support/impl/JVMProcessSystemMetricsContents.java
```diff
@@ -47,22 +48,25 @@ public boolean isSelectedByDefault() {
...
-        contents.put("/proc/self/limits", "self/limits.txt");
-        contents.put("/proc/self/mountstats", "self/mountstats.txt");
-        UNIX_PROC_CONTENTS = Collections.unmodifiableMap(contents);
+        Set<ProcFile> contents = new HashSet<>();
+        // These files don't need filtering
+        contents.add(ProcFile.of("/proc/meminfo", "meminfo.txt", false));
+        contents.add(ProcFile.of("/proc/self/status", "self/status.txt", false));
+        contents.add(ProcFile.of("/proc/self/cmdline", "self/cmdline", false));
+        // This one should be filtered
+        contents.add(ProcFile.of("/proc/self/environ", "self/environ", true));
```
MRamonLeon: This one is filtered

src/main/java/com/cloudbees/jenkins/support/impl/RootCAs.java
```diff
@@ -89,35 +85,30 @@ private void addContents(@NonNull Container container, final @NonNull Node node)
...
+
     String name;
+    String[] params;
     if (node instanceof Jenkins) {
-      name = "master";
+      name = "nodes/master/RootCA.txt";
+      params = new String[0];
     } else {
-      name = "slave/" + node.getNodeName();
+      name = "nodes/slave/{0}/RootCA.txt";
```
MRamonLeon: The name of the file needs to be filtered

src/main/java/com/cloudbees/jenkins/support/impl/ThreadDumps.java
```diff
@@ -279,21 +283,21 @@ public static void printThreadInfo(PrintWriter writer, ThreadInfo t, ThreadMXBea
             cpuPercentage = 0;
         }
         writer.printf("\"%s\" id=%d (0x%x) state=%s cpu=%d%%",
-                t.getThreadName(),
+                filter == null ? t.getThreadName() : filter.filter(t.getThreadName()),
```
MRamonLeon: The name of the thread could have a job name or so

src/main/java/com/cloudbees/jenkins/support/impl/ThreadDumps.java
```diff
@@ -279,21 +283,21 @@ public static void printThreadInfo(PrintWriter writer, ThreadInfo t, ThreadMXBea
...
                 t.getThreadId(),
                 t.getThreadId(),
                 t.getThreadState(),
                 cpuPercentage);
         final LockInfo lock = t.getLockInfo();
         if (lock != null && t.getThreadState() != Thread.State.BLOCKED) {
-            writer.printf("\n    - waiting on <0x%08x> (a %s)",
+            writer.printf("%n    - waiting on <0x%08x> (a %s)",
```
MRamonLeon: make findbugs happy

varyvol:
Regarding @aheritier 's comment I think that is happening because of the new API? Why cannot we make the filter to be NonNull by design? Also Findbug is reporting an error.

src/test/java/com/cloudbees/jenkins/support/CheckFilterTest.java
```diff
@@ -0,0 +1,412 @@
+package com.cloudbees.jenkins.support;
```
MRamonLeon: A big test to check a lot of components doing a bundle with the components to check and checking the content of the files inside.

src/test/java/com/cloudbees/jenkins/support/CheckFilterTest.java
```diff
@@ -0,0 +1,412 @@
...
+        QueueTaskFuture<FreeStyleBuild> build = createObjectsWithNames();
+
+        // Reload the mappings, after the objects were created
+        ContentFilters.get().setEnabled(true);
+        ContentFilter.ALL.reload();
+        // Generate the filtered words
+        checker.generateFilteredWords();
+
+        // Check the components are filtered correctly
+        assertComponent(Arrays.asList(  AboutJenkins.class,
```
MRamonLeon: Components to check

src/test/java/com/cloudbees/jenkins/support/CheckFilterTest.java
```diff
@@ -0,0 +1,412 @@
...
+
+    @Rule
+    public JenkinsRule jenkins = new JenkinsRule();
+
+    @Test
+    public void checkFilterTest() throws Exception {
+        // Create the files to check
+        FileChecker checker = new FileChecker(jenkins.jenkins);
+        // Create the objects needed for some contents to be included
+        QueueTaskFuture<FreeStyleBuild> build = createObjectsWithNames();
```
MRamonLeon: Create objects with some interesting names to have the mappings created

src/test/java/com/cloudbees/jenkins/support/CheckFilterTest.java
```diff
@@ -0,0 +1,412 @@
...
+    @Rule
+    public TemporaryFolder temp = new TemporaryFolder();
+
+    @Rule
+    public JenkinsRule jenkins = new JenkinsRule();
+
+    @Test
+    public void checkFilterTest() throws Exception {
+        // Create the files to check
+        FileChecker checker = new FileChecker(jenkins.jenkins);
```
MRamonLeon: The object doing the checks

src/test/java/com/cloudbees/jenkins/support/CheckFilterTest.java
```diff
@@ -0,0 +1,412 @@
...
+        // Create a job to have something pending in the queue for the BuilQueue component
+        FreeStyleProject project = jenkins.createFreeStyleProject(JOB_NAME);
+        project.setAssignedLabel(new LabelAtom("foo")); //it's mandatory
+
+        //project.getBuildersList().add(new SimpleBuilder());
+        QueueTaskFuture<FreeStyleBuild> buildQueued = project.scheduleBuild2(0);
+        return buildQueued;
+    }
+
+    private void assertComponent(List<Class<? extends Component>> componentClasses, FileChecker checker) throws IOException {
```
MRamonLeon: Create a bundle with the components specified and check with the ```checker``` object the contents of the files in the bundle

src/test/java/com/cloudbees/jenkins/support/CheckFilterTest.java
```diff
@@ -0,0 +1,412 @@
...
+            try(ZipFile zip = new ZipFile(fileZip)) {
+                Enumeration<ZipArchiveEntry> zipFileEntries = zip.getEntries();
+
+                while (zipFileEntries.hasMoreElements()) {
+                    ZipArchiveEntry entry = zipFileEntries.nextElement();
+                    if (entry.isDirectory()) {
+                        break;
+                    }
+                    if (entry.getSize() > 2048 * 1024) {
+                        LOGGER.log(Level.WARNING, "Cannot check file '%s' because its content is bigger than 2Mb", entry.getName());
```
MRamonLeon: Avoid checking big files, just in case it becomes unfeasible in the CI

src/test/java/com/cloudbees/jenkins/support/CheckFilterTest.java
```diff
@@ -0,0 +1,412 @@
...
+                        break;
+                    }
+
+                    String content = getContentFromEntry(zip, entry);
+                    checker.check(entry.getName(), content);
```
MRamonLeon: Check if the content of the file in the bundle is expected

src/test/java/com/cloudbees/jenkins/support/CheckFilterTest.java
```diff
@@ -0,0 +1,412 @@
...
+        return content;
+    }
+
+    private static class FileChecker {
+        private Set<FileToCheck> fileSet = new HashSet<>();
+        private Set<String> words = new HashSet<>();
+
+
+        private FileChecker(Jenkins jenkins) throws UnknownHostException {
+            fileSet.add(of("manifest.md", "about", false));
```
MRamonLeon: Here we're creating the checkings to do: the file manifest.md must have the word about not filtered (false param). For every file we set what is expected to be found and whether it should be filtered or not.

src/test/java/com/cloudbees/jenkins/support/CheckFilterTest.java
```diff
@@ -0,0 +1,412 @@
...
+         */
+        private void check(String file, String content) {
+            final int MAX_CONTENT_LENGTH = Integer.MAX_VALUE;
+            for (FileToCheck value : fileSet) {
+                // If there is a filePattern to check that matches this entry of the bundle, we check
+                if(value.match(file)) {
+                    if (content == null) {
+                        fail(String.format("Error checking the file %s because its content was null", file));
+                    } else {
+                        Assert.assertTrue(String.format("The file '%s' should have the word '%s'. File content:\n\n----------\n%s\n%s----------\n\n", file, value.wordFiltered, content.substring(0, Math.min(MAX_CONTENT_LENGTH, content.length())), content.length() > MAX_CONTENT_LENGTH ? "...\n(content cut off)\n" : ""), content.toLowerCase(Locale.ENGLISH).contains(value.wordFiltered.toLowerCase(Locale.ENGLISH)));
```
MRamonLeon: It is useful for diagnosing bug in future changes but should we avoid printing the content of the file for security reasons? Maybe we are showing information of the CI. Although this information is also accessible by anyone doing a PR anywhere.

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
MRamonLeon: Before we were reloading and saving if the filter was enabled, now it's the same as the filter is not empty.

src/main/java/com/cloudbees/jenkins/support/api/BaseCommandOutputContent.java
MRamonLeon: It has no effect, it's a string buffer, not a real stream. Anyway, the code was there, I just moved it to a base class.

src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
MRamonLeon: :thinking: What do you mean? If the filter is null we return the url without filtering. If the filter is not null, we filter the url. The code seems fine to me. Do you mean do ```filter != null ? filter.filter(...```?

src/main/java/com/cloudbees/jenkins/support/impl/JVMProcessSystemMetricsContents.java
MRamonLeon: Because now you can decide whether to filter or not for every file. The new class ```ProcFile``` let you do that. Instead of just add the file and filter it. But I think this change is a mistake, because I had to have created a new ```AdvancedJVMProcessSystemMetricsContents``` and leave this class as it was. I've searched for **UNIX_PROC_CONTENTS** along GitHub and there are no occurrences of this outside of the plugin.

MRamonLeon:
@aheritier @varyvol I was pondering creating a new IdentityFilter to avoid all this code, but I didn't do that because of performance issues. 

If I create the new filter, this filter is going to be added to the list of _ALL_ filters because it implements _ContentFilter_, and therefore it's going to be used in every filtering. Including every line of the log files. So I decided not to do that even though the code is uglier, but at least, faster.

MRamonLeon:
@varyvol I've introduced this bug in my latest commit, just 1 line with 1 bug. :man_facepalming: One cannot get relaxed or findbugs will come with the hammer

src/main/java/com/cloudbees/jenkins/support/impl/JVMProcessSystemMetricsContents.java
varyvol: So do you pretend to create the new class or leave it as it is?

src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
varyvol: Yes, my mistake sorry. Anyway could jlc.getUrl() be null and, in that case, does the filter handle that correctly?

varyvol:
@MRamonLeon what I mean is: there are writeTo() methods that receive a filter. Why do they allow the filter to be null? I think it was there before but maybe it's worth to have a look and avoid that situation.

If not, I think a utility function in charge of checking the nullability of the filter and using it in a appropriate way would improve the code.

src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
MRamonLeon: I have changed this: https://github.com/jenkinsci/support-core-plugin/pull/174/commits/753c58cb285ada8ccd4efa4743345fcdb8056533#diff-f2c7ab60306f05af2fab348ce75750cfR498

src/main/java/com/cloudbees/jenkins/support/impl/JVMProcessSystemMetricsContents.java
MRamonLeon: My thought is because it's not used anywhere and because it has the feature of the previous class, we can leave it so

MRamonLeon:
@aheritier @varyvol I've added a utility method to make the code cleaner.

MRamonLeon:
Why do they allow the filter to be null? 
@varyvol When the anonymization is disabled the filter is null.

varyvol:
@aheritier that design "flaw" was already there so I am merging this and maybe that can be fixed in the future.

PR #176, opened 2019-07-01
Fixing some issues after JENKINS-57990

MRamonLeon:
* Fix the log contents not being filtered
* Filter java arguments
* Avoid filtering twice the IPs
* Avoid escaping the _ for the labels in the nodes.md file

@aheritier @PierreBtz @Evildethow @varyvol @fcojfernandez @alecharp @oleg-nenashev @batmat


MRamonLeon:
@varyvol please wait before merging because I'm with another issue that may come here.

MRamonLeon:
@PierreBtz The escape of the _ for labels just in node was already there, since 4ae4115305e645fc323f2a3a4e24976202966ff9

I've reverted this change.

MRamonLeon:
We have to rethink how we are writing the markdowns because there are a lot of special characters not being escaped. Like underscores in the file paths and so on. But it's not a change to do in this PR.

MRamonLeon:
Created https://issues.jenkins-ci.org/browse/JENKINS-58341 to handle markdown formatting

varyvol:
@MRamonLeon can you please remove the Markdown calls then?

MRamonLeon:
> @MRamonLeon can you please remove the Markdown calls then?

They are new ones, going with the filtering, to do things right because the data go to a markdown file.

PR #177, opened 2019-07-05
Recommended configuration for building the plugin

MRamonLeon:
Change the Jenkinsfile to use the recommended configuration.


MRamonLeon:
It fails because the Jenkins version should be 2.164. I think we don't want to do that.

PR #179, opened 2019-07-11
[JENKINS-57602] Check null strings to avoid a message in the log

MRamonLeon:
See [JENKINS-57602](https://issues.jenkins-ci.org/browse/JENKINS-57602)

It doesn't deserve more effort, it's usual to have problems while gathering the data.

Avoid an NPE message in the logs.
And a not used import.

Changelog:
* Avoid an extra message in the log

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/util/Markdown.java
+++ b/src/main/java/com/cloudbees/jenkins/support/util/Markdown.java
@@ -1,6 +1,6 @@
 package com.cloudbees.jenkins.support.util;
 
-import hudson.model.Describable;
+import javax.annotation.CheckForNull;
 
 public final class Markdown {
     public static final String NONE_STRING = "(none)";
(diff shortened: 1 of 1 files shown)


MRamonLeon:
Relaunching CI

MRamonLeon:
The problem comes because the nodes configured by JCasC have their _Remote FS root_ as `null`, whereas the nodes configured via the UI have this field as `""` (empty string):
image

JCasC piece of code:
``` 
nodes:
  - permanent:
      labelString: "jnlp"
      launcher:
        jnlp:
          workDirSettings:
            disabled: false
            failIfWorkDirIsMissing: false
            internalDir: "remoting"
      name: "Java Web Start Agent"
      numExecutors: 1
      retentionStrategy: "always"
```

I think maybe JCasC should ensure it's empty, instead of `null`. I even don't know why you can save the node with the field empty and when you configure again the node, you can see the message that it's mandatory:
image

Anyway, I think the Markdown fix is convenient to make the support-core-plugin more robust.

PR #180, opened 2019-07-17
Remove agent command statistics when agent is removed

olivergondza:
https://issues.jenkins-ci.org/browse/JENKINS-58528

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/SlaveCommandStatistics.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/SlaveCommandStatistics.java
@@ -27,12 +27,14 @@ package com.cloudbees.jenkins.support.impl;
 import com.cloudbees.jenkins.support.api.Component;
 import com.cloudbees.jenkins.support.api.Container;
 import com.cloudbees.jenkins.support.api.PrintedContent;
+import com.google.common.annotations.VisibleForTesting;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.Extension;
 import hudson.ExtensionList;
 import hudson.FilePath;
 import hudson.Util;
 import hudson.model.Computer;
+import hudson.model.Node;
 import hudson.model.TaskListener;
 import hudson.remoting.Channel;
 import hudson.remoting.Command;
@@ -41,12 +43,19 @@ import hudson.remoting.Response;
 import hudson.security.Permission;
 import hudson.slaves.ComputerListener;
 import jenkins.model.Jenkins;
+import jenkins.model.NodeListener;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
 
+import javax.annotation.Nonnegative;
+import javax.annotation.Nonnull;
+import javax.annotation.concurrent.GuardedBy;
 import java.io.File;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.Map;
 import java.util.Set;
@@ -56,7 +65,14 @@ import java.util.regex.Pattern;
 @Extension
 public final class SlaveCommandStatistics extends Component {
 
+    /*protected*/ static @Nonnegative int MAX_STATS_SIZE = 1000;
+
+    private final Object statLock = new Object();
+    @GuardedBy("statLock")
     private final Map<String, Statistics> statistics = Collections.synchronizedSortedMap(new TreeMap<>());
+    // Log of statistics for late rotation that no longer has its computer. Oldest entries first.
+    @GuardedBy("statLock")
+    private final LinkedHashMap<String, Statistics> statLog = new LinkedHashMap<>();
 
     @Override
     public String getDisplayName() {
@@ -70,7 +86,7 @@ public final class SlaveCommandStatistics extends Component {
 
     @Override
     public void addContents(Container container) {
-        statistics.forEach((name, stats) -> container.add(new PrintedContent("nodes/slave/{0}/command-stats.md", name) {
+        getStatistics().forEach((name, stats) -> container.add(new PrintedContent("nodes/slave/{0}/command-stats.md", name) {
             @Override
             protected void printTo(PrintWriter out) throws IOException {
                 stats.print(out);
@@ -82,6 +98,16 @@ public final class SlaveCommandStatistics extends Component {
                 return false;
             }
         }));
+    }
+
+    @VisibleForTesting
+    /*package*/ Map<String, Statistics> getStatistics() {
+        synchronized (statLock) {
+            Map<String, Statistics> out = new HashMap<>(statistics.size() + statLog.size());
+            out.putAll(statistics);
+            out.putAll(statLog);
+            return out;
+        }
     }
 
     private static final class Statistics extends Channel.Listener {
@@ -164,7 +190,6 @@ public final class SlaveCommandStatistics extends Component {
             out.println("# JARs sent");
             jars.forEach(jar -> out.printf("* `%s`: %db%n", jar.getName(), jar.length()));
         }
-
     }
 
     @Extension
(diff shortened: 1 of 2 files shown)


olivergondza:
Thanks for confirmation, will look into that.

olivergondza:
I did not found (again) a decent reusable collection to use as a ring buffer in java. What I have confident is working well is https://github.com/jenkinsci/cloud-stats-plugin/blob/master/src/main/java/org/jenkinsci/plugins/cloudstats/CyclicThreadSafeCollection.java, but I will try to come up with something minimalistic here instead.

olivergondza:
@jglick, please, have another look.

varyvol:
@olivergondza the test seems to be failing on Windows.

olivergondza:
@varyvol, yeah, the windows node took a day off.

```
13:06:17  win2012-ae88d1 was marked offline: Connection was broken: java.util.concurrent.TimeoutException: Ping started at 1563447737472 hasn't completed by 1563447977472
13:06:17  	at hudson.remoting.PingThread.ping(PingThread.java:134)
13:06:17  	at hudson.remoting.PingThread.run(PingThread.java:90)
```

olivergondza:
Alright, tests are green. Please review.

src/test/java/com/cloudbees/jenkins/support/impl/SlaveCommandStatisticsTest.java
```diff
@@ -77,4 +78,44 @@ public Void invoke(File f, VirtualChannel channel) throws IOException, Interrupt
...
+        assertThat(scs.getStatistics().size(), equalTo(4));
+        r.jenkins.removeNode(s0);
+        assertThat(scs.getStatistics().size(), equalTo(4));
+        r.jenkins.removeNode(s1);
+        assertThat(scs.getStatistics().size(), equalTo(4));
+        r.jenkins.removeNode(s2);
+        assertThat(scs.getStatistics().size(), equalTo(4));
+        r.jenkins.removeNode(s3);
+        assertThat(scs.getStatistics().size(), equalTo(3));
+        assertThat("Latest preserved", scs.getStatistics().keySet(), contains(s3.getNodeName(), s2.getNodeName(), s1.getNodeName()));
```
varyvol: Is this checking the order? Or only that they are contained?

src/main/java/com/cloudbees/jenkins/support/impl/SlaveCommandStatistics.java
```diff
@@ -56,7 +65,14 @@
 @Extension
 public final class SlaveCommandStatistics extends Component {
 
+    /*protected*/ static @Nonnegative int MAX_STATS_SIZE = 1000;
```
varyvol: Would it be interesting to allow customizing this value through a system property?

src/test/java/com/cloudbees/jenkins/support/impl/SlaveCommandStatisticsTest.java
olivergondza: yep, `containsInAnyOrder` does the other thing.

src/main/java/com/cloudbees/jenkins/support/impl/SlaveCommandStatistics.java
olivergondza: I guess it can always be added if someone feels the need. For now, clearly, noone had a reason to even consider it since we seem to be the first ones running into this problem.

src/test/java/com/cloudbees/jenkins/support/impl/SlaveCommandStatisticsTest.java
varyvol: Shouldn't this have the opposite order then? The ones that are removed first are the ones that should be in the result first.

olivergondza: Note that current implementation has no time based ordering. Given the only thing we do with the data is we dump them into the bundle where they will be consumed after extraction in different order (alphabetical or FS specific one), I do not see much point in doing that. What is important here is those are the _most recent_ statistics we keep.

varyvol: I'm not worried about the order itself (though I don't understand why it's this one) but by the fact that the wrong ones could be removed. But that case seems to be covered by the fact that s0 is not present.

olivergondza: The chosen order is given by implementation specifics of `LinkedHashMap` where popping from the end would have linear complexity while from the front it has constant: https://github.com/jenkinsci/support-core-plugin/pull/180/files#diff-f2444e1194b4e202ab31a9f3e90e9444R215

I was concerned about the same thing and I believe the tests are covering the fact it is the latest elements that are present after rotation, as you said.

src/main/java/com/cloudbees/jenkins/support/impl/SlaveCommandStatistics.java
```diff
@@ -56,7 +65,14 @@
...
+
+    private final Object statLock = new Object();
```
jglick: Simpler to just synchronize on `this` (e.g., `synchronized` method modified).

src/main/java/com/cloudbees/jenkins/support/impl/SlaveCommandStatistics.java
```diff
@@ -56,7 +65,14 @@
...
+    @GuardedBy("statLock")
     private final Map<String, Statistics> statistics = Collections.synchronizedSortedMap(new TreeMap<>());
```
jglick: Presumably do not need the synchronization wrapper if you are explicitly synchronizing anyway?

src/main/java/com/cloudbees/jenkins/support/impl/SlaveCommandStatistics.java
```diff
@@ -84,6 +100,16 @@ public boolean shouldBeFiltered() {
         }));
     }
 
+    @VisibleForTesting
+    /*package*/ Map<String, Statistics> getStatistics() {
+        synchronized (statLock) {
+            Map<String, Statistics> out = new HashMap<>(statistics.size() + statLog.size());
```
jglick: This discards the sort order. Recommend a `TreeMap` here (in which case `statistics` would be a `HashMap`).

src/main/java/com/cloudbees/jenkins/support/impl/SlaveCommandStatistics.java
```diff
@@ -174,7 +199,25 @@ private void print(PrintWriter out) {
         public void preOnline(Computer c, Channel channel, FilePath root, TaskListener listener) throws IOException, InterruptedException {
             channel.addListener(ExtensionList.lookupSingleton(SlaveCommandStatistics.class).statistics.computeIfAbsent(c.getName(), k -> new Statistics()));
         }
-
     }
 
+    // Rotate Statistics entries between #statistics and #statLog separating the live entries from historical ones so we
+    // can put a capacity constrain on the latter.
```
jglick:
```suggestion
    // can put a capacity constraint on the latter.
```

jglick: IMO it would suffice just to have to simple cap on the number of statistics kept, with some sort of LRU behavior. There is no particular reason that we _must_ keep statistics for a node just because it happens to still be defined. Someone reviewing the bundle is most likely just interested in statistics for recently active nodes.

src/main/java/com/cloudbees/jenkins/support/impl/SlaveCommandStatistics.java
olivergondza: I prefer to have the lock encapsulated so no unrelated code would share the same monitor accidentally - which happens to the intrinsic one. Do you feel strongly against?

src/main/java/com/cloudbees/jenkins/support/impl/SlaveCommandStatistics.java
olivergondza: Yeah, I guess we can get rid of that now.

src/main/java/com/cloudbees/jenkins/support/impl/SlaveCommandStatistics.java
jglick: Not at all, just a thought. If I “felt strongly” I would request changes!

src/main/java/com/cloudbees/jenkins/support/impl/SlaveCommandStatistics.java
olivergondza: Yeah well, this is safer and already implemented so ...

src/main/java/com/cloudbees/jenkins/support/impl/SlaveCommandStatistics.java
olivergondza: Agreed with preserving the sort order. Do not understand the "in which case statistics would be a HashMap" path.

jglick: Or a `LinkedHashMap`, preserving sorted `statistics` followed by chronological `statLog`. Whatever.

olivergondza:
Thanks, Jesse. Review comments addressed, build is green.

olivergondza:
@varyvol, do you think this can be merged and released?

varyvol:
@olivergondza sure! Merging and will try to release it during the day.

varyvol:
@olivergondza  released.

olivergondza:
Thanks alot!

PR #182, opened 2019-08-19
[JENKINS-58165] Handle parameterized GC logs filenames

Dohbedoh:
[JENKINS-58165](https://issues.jenkins-ci.org/browse/JENKINS-58165): Handle parameterized GC logs files name.

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/GCLogs.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/GCLogs.java
@@ -109,7 +109,7 @@ public class GCLogs extends Component {
         File gcLogFile = new File(gcLogFileLocation);
 
         // always add .* in the end because this is where the numbering is going to happen
-        String regex = gcLogFile.getName().replaceAll("%[ptn]", ".*") + ".*";
+        String regex = gcLogFile.getName().replaceAll("%[pt]", ".*") + ".*";
         final Pattern gcLogFilesPattern = Pattern.compile(regex);
 
         File parentDirectory = gcLogFile.getParentFile();
--- a/src/main/java/com/cloudbees/jenkins/support/impl/GCLogs.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/GCLogs.java
@@ -74,7 +74,7 @@ public class GCLogs extends Component {
             return;
         }
 
-        if (isGcLogRotationConfigured()) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/cloudbees/jenkins/support/impl/GCLogs.java
```diff
@@ -147,6 +147,10 @@ private String getGcLogFileLocation() {
         }
         return gcLogSwitch.substring(GCLOGS_JRE_SWITCH.length());
     }
+    
+    public boolean isFileLocationParameterized(String fileLocation) {
+        return Pattern.compile(".*%[tnp].*").matcher(fileLocation).find();
```
varyvol: Is it possible a \n in the file location?

src/main/java/com/cloudbees/jenkins/support/impl/GCLogs.java
```diff
@@ -95,7 +95,7 @@ public boolean isSelectedByDefault() {
     /**
      * Two cases:
      * <ul>
-     * <li>The file name contains <code>%t</code> or <code>%p</code> somewhere in the middle:
+     * <li>The file name contains <code>%t</code>, <code>%p</code> or <code>%n</code> somewhere in the middle:
```
MRamonLeon: I didn't find any reference to **%n**. What does it mean? See: 

- https://docs.oracle.com/en/java/javase/11/tools/java.html#GUID-BE93ABDC-999C-4CB5-A88B-1994AAAC74D5
- https://blog.codefx.org/java/unified-logging-with-the-xlog-option/

src/main/java/com/cloudbees/jenkins/support/impl/GCLogs.java
MRamonLeon: It's %n, not resolved to \n I guess, but I couldn't find it documented. Just %t and %p

Dohbedoh: `%n` was actually a feature request. But you are right it is not actually supported.

aheritier:
@Dohbedoh let's move forward ? cc @Evildethow

src/main/java/com/cloudbees/jenkins/support/impl/GCLogs.java
Dohbedoh: ditto

PR #183, opened 2019-08-19
[JENKINS-56245] Add the Instance ID to AboutJenkins.

Dohbedoh:
[JENKINS-56245](https://issues.jenkins-ci.org/browse/JENKINS-56245): Add the Jenkins Instance ID to the support bundle

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
@@ -502,7 +502,7 @@ public class AboutJenkins extends Component {
             out.println("---------------");
             out.println();
             out.println("  * Version: `" + Markdown.escapeBacktick(Jenkins.VERSION) + "`");
-            out.println("  * Instance ID: `" + Markdown.escapeBacktick(Jenkins.getInstanceOrNull().getLegacyInstanceId()) + "`");
+            out.println("  * Instance ID: `" + Markdown.escapeBacktick(Jenkins.get().getLegacyInstanceId()) + "`");
             File jenkinsWar = Lifecycle.get().getHudsonWar();
             if (jenkinsWar == null) {
                 out.println("  * Mode:    Webapp Directory");
--- a/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
@@ -19,6 +19,7 @@ import com.codahale.metrics.Snapshot;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.SuppressWarnings;
 import hudson.Extension;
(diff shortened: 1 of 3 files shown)


src/test/java/com/cloudbees/jenkins/support/SupportTestUtils.java
```diff
@@ -16,23 +16,48 @@
...
+     * Invoke a component, and return a component content as a String.
+     * Useful to unit test a specific content.
+     */
+    public static String invokeContentToString(final Component component, final String contentName) {
+        final ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        component.addContents(
+                new Container() {
+                    @Override
+                    public void add(@CheckForNull Content content) {
+                        if(!contentName.equals(content.getName())) {
```
varyvol: Why this check?

src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
```diff
@@ -501,6 +501,7 @@ private static String humanReadableSize(long size) {
             out.println("---------------");
             out.println();
             out.println("  * Version: `" + Markdown.escapeBacktick(Jenkins.VERSION) + "`");
+            out.println("  * Instance ID: `" + Markdown.escapeBacktick(Jenkins.get().getLegacyInstanceId()) + "`");
```
varyvol: The ticket speaks about the non-legacy one too?

fcojfernandez: That's my understanding as well. See:

- https://wiki.jenkins.io/display/JENKINS/Instance+Identity
- https://github.com/jenkinsci/jenkins/blob/01d7f443c9e3178da7222e1dff3fc8cf82bc6943/core/src/main/java/jenkins/model/identity/IdentityRootAction.java#L39-L42

aheritier:
@Dohbedoh let's move forward ? cc @Evildethow 
I would have both and at the minimum the legacy one which is used by a known company to identify his customers instances ... @varyvol @MRamonLeon

src/test/java/com/cloudbees/jenkins/support/SupportTestUtils.java
Dohbedoh: I wanted to just grab a specific content rather than a full component. I reverted this.

src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
Dohbedoh: I added the one from the Identity Module.

Dohbedoh:
Some transient issue in the build:

```
[2019-09-23T17:52:11.408Z] java.io.IOException: Tried to load head FlowNodes for execution Owner[Plugins/support-core-plugin/PR-183/9:Plugins/support-core-plugin/PR-183 #9] but FlowNode was not found in storage for head id:FlowNodeId 1:99
```

PR #187, opened 2019-09-04
JENKINS-59197 (OSS-2771) - Should report available processes for master

Evildethow:
https://issues.jenkins-ci.org/browse/JENKINS-59197

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
@@ -375,6 +375,7 @@ public class AboutJenkins extends Component {
                     break;
                 }
             }
+            result.append(min).append(" Available CPUs:   ").append(Runtime.getRuntime().availableProcessors()).append("\n");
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
```diff
@@ -375,6 +375,7 @@ private String getInfo(ContentFilter filter) {
                     break;
                 }
             }
+            result.append(min).append(" Available CPUs:   ").append(Runtime.getRuntime().availableProcessors()).append("\n");
```
aheritier:
```suggestion
            result.append(min).append(" Available CPUs:   ").append(runtime.availableProcessors()).append("\n");
```

You have a runtime instance created few lines before :
https://github.com/jenkinsci/support-core-plugin/blob/dbb58348f625dc61c33542c73aafaa8d8f5b5aa7/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java#L334

PR #188, opened 2019-09-12
JENKINS-59320 (OSS-2774) - Should log all running jobs when bundle is generated

Evildethow:
https://issues.jenkins-ci.org/browse/JENKINS-59320

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/RunningJobs.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/RunningJobs.java
@@ -1,0 +1,61 @@
+package com.cloudbees.jenkins.support.impl;
+
+import com.cloudbees.jenkins.support.api.Component;
+import com.cloudbees.jenkins.support.api.Container;
+import com.cloudbees.jenkins.support.api.Content;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import hudson.Extension;
+import hudson.model.Job;
+import hudson.security.Permission;
+import jenkins.model.Jenkins;
+
+import java.io.BufferedWriter;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.io.OutputStreamWriter;
+import java.io.PrintWriter;
+import java.nio.charset.StandardCharsets;
+import java.util.Collections;
+import java.util.Optional;
+import java.util.Set;
+
+@Extension
+public class RunningJobs extends Component {
+
+    @NonNull
+    @Override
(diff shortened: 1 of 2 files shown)


aheritier:
@jenkinsci/code-reviewers @reviewbybees @jenkinsci/core Could you have a look at this PR to see if there is any risk in term of performance / stability please? ( cc @oleg-nenashev @alecharp @jglick @jtnord )

src/main/java/com/cloudbees/jenkins/support/impl/RunningJobs.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @Override
+    public void addContents(@NonNull Container result) {
+        result.add(
+            new Content("nodes/master/running-jobs.txt") {
+                @Override
+                public void writeTo(OutputStream os) throws IOException {
+                    try (PrintWriter out = new PrintWriter(new BufferedWriter(new OutputStreamWriter(os, StandardCharsets.UTF_8)))) {
+                        Optional.ofNullable(Jenkins.getInstanceOrNull())
+                            .ifPresent(jenkins -> jenkins.getAllItems(Job.class)
```
aheritier: I'm really afraid about the call to `getAllItems` on big instances with thousands of jobs

In the UI, to display the running jobs (thus we exclude the lightweight ones (pipeline jobs & co) they are listing them from the  `computers` /  `executors`

https://github.com/jenkinsci/jenkins/blob/22aa2e6e766074d11249893e3f35e0b99e20d3d0/core/src/main/resources/lib/hudson/executors.jelly#L122

https://javadoc.jenkins.io/jenkins/model/Jenkins.html#getComputers--

varyvol:
Yes, that also worries me.

src/main/java/com/cloudbees/jenkins/support/impl/RunningJobs.java
```diff
@@ -0,0 +1,67 @@
...
+                                .stream()
+                                .filter(Job::isBuilding)
+                                .forEach(job -> {
+                                    ((Job<?, ?>) job)
+                                        .getBuilds()
+                                        .filter(Run::isBuilding)
+                                        .forEach(run -> out.println(String.format(MESSAGE_FORMAT, job.getFullName(), run.getNumber())));
```
MRamonLeon: Shouldn't it be filtered in case the anonymization is on? Even worse in terms of performance.

PR #189, opened 2019-09-13
JENKINS-59320 - Should log all running jobs when bundle is generated

Evildethow:
https://issues.jenkins-ci.org/browse/JENKINS-59320

Replaces https://github.com/jenkinsci/support-core-plugin/pull/188

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -157,7 +157,7 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <artifactId>workflow-durable-task-step</artifactId>
-      <optional>true</optional>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -156,6 +156,11 @@
     <!-- Managed by BOM -->
     <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-durable-task-step</artifactId>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <artifactId>workflow-cps</artifactId>
       <scope>test</scope>
     </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -93,7 +93,7 @@
       <dependency>
         <groupId>io.jenkins.tools.bom</groupId>
         <artifactId>bom</artifactId>
-        <version>2.138.1</version>
+        <version>2.138.2</version>
         <scope>import</scope>
         <type>pom</type>
       </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -169,6 +169,12 @@
       <artifactId>workflow-job</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-support</artifactId>
+      <classifier>tests</classifier>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
 </project>
--- a/pom.xml
+++ b/pom.xml
@@ -71,7 +71,7 @@
   <properties>
     <revision>2.62</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.107</jenkins.version>
+    <jenkins.version>2.138.4</jenkins.version>
     <java.level>8</java.level>
   </properties>
 
@@ -87,6 +87,18 @@
       <url>https://repo.jenkins-ci.org/public/</url>
     </pluginRepository>
   </pluginRepositories>
+
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>io.jenkins.tools.bom</groupId>
+        <artifactId>bom</artifactId>
+        <version>2.138.1</version>
+        <scope>import</scope>
+        <type>pom</type>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
 
   <dependencies>
     <!-- regular dependencies -->
@@ -139,6 +151,17 @@
       <groupId>org.quicktheories</groupId>
       <artifactId>quicktheories</artifactId>
       <version>0.25</version>
+      <scope>test</scope>
+    </dependency>
+    <!-- Managed by BOM -->
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-cps</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-job</artifactId>
       <scope>test</scope>
     </dependency>
   </dependencies>
--- a/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
@@ -510,7 +510,7 @@ public class AboutJenkins extends Component {
             }
             final JenkinsLocationConfiguration jlc = JenkinsLocationConfiguration.get();
 
-            out.println("  * Url:     " + (jlc != null ? ContentFilter.filter(filter, jlc.getUrl()) : "No JenkinsLocationConfiguration available"));
+            out.println("  * Url:     " + (ContentFilter.filter(filter, jlc.getUrl())));
             try {
                 final ServletContext servletContext = Stapler.getCurrent().getServletContext();
                 out.println("  * Servlet container");
--- a/src/main/java/com/cloudbees/jenkins/support/impl/RunningBuilds.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/RunningBuilds.java
@@ -35,7 +35,7 @@ public class RunningBuilds extends Component {
     @Override
     public void addContents(@NonNull Container result) {
         result.add(
-            new PrefilteredPrintedContent("nodes/master/running-builds.txt") {
+            new PrefilteredPrintedContent("running-builds.txt") {
                 @Override
                 protected void printTo(PrintWriter out, ContentFilter filter) {
                     Optional.ofNullable(Jenkins.getInstanceOrNull())
--- a/src/main/java/com/cloudbees/jenkins/support/impl/RunningBuilds.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/RunningBuilds.java
@@ -8,7 +8,6 @@ import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.Extension;
 import hudson.security.Permission;
 import jenkins.model.Jenkins;
-import org.jenkinsci.plugins.workflow.support.steps.ExecutorStepExecution;
 
 import java.io.PrintWriter;
 import java.util.Arrays;
@@ -43,7 +42,7 @@ public class RunningBuilds extends Component {
                             .flatMap(computer -> computer.getAllExecutors().stream())
                             .collect(Collectors.toList())
                             .forEach(executor -> Optional.ofNullable(executor.getCurrentExecutable())
-                                .filter(executable -> !(executable.getParent() instanceof ExecutorStepExecution.PlaceholderTask))
+                                .filter(executable -> executable.getParent() == executable.getParent().getOwnerTask())
                                 .ifPresent(executable -> out.println(ContentFilter.filter(filter, executable.toString())))
                             )
                         );
--- a/src/main/java/com/cloudbees/jenkins/support/impl/RunningBuilds.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/RunningBuilds.java
@@ -37,14 +37,12 @@ public class RunningBuilds extends Component {
             new PrefilteredPrintedContent("running-builds.txt") {
                 @Override
                 protected void printTo(PrintWriter out, ContentFilter filter) {
-                    Optional.ofNullable(Jenkins.getInstanceOrNull())
-                        .ifPresent(jenkins -> Arrays.stream(jenkins.getComputers())
-                            .flatMap(computer -> computer.getAllExecutors().stream())
-                            .collect(Collectors.toList())
-                            .forEach(executor -> Optional.ofNullable(executor.getCurrentExecutable())
-                                .filter(executable -> executable.getParent() == executable.getParent().getOwnerTask())
-                                .ifPresent(executable -> out.println(ContentFilter.filter(filter, executable.toString())))
-                            )
+                    Arrays.stream(Jenkins.get().getComputers())
+                        .flatMap(computer -> computer.getAllExecutors().stream())
+                        .collect(Collectors.toList())
+                        .forEach(executor -> Optional.ofNullable(executor.getCurrentExecutable())
+                            .filter(executable -> executable.getParent() == executable.getParent().getOwnerTask())
+                            .ifPresent(executable -> out.println(ContentFilter.filter(filter, executable.toString())))
                         );
                 }
             });
--- a/src/main/java/com/cloudbees/jenkins/support/impl/RunningBuilds.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/RunningBuilds.java
@@ -18,7 +18,7 @@ import java.util.Set;
 import java.util.stream.Collectors;
 
 @Extension
-public class RunningJobs extends Component {
+public class RunningBuilds extends Component {
 
     @NonNull
     @Override
@@ -29,13 +29,13 @@ public class RunningJobs extends Component {
     @Override
     @NonNull
     public String getDisplayName() {
-        return "Running Jobs";
+        return "Running Builds";
     }
 
     @Override
     public void addContents(@NonNull Container result) {
         result.add(
-            new PrefilteredPrintedContent("nodes/master/running-jobs.txt") {
+            new PrefilteredPrintedContent("nodes/master/running-builds.txt") {
                 @Override
                 protected void printTo(PrintWriter out, ContentFilter filter) {
                     Optional.ofNullable(Jenkins.getInstanceOrNull())
--- a/src/main/java/com/cloudbees/jenkins/support/impl/RunningJobs.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/RunningJobs.java
@@ -39,7 +39,7 @@ public class RunningJobs extends Component {
                 protected void printTo(PrintWriter out) {
                     Optional.ofNullable(Jenkins.getInstanceOrNull())
                         .ifPresent(jenkins -> Arrays.stream(jenkins.getComputers())
-                            .flatMap(computer -> computer.getExecutors().stream())
+                            .flatMap(computer -> computer.getAllExecutors().stream())
                             .filter(Executor::isBusy)
                             .collect(Collectors.toList())
                             .forEach(executor -> Optional.ofNullable(executor.getCurrentExecutable())
--- a/src/main/java/com/cloudbees/jenkins/support/impl/RunningJobs.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/RunningJobs.java
@@ -1,0 +1,58 @@
+package com.cloudbees.jenkins.support.impl;
+
+import com.cloudbees.jenkins.support.api.Component;
+import com.cloudbees.jenkins.support.api.Container;
+import com.cloudbees.jenkins.support.api.Content;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import hudson.Extension;
+import hudson.model.Executor;
+import hudson.security.Permission;
+import jenkins.model.Jenkins;
+
+import java.io.BufferedWriter;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.io.OutputStreamWriter;
+import java.io.PrintWriter;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+@Extension
+public class RunningJobs extends Component {
+
+    @NonNull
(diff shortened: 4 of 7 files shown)


varyvol:
I think the anonymization concerns that were raised in the previous PR still apply.

aheritier:
yes good point @varyvol URLs should be anomymized @Evildethow if the feature is active

src/main/java/com/cloudbees/jenkins/support/impl/RunningJobs.java
```diff
@@ -0,0 +1,58 @@
...
+                @Override
+                public void writeTo(OutputStream os) throws IOException {
+                    try (OutputStreamWriter osw = new OutputStreamWriter(os, StandardCharsets.UTF_8); BufferedWriter bw = new BufferedWriter(osw); PrintWriter out = new PrintWriter(bw)) {
+                        Optional.ofNullable(Jenkins.getInstanceOrNull())
+                            .ifPresent(jenkins -> Arrays.stream(jenkins.getComputers())
+                                .flatMap(computer -> computer.getExecutors().stream())
+                                .filter(Executor::isBusy)
+                                .collect(Collectors.toList())
+                                .forEach(executor -> Optional.ofNullable(executor.getCurrentExecutable())
+                                    .ifPresent(executable -> out.println(executable.getParent().getOwnerTask().getUrl()))));
```
MRamonLeon: Well, it really got filtered without doing anything. It's not a prefiltered content, but it doesn't matter. Maybe for performance reasons, it's good to be a pre-filtered content, but it works, IIRC.

src/main/java/com/cloudbees/jenkins/support/impl/RunningJobs.java
```diff
@@ -0,0 +1,58 @@
...
+    @Override
+    @NonNull
+    public String getDisplayName() {
+        return "Running Jobs";
+    }
+
+    @Override
+    public void addContents(@NonNull Container result) {
+        result.add(
+            new Content("nodes/master/running-jobs.txt") {
```
MRamonLeon: A **PrintedContent** makes it easier.

src/test/java/com/cloudbees/jenkins/support/impl/RunningJobsTest.java
```diff
@@ -0,0 +1,30 @@
...
+        String jobName = "running-job";
+        FreeStyleProject p = j.createFreeStyleProject(jobName);
+        QueueTaskFuture<FreeStyleBuild> build = p.scheduleBuild2(0);
+        build.waitForStart();
+
+        String output = SupportTestUtils.invokeComponentToString(new RunningJobs());
+
+        assertThat(output, containsString(p.getUrl()));
+    }
+}
```
MRamonLeon: nit: lack the new line

src/main/java/com/cloudbees/jenkins/support/impl/RunningJobs.java
```diff
@@ -0,0 +1,58 @@
...
+                @Override
+                public void writeTo(OutputStream os) throws IOException {
+                    try (OutputStreamWriter osw = new OutputStreamWriter(os, StandardCharsets.UTF_8); BufferedWriter bw = new BufferedWriter(osw); PrintWriter out = new PrintWriter(bw)) {
+                        Optional.ofNullable(Jenkins.getInstanceOrNull())
+                            .ifPresent(jenkins -> Arrays.stream(jenkins.getComputers())
+                                .flatMap(computer -> computer.getExecutors().stream())
+                                .filter(Executor::isBusy)
```
jglick: This method is close to useless. Several years ago, @kohsuke rewrote `Executor` to create a fresh one for each task, so basically every executor is busy almost as soon as it is created.

src/main/java/com/cloudbees/jenkins/support/impl/RunningJobs.java
```diff
@@ -0,0 +1,58 @@
...
+    @Override
+    public void addContents(@NonNull Container result) {
+        result.add(
+            new Content("nodes/master/running-jobs.txt") {
+                @Override
+                public void writeTo(OutputStream os) throws IOException {
+                    try (OutputStreamWriter osw = new OutputStreamWriter(os, StandardCharsets.UTF_8); BufferedWriter bw = new BufferedWriter(osw); PrintWriter out = new PrintWriter(bw)) {
+                        Optional.ofNullable(Jenkins.getInstanceOrNull())
+                            .ifPresent(jenkins -> Arrays.stream(jenkins.getComputers())
+                                .flatMap(computer -> computer.getExecutors().stream())
```
jglick: As previously mentioned, this will _not_ include Pipeline builds. You need `getAllExecutors`.

(Then you _may_ want to exclude, or specially display in a nested format, `node` blocks which are owned by the build overall. Unfortunately there is no equivalent of `SubTask.getOwnerTask` for `Executable`, only `getParent`, so from a `PlaceholderTask` you can only infer the owning build via `Task.url` being equal to an `Executable.url`.)

src/main/java/com/cloudbees/jenkins/support/impl/RunningJobs.java
jglick: This does not make much sense. You are printing the job name when you could be printing the specific build. And there is no deduplication, and the format is URL-oriented which is less than readable. So if three builds are running for one branch of a multibranch project, you will see something like

```
job/some-repo/job/master/
job/some-repo/job/master/
job/some-repo/job/master/
```

rather than what you might expect

```
some-repo/master#112
some-repo/master#113
some-repo/master#114
```

Suggest, as I did before:

```suggestion
                                    .ifPresent(executable -> out.println(executable))));
```

src/test/java/com/cloudbees/jenkins/support/impl/RunningJobsTest.java
```diff
@@ -0,0 +1,30 @@
...
+
+public class RunningJobsTest {
+
+    @Rule
+    public JenkinsRule j = new JenkinsRule();
+
+    @Test
+    public void addContents() throws Exception {
+        String jobName = "running-job";
+        FreeStyleProject p = j.createFreeStyleProject(jobName);
```
jglick: Really need to import the BOM, add Pipeline test dependencies, and verify behavior for a Pipeline build running in a couple of `node` blocks in `parallel` to see what this would look like.

src/main/java/com/cloudbees/jenkins/support/impl/RunningJobs.java
jglick: https://github.com/jenkinsci/jenkins/pull/944 FTR

src/main/java/com/cloudbees/jenkins/support/impl/RunningJobs.java
jglick: …though it looks like this will not currently print friendly content for `PlaceholderExecutable`s. Based on my reading of `Executable.toString` Javadoc, this would be a bug in `workflow-durable-task-step-plugin`, easy enough to fix.

Evildethow: @MRamonLeon its not clear to me what I have to do to add anonymization support. Can you provide what steps are required? Also, it might be useful to document how to use this feature.

src/main/java/com/cloudbees/jenkins/support/impl/RunningJobs.java
Evildethow: @jglick I don't understand what you are suggesting that I do here. How do I filter executors if this method cannot be trusted?

src/main/java/com/cloudbees/jenkins/support/impl/RunningJobs.java
Evildethow: @jglick I don't understand what you are suggesting I do here.

src/test/java/com/cloudbees/jenkins/support/impl/RunningJobsTest.java
Evildethow: @jglick I assume you are referring to https://github.com/jenkinsci/bom

Do you have an example of what you are asking me to do? I'd rather not play PR ping pong trying to guess.

src/main/java/com/cloudbees/jenkins/support/impl/RunningJobs.java
Evildethow: Resolved by filtering PlaceholderExecutables

```
.forEach(executor -> Optional.ofNullable(executor.getCurrentExecutable())
                                .filter(executable -> !(executable.getParent() instanceof ExecutorStepExecution.PlaceholderTask))
```

src/test/java/com/cloudbees/jenkins/support/impl/RunningJobsTest.java
Evildethow: imported BOM and added a Pipeline test.

NOTE: Required bumping required from 2.107 -> 2.138.4

src/main/java/com/cloudbees/jenkins/support/impl/RunningJobs.java
Evildethow: Using PrefilteredPrintedContent

src/main/java/com/cloudbees/jenkins/support/impl/RunningJobs.java
Evildethow: Using PrefilteredPrintedContent

Evildethow:
@jglick @MRamonLeon thanks for the review. I've made the changes to the best of my ability.

@aheritier NOTE: To get this working w/ Pipeline the base version had to move from 2.107 -> 2.138.4, and we also need to pull in workflow-durable-task-step as an optional dep.

Thanks, all

src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
```diff
@@ -510,7 +510,7 @@ private static String humanReadableSize(long size) {
             }
             final JenkinsLocationConfiguration jlc = JenkinsLocationConfiguration.get();
 
-            out.println("  * Url:     " + (jlc != null ? ContentFilter.filter(filter, jlc.getUrl()) : "No JenkinsLocationConfiguration available"));
+            out.println("  * Url:     " + (ContentFilter.filter(filter, jlc.getUrl())));
```
aheritier: Why did you remove this Null test @Evildethow  ? (I don't know that code thus maybe it's useless..)

jglick: `get` is `@Nonnull`.

src/test/java/com/cloudbees/jenkins/support/impl/RunningJobsTest.java
jglick: To be clear, the BOM part is just a convenience, but anyway there is little reason to still be putting out releases for very old LTS lines.

pom.xml
```diff
@@ -88,6 +88,18 @@
     </pluginRepository>
   </pluginRepositories>
 
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>io.jenkins.tools.bom</groupId>
+        <artifactId>bom</artifactId>
+        <version>2.138.1</version>
```
jglick:
```suggestion
        <version>2.138.2</version>
```

pom.xml
```diff
@@ -141,6 +153,22 @@
       <version>0.25</version>
       <scope>test</scope>
     </dependency>
+    <!-- Managed by BOM -->
```
jglick: Note that you _could_ also remove version overrides for some other plugins, such as `credentials`.

src/main/java/com/cloudbees/jenkins/support/impl/RunningJobs.java
```diff
@@ -0,0 +1,53 @@
...
+    @NonNull
+    @Override
+    public Set<Permission> getRequiredPermissions() {
+        return Collections.singleton(Jenkins.ADMINISTER);
+    }
+
+    @Override
+    @NonNull
+    public String getDisplayName() {
+        return "Running Jobs";
```
jglick:
```suggestion
        return "Running Builds";
```

(also preferably renaming class to match)

src/main/java/com/cloudbees/jenkins/support/impl/RunningJobs.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @Override
+    public void addContents(@NonNull Container result) {
+        result.add(
+            new PrefilteredPrintedContent("nodes/master/running-jobs.txt") {
```
jglick:
```suggestion
            new PrefilteredPrintedContent("nodes/master/running-builds.txt") {
```

jglick: Perhaps this file should be at top level? It would be pretty misleading for something under `nodes/master/` to list freestyle builds running on an agent. Those _could_ be moved to a separate `nodes/*/running-builds.txt`, though I am not sure if that is helpful or just annoying.

In the case of Pipeline builds, there is no unique associated node, though in many cases there will be exactly one `PlaceholderTask`. (There could be several, if in `parallel`, or none, if paused for example in `input`.)

src/test/java/com/cloudbees/jenkins/support/impl/RunningJobsTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Test
+    public void addContents() throws Exception {
+        FreeStyleProject p = j.createFreeStyleProject(JOB_NAME);
+        QueueTaskFuture<FreeStyleBuild> build = p.scheduleBuild2(0);
+        build.waitForStart();
+
+        String output = SupportTestUtils.invokeComponentToString(new RunningJobs());
+
+        assertThat(output, containsString(p.getName()));
```
jglick: Could perhaps be more specific:
```suggestion
        assertThat(output, containsString(p.getName() + "#1"));
```

or whatever it is.

(BTW not a bad idea to use `JenkinsRule.createFolder` and assert that content contains the full path.)

src/test/java/com/cloudbees/jenkins/support/impl/RunningJobsTest.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    @Test
+    public void addContentsPipeline() throws Exception {
+        WorkflowJob p = j.createProject(WorkflowJob.class, JOB_NAME);
```
jglick: I think you are going to need an actual script here! Or the build will just be some degenerate broken thing. Surprised this test can even pass.

```java
p.setDefinition(new CpsFlowDefinition("node {semaphore 'wait'}", true);
WorkflowRun build = p.scheduleBuild2(0).waitForStart();
SemaphoreStep.waitForStart("wait/1", build);
```

and then you want to assert that the output does contain a representation of the build but does _not_ contain a representation of the `node` block.

src/main/java/com/cloudbees/jenkins/support/impl/RunningJobs.java
```diff
@@ -0,0 +1,53 @@
...
+                @Override
+                protected void printTo(PrintWriter out, ContentFilter filter) {
+                    Optional.ofNullable(Jenkins.getInstanceOrNull())
+                        .ifPresent(jenkins -> Arrays.stream(jenkins.getComputers())
+                            .flatMap(computer -> computer.getAllExecutors().stream())
+                            .collect(Collectors.toList())
+                            .forEach(executor -> Optional.ofNullable(executor.getCurrentExecutable())
+                                .filter(executable -> !(executable.getParent() instanceof ExecutorStepExecution.PlaceholderTask))
```
jglick: Would I think suffice to say
```suggestion
                                .filter(executable -> executable.getParent() == executable.getParent().getOwnerTask())
```

which will filter out `PlaceholderTask`s without needing the dependency on `workflow-durable-task-step`—this could become `test`-scoped.

For reasons I do not fully understand, neither `MatrixConfiguration` nor `MavenModule` seem to override this method, so these would not be filtered out.

pom.xml
Evildethow: yes, understood, although I'm considering it out-of-scope currently.

src/main/java/com/cloudbees/jenkins/support/impl/RunningJobs.java
Evildethow: yes, that makes sense.

src/main/java/com/cloudbees/jenkins/support/impl/RunningJobs.java
Evildethow: Interesting. I'll see what happens when I add some tests.

Evildethow:
Thanks for the review, @jglick

Evildethow:
Requested changes made. @aheritier I think this one is good to go.

src/main/java/com/cloudbees/jenkins/support/impl/RunningBuilds.java
```diff
@@ -0,0 +1,52 @@
...
+        result.add(
+            new PrefilteredPrintedContent("running-builds.txt") {
+                @Override
+                protected void printTo(PrintWriter out, ContentFilter filter) {
+                    Optional.ofNullable(Jenkins.getInstanceOrNull())
+                        .ifPresent(jenkins -> Arrays.stream(jenkins.getComputers())
+                            .flatMap(computer -> computer.getAllExecutors().stream())
+                            .collect(Collectors.toList())
+                            .forEach(executor -> Optional.ofNullable(executor.getCurrentExecutable())
+                                .filter(executable -> executable.getParent() == executable.getParent().getOwnerTask())
```
varyvol: Is it correct filtering before asking if the executable is present?

src/test/java/com/cloudbees/jenkins/support/impl/RunningBuildsTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Test
+    public void addContents() throws Exception {
+        FreeStyleProject p = j.createFreeStyleProject(JOB_NAME);
+        QueueTaskFuture<FreeStyleBuild> freeStyleBuildQueueTaskFuture = p.scheduleBuild2(0);
+        FreeStyleBuild build = freeStyleBuildQueueTaskFuture.waitForStart();
+
+        String output = SupportTestUtils.invokeComponentToString(new RunningBuilds());
+
+        assertThat(output, containsString(String.format(EXPECTED_OUTPUT_FORMAT, p.getName(), build.getNumber())));
```
varyvol: So is this eventually filtered or not?

src/main/java/com/cloudbees/jenkins/support/impl/RunningBuilds.java
Evildethow: Yes. See https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#filter-java.util.function.Predicate-

src/test/java/com/cloudbees/jenkins/support/impl/RunningBuildsTest.java
Evildethow: I don't really understand your question. Sorry.

varyvol: The test is checking the actual job name is displayed in the output string, when that should be something anonymized. Is it that the info is not anonymized or that the utility test function is not taking that into account?

Evildethow: I'll add a test for anonymization.

Evildethow: https://github.com/jenkinsci/support-core-plugin/pull/189/commits/6d986a5207a3633930088c71c9f90cacbf090018

src/test/java/com/cloudbees/jenkins/support/SupportTestUtils.java
```diff
@@ -35,4 +38,24 @@ public void add(@CheckForNull Content content) {
 
     return baos.toString();
   }
+
+    /**
+     * Invoke a component with {@link ContentFilter}, and return the component contents as a String.
+     */
+    public static String invokeComponentToString(final Component component, final ContentFilter filter) {
+        final ByteArrayOutputStream baos = new ByteArrayOutputStream();
```
fcojfernandez:
```suggestion
        try(ByteArrayOutputStream baos = new ByteArrayOutputStream()) {
```

src/test/java/com/cloudbees/jenkins/support/SupportTestUtils.java
```diff
@@ -35,4 +38,24 @@ public void add(@CheckForNull Content content) {
...
+                public void add(@CheckForNull Content content) {
+                    try {
+                        ((PrefilteredContent) Objects.requireNonNull(content)).writeTo(baos, filter);
+                    } catch (IOException e) {
+                        e.printStackTrace();
+                    }
+                }
+            });
+
+        return baos.toString();
```
fcojfernandez:
```suggestion
            return baos.toString();
        }
```

src/main/java/com/cloudbees/jenkins/support/impl/RunningBuilds.java
```diff
@@ -0,0 +1,52 @@
...
+        return "Running Builds";
+    }
+
+    @Override
+    public void addContents(@NonNull Container result) {
+        result.add(
+            new PrefilteredPrintedContent("running-builds.txt") {
+                @Override
+                protected void printTo(PrintWriter out, ContentFilter filter) {
+                    Optional.ofNullable(Jenkins.getInstanceOrNull())
```
fcojfernandez: Why don't just use `Jenkins.get()`? It's already Nonnull.

Evildethow:
> Plugin dependencies could be also cleaned up after moving to BOM

Understood, but I'd rather address that separately. I'm considering it out-of-scope for now.

aheritier:
yes let's cleanup the deps in another PR @Evildethow

Evildethow:
As there seem to be no further requests for changes I'm going to merge this.

Thanks, all.

Evildethow:
I've created https://issues.jenkins-ci.org/browse/JENKINS-59453 to address migrating to the BOM

PR #191, opened 2019-09-16
Avoid capturing an automatic bundle immediately after startup

jglick:
I have noticed that managed masters in CloudBees Core can be unexpectedly slow to start, and thread dumps often implicate automatic support bundle capture. It seems this happens within ~7s of Jenkins starting up, during which time lots of stuff is still being initialized, so attempting to browse the web UI can be quite slow as Jelly compilation competes with background support file writing often of hundreds of Kib of content. Better to defer this for a few minutes until Jenkins is comfortably up and running.

Workaround: `-Dcom.cloudbees.jenkins.support.SupportPlugin.AUTO_BUNDLE_PERIOD_HOURS=0`

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
+++ b/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
@@ -816,6 +816,11 @@ public class SupportPlugin extends Plugin {
         }
 
         @Override
+        public long getInitialDelay() {
+            return TimeUnit.MINUTES.toMillis(5);
+        }
+
+        @Override
         protected synchronized void doRun() throws Exception {
             if (Main.isUnitTest) {
                 return;
--- a/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
+++ b/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
@@ -817,7 +817,7 @@ public class SupportPlugin extends Plugin {
 
         @Override
         public long getInitialDelay() {
-            return TimeUnit.MINUTES.toMillis(5);
+            return TimeUnit.MINUTES.toMillis(3);
         }
 
         @Override


jglick:
Got a `CheckFilterTest` failure, but I also get this locally in `master`. Not sure what it means. Written by @MRamonLeon.

jglick:
I just picked 5m arbitrarily; could easily be 3, or 10. The default _interval_ between bundles is 60m.

Evildethow:
Maybe @aheritier has some ideas here? I'm not sure what best option is, TBH.

aheritier:
@Evildethow @jglick what I don't know is what is the average generation time of a bundle (it may vary a lot from an environment to another) thus it will be hard to define some values which are not too often in conflict.
I would love to not have to generate a bundle at advisor level but as we are allowing our users to tune the generation settings it's not always possible to just grab the latest bundle available on the disk.
Advisor is a service less important/critical than support-core thus maybe we should try to verify if support-core isn't already doing it before launching our generation. maybe by trying every X minutes and if we are failing too much times we trigger a monitor to inform about the issue ...

jglick:
Sorry, I do not know much about Advisor or the possible interactions with it. This was just filed in an attempt to solve a pretty clear startup performance problem affecting any installation with `support-core`.

aheritier:
yes no worries @jglick 
Your PR is fine. We have to see how Advisor could do better. It has to be executed 1 time per day thus maybe we should just increase a lot our initial delay. We don't really care to have it running in 1h or more

jglick:
Ah, I see. Agreed that getting an automated support bundle is a more timely concern.

Evildethow:
@aheritier There is no issue in increasing the Advisor delay. The issue is if the customer is using this version of support-core with the current version of Advisor. In that case, the bundle creation will kick off simultaneously as both would have the same 5min delay.

I'm not sure how you avoid that without changing this value.

My proposal is to change this value `5 -> 3`, then look at increasing the same value in Advisor at a later point.

WDYT?

Evildethow:
@MRamonLeon can you help us understand why this test is failing? See https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fsupport-core-plugin/detail/PR-191/1/tests

aheritier:
yes I agree @Evildethow there is no perfect solution to avoid to have this version of support-core installed with an old version of Advisor.
Let's try this:
* @jglick can you update this PR to a value of 2 or 3 minutes? I still think that it should be triggered before advisor's one. It will give us at least 2 or 3 minutes with no conflict. I'm not sure if it will be enough ... but today we have 5 minutes and no guarantee to not have them running in parallel 
* @Evildethow let's fill an issue/PR to move the Advisor one to 30min by default

aheritier:
Thanks @jglick

jglick:
Still

```
The file 'nodes/slave/computer_deaf_ivory/proc/self/environ' should have the word 'user_grateful_fisherman'.
```

These test failures really read like found poetry!

Evildethow:
Created https://issues.jenkins-ci.org/browse/JENKINS-59454 to address increasing the Advisor plugin delay.

Dohbedoh:
Fails for me locally, every time. From the master branch. From this PR code too. I think this is a bug. It is weird that the build of the master branch worked.
I think this PR is good, and we should file an issue for that test.

MRamonLeon:
> @MRamonLeon can you help us understand why this test is failing? See [ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fsupport-core-plugin/detail/PR-191/1/tests](https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fsupport-core-plugin/detail/PR-191/1/tests)

The test is checking than the word _path_ is filtered in the _nodes/slave/whatever/environment.txt_ file of the bundle. The content of the file is the list of environment variables of the node. The word _path_ is effectively there: ..._CHANGE_BRANCH=delay&amp#0;PATH=/home/jenkins_... so no sure why it was not filtered. I've tried to reproduce it locally with upstream/master branch and with jesse-fork/delay branch without any success.

Maybe some test disabling the filtering interfered with this one? It's the only thing I can think of. 

The cache shouldn't interfere as the Env Vars should be the same. :man_shrugging:

PR #192, opened 2019-09-22
Replace usage of deprecated APIs

darxriggs:


aheritier:
(Also our CI environment is a bit broken these days and I would like that it validates your changes - in addition of the local test we might do)

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -853,8 +852,6 @@ protected synchronized void doRun() throws Exception {
                             cleanupOldBundles(bundleDir, file);
                         } catch (Throwable t) {
                             logger.log(Level.WARNING, "Could not save support bundle", t);
-                        } finally {
-                            SecurityContextHolder.setContext(old);
```
MRamonLeon: is this done when `ACLContext` is closed?

src/main/java/com/cloudbees/jenkins/support/impl/SmartLogFetcher.java
```diff
@@ -106,28 +106,16 @@ public ForNode forNode(Node n) throws IOException {
                     final long localLength = local.length();
                     if (entry.getValue() < Long.MAX_VALUE) {
                         // only copy the new content
-                        FileOutputStream fos = null;
-                        InputStream is = null;
-                        try {
-                            fos = new FileOutputStream(local, true);
-                            is = remoteDir.child(entry.getKey()).readFromOffset(localLength);
+                        try (FileOutputStream fos = new FileOutputStream(local, true);
+                             InputStream is = remoteDir.child(entry.getKey()).readFromOffset(localLength)) {
```
MRamonLeon: Is the _close_ the same as _closeQuietly_? or may the close throw an exception?

src/test/java/com/cloudbees/jenkins/support/slowrequest/InflightRequestTest.java
```diff
@@ -62,7 +62,7 @@ public HttpResponse doSubmitReferer(StaplerRequest req, StaplerResponse rsp) thr
                                  "    <a href='" + req.getContextPath() + "/mockSlowURLCall/submit"+ "' id='link'>link</a>" +
                                  "  </body>" +
                                  "</html>";
-            return HttpResponses.html(literalHTML);
+            return HttpResponses.literalHtml(literalHTML);
```
MRamonLeon: It does `flush` whereas `html` doesn't, not sure if it may impact posterior management of the streams

aheritier:
thanks for these feedbacks @MRamonLeon 
@varyvol Let's wait a bit for this one before merging it.
Maybe @jglick or another core committer could confirm the doubts @MRamonLeon (they look legitimate for me but I don't have the time to test the PR for now)

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
darxriggs: Yes. Have a look [here](https://github.com/jenkinsci/jenkins/blob/jenkins-2.198/core/src/main/java/hudson/security/ACLContext.java#L64).

src/test/java/com/cloudbees/jenkins/support/slowrequest/InflightRequestTest.java
darxriggs: This is just a mock inside a test.

src/main/java/com/cloudbees/jenkins/support/impl/SmartLogFetcher.java
darxriggs: Good catch. The behaviour is different now. When `close` of an `AutoClosable` throws an exception, this is visible now.

The question is: Was the former implementation, hiding exceptions, the better one.

See https://github.com/jenkinsci/jenkins/pull/2830 and [IO-504](https://issues.apache.org/jira/browse/IO-504) for similar changes.

varyvol:
Let's merge given the behaviour change seems reasonable and actually improves the situation.

PR #194, opened 2019-09-26
JENKINS-59453 (OSS-2775) - Cleanup plugin dependencies / Move to BOM

Evildethow:
https://issues.jenkins-ci.org/browse/JENKINS-59453

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -125,11 +125,6 @@
       <optional>true</optional>
     </dependency>
     <dependency>
-      <groupId>org.jenkins-ci.plugins</groupId>
-      <artifactId>credentials</artifactId>
-      <version>2.1.10</version>
-    </dependency>
-    <dependency>
       <groupId>org.kohsuke</groupId>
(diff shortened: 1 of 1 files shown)


varyvol:
@Evildethow I think the version differs from the current one to the one in the BOM. Is that expected/desired?

Evildethow:
> @Evildethow I think the version differs from the current one to the one in the BOM. Is that expected/desired?

@varyvolit

RE: https://github.com/jenkinsci/support-core-plugin/pull/189#pullrequestreview-289830027 https://github.com/jenkinsci/bom/blob/bom-2.138.2/bom/pom.xml#L120

It's intentional.

It was requested in https://github.com/jenkinsci/support-core-plugin/pull/189#pullrequestreview-289830027

varyvol:
@Evildethow yes, but my understanding for cleanup is only removing them from the POM, not updating them. I'm fine, I just wanted to make sure you were aware of the upgrade.

PR #197, opened 2019-10-09
[JENKINS-59714] Remove the pinned/unpinned information

PierreBtz:
* The pinned/unpinned feature is not used in Jenkins 2.x
* Core always returns unpinned
* It makes the active.txt incompatible with downstream tools.


src/test/java/com/cloudbees/jenkins/support/impl/AboutJenkinsTest.java
```diff
@@ -6,6 +6,7 @@
 import com.cloudbees.jenkins.support.SupportTestUtils;
 import com.cloudbees.jenkins.support.api.Component;
 import hudson.ExtensionList;
+import hudson.PluginWrapper;
```
varyvol: Unnecessary I guess?

PR #198, opened 2019-10-18
[JENKINS-59342] Builds / Nodes / Items bundles

Dohbedoh:
[JENKINS-59342](https://issues.jenkins-ci.org/browse/JENKINS-59342): Provide the ability for **Administrators** to download contents for Builds, Items and Nodes. Details in the JIRA.

@MRamonLeon @aheritier @Evildethow @varyvol


Dohbedoh:
Still need to check on those failing tests in windows.

src/main/java/com/cloudbees/jenkins/support/actions/SupportAbstractItemAction.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.logging.Logger;
+
+/**
+ * A {@link SupportObjectAction} applicable to {@link AbstractItem}.
+ */
+public class SupportAbstractItemAction extends SupportObjectAction<AbstractItem> {
+
+    private final Logger logger = Logger.getLogger(SupportAbstractItemAction.class.getName());
```
varyvol: Not used.

src/main/java/com/cloudbees/jenkins/support/actions/SupportComputerAction.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.logging.Logger;
+
+/**
+ * A {@link SupportObjectAction} applicable to {@link Computer}.
+ */
+public class SupportComputerAction extends SupportObjectAction<Computer> {
+
+    private final Logger logger = Logger.getLogger(SupportComputerAction.class.getName());
```
varyvol: Ditto.

src/main/java/com/cloudbees/jenkins/support/actions/SupportObjectAction.java
```diff
@@ -0,0 +1,173 @@
...
+
+        LOGGER.fine("Preparing response...");
+        rsp.setContentType("application/zip");
+
+        JSONObject json = req.getSubmittedForm();
+        if (!json.has("components")) {
+            rsp.sendError(HttpServletResponse.SC_BAD_REQUEST);
+            return;
+        }
+        LOGGER.fine("Parsing request...");
```
varyvol: This is already parsed inside the method.

src/main/java/com/cloudbees/jenkins/support/impl/DumpExportTable.java
```diff
@@ -24,53 +25,94 @@
...
+                }
+        );
+    }
 
-  @NonNull
-  @Override
-  public String getDisplayName() {
-    return "Dump agent export tables (could reveal some memory leaks)";
-  }
+//  @Override
```
varyvol: Why is this commented?

src/main/java/com/cloudbees/jenkins/support/impl/SystemConfiguration.java
```diff
@@ -25,25 +25,22 @@
 package com.cloudbees.jenkins.support.impl;
 
 import com.cloudbees.jenkins.support.AsyncResultCache;
-import com.cloudbees.jenkins.support.api.CommandOutputContent;
-import com.cloudbees.jenkins.support.api.Container;
-import com.cloudbees.jenkins.support.api.UnfilteredCommandOutputContent;
-import com.cloudbees.jenkins.support.api.UnfilteredStringContent;
+import com.cloudbees.jenkins.support.api.*;
```
varyvol: Better to avoid star imports in production code.

src/main/java/com/cloudbees/jenkins/support/impl/SystemConfiguration.java
```diff
@@ -25,25 +25,22 @@
...
+import org.kohsuke.stapler.DataBoundConstructor;
 
 import java.io.File;
 import java.io.IOException;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
-import java.util.WeakHashMap;
+import java.util.*;
```
varyvol: Ditto.

src/main/java/com/cloudbees/jenkins/support/impl/DirectoryComponent.java
```diff
@@ -0,0 +1,234 @@
...
+                              boolean useDefaultExcludes, 
+                              boolean followSymlinks) {
+            this.includes = includes;
+            this.excludes = excludes;
+            this.useDefaultExcludes = useDefaultExcludes;
+            this.followSymlinks = followSymlinks;
+        }
+
+        public void scan(File dir, FileVisitor visitor) throws IOException {
+            if (Util.fixEmpty(this.includes) == null && this.excludes == null) {
```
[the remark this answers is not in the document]
Dohbedoh: Should be an "and". We do not use any glob in that case.

src/main/java/com/cloudbees/jenkins/support/impl/DumpExportTable.java
Dohbedoh: Should be assignable to Jenkins and Computer.

src/main/java/com/cloudbees/jenkins/support/configfiles/AgentsConfigFile.java
```diff
@@ -58,20 +68,76 @@ public boolean isSelectedByDefault() {
...
-            File config = new File(agentDir, "config.xml");
-            container.add(new XmlRedactedSecretFileContent("nodes/slave/{0}/config.xml", new String[]{agentDir.getName()}, config));
-        }
+        File agentDir = new File(Jenkins.get().getRootDir(), MessageFormat.format("nodes/{0}", item.getName()));
+        File config = new File(agentDir, "config.xml");
+        container.add(new XmlRedactedSecretFileContent("nodes/slave/" + agentDir.getName() + "/config.xml", config));
     }
+    
+    @Extension
+    public static class AgentConfigFile extends ObjectComponent<Computer> {
```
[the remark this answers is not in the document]
Dohbedoh: yeah. It could actually be a single class. Let me fix this.

src/main/java/com/cloudbees/jenkins/support/impl/DirectoryComponent.java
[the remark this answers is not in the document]
Dohbedoh: Actually we could just always use the glob approach, simpler.

src/main/java/com/cloudbees/jenkins/support/api/ObjectComponent.java
```diff
@@ -0,0 +1,126 @@
...
+                .map(component -> ((ObjectComponent<T>) component))
+                .filter(component -> component.isApplicable(item))
+                .collect(Collectors.toList());
+    }
+
+    /**
+     * All applicable {@link ObjectComponentDescriptor}s for the class.
+     */
+    public static <T extends AbstractModelObject> List<ObjectComponentDescriptor<T>> for_(T item) {
+        return Jenkins.get().getExtensionList(ObjectComponent.class).stream()
```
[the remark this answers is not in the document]
Dohbedoh: yes. Thanks!

Dohbedoh:
> One question: what happens when the content to be zipped and downloaded is big? Are there issues creating the ZIP? Or downloading it?

@varyvol This is going through the same method than the root Support Bundle generation. Subjected to the same technical vulnerabilities.

src/main/java/com/cloudbees/jenkins/support/api/Component.java
```diff
@@ -102,9 +103,26 @@ public boolean isSelectedByDefault() {
...
     }
 
+    /**
+     * Return if this component is applicable to a specific class of item.
+     *
+     * @param clazz the class
+     * @param <C> Object that extends {@link AbstractModelObject}
+     * @return
+     */
+    public <C extends AbstractModelObject> boolean isApplicable(Class<C> clazz) {
```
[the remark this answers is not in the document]
Dohbedoh: To be able to have component that can be used at root level, and also at another level (`Computer`, `Build`, `Item`, ....). This method is used to list the components that can be used by from an Action. For example, we have a component `DumpExportTable` that get the exports table of all nodes when used from the root action. We can use the same component from the Node action to get the content only for the node.
Maybe there is a better design for this. I am open to suggestions.

src/main/java/com/cloudbees/jenkins/support/impl/AbstractItemDirectoryComponent.java
```diff
@@ -0,0 +1,161 @@
...
+                    container.add(new PrintedContent("items/{0}/{1}", relativeToRoot.toString(), relativePath) {
+
+                        @Override
+                        protected void printTo(PrintWriter out) {
+                            out.println("symlink -> " + target);
+                        }
+
+                        @Override
+                        public boolean shouldBeFiltered() {
+                            return true;
```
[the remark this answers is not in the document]
Dohbedoh: I do think it will still be useful yes.

varyvol:
@Dohbedoh but the same technical limitations might work for a situation and not for others. For example, imagine a limitation is the bundle has a maximum capacity of 10mb that is very rarely reached with the usual bundle but will always be reached through this use case. That would make this use case unusable in the practice.

Dohbedoh:
I will check on this. Tests are failing on Windows. I have tested the plugin in a instance running on windows and confirmed that it works. I haven't run the tests in a windows environment though...

Dohbedoh:
I did find out that the ZIP entries in the case of windows have a mix of paths, for example `items/testFolder/jobs/testFreestyle/builds\1\build.xml`. We should make sure that ZipEntry uses only forward slash.

PR #199, opened 2019-10-24
[JENKINS-59696] Fix ContentFilters dynamic loading issue

Dohbedoh:
[JENKINS-59696](https://issues.jenkins-ci.org/browse/JENKINS-59696): This should fix the dynamic loading issues we encounter. The main fix is what has been done in the `DeadlockTrackChecker`. 
Note: Based on my findings, the use of `GlobalConfiguration.all().get(ContentFilters.class)` instead of `ExtensionList.lookupSingleton(ContentFilters.class)` seems to be general workaround to prevent loading the extension twice.

@MRamonLeon @jglick

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/filter/ContentFilters.java
+++ b/src/main/java/com/cloudbees/jenkins/support/filter/ContentFilters.java
@@ -25,7 +25,6 @@
 package com.cloudbees.jenkins.support.filter;
 
 import hudson.Extension;
-import hudson.ExtensionList;
 import jenkins.model.GlobalConfiguration;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
@@ -44,7 +43,7 @@ import javax.annotation.Nonnull;
 public class ContentFilters extends GlobalConfiguration {
 
     public static ContentFilters get() {
-        return ExtensionList.lookupSingleton(ContentFilters.class);
+        return GlobalConfiguration.all().get(ContentFilters.class);
     }
 
     private boolean enabled;
--- a/src/main/java/com/cloudbees/jenkins/support/timer/DeadlockTrackChecker.java
+++ b/src/main/java/com/cloudbees/jenkins/support/timer/DeadlockTrackChecker.java
@@ -55,10 +55,10 @@ public class DeadlockTrackChecker extends PeriodicWork {
                 builder.println("==============");
                 ThreadInfo[] deadLockThreads = mbean.getThreadInfo(deadLocks, Integer.MAX_VALUE);
 
-                Optional<ContentFilter> contentFilter = SupportPlugin.getContentFilter();
+                ContentFilter contentFilter = SupportPlugin.getContentFilter().orElse(null);
                 for (ThreadInfo threadInfo : deadLockThreads) {
                     try {
-                        ThreadDumps.printThreadInfo(builder, threadInfo, mbean, contentFilter.orElse(null));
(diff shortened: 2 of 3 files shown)


src/main/java/com/cloudbees/jenkins/support/timer/DeadlockTrackChecker.java
```diff
@@ -58,7 +55,8 @@ protected void doRun() throws Exception {
 
                 for (ThreadInfo threadInfo : deadLockThreads) {
                     try {
-                        ThreadDumps.printThreadInfo(builder, threadInfo, mbean, filter);
+                        ThreadDumps.printThreadInfo(builder, threadInfo, mbean,
```
varyvol: I guess if we get the filter before the for loop, it will still be valid? And so we avoid loading it per each thread.

src/main/java/com/cloudbees/jenkins/support/filter/ContentFilters.java
```diff
@@ -44,7 +43,7 @@
 public class ContentFilters extends GlobalConfiguration {
 
     public static ContentFilters get() {
-        return ExtensionList.lookupSingleton(ContentFilters.class);
+        return GlobalConfiguration.all().get(ContentFilters.class);
```
varyvol: Do you think it's actually this and not DeadlockTrackChecker requiring this before it should? Have you tested that scenario?

src/main/java/com/cloudbees/jenkins/support/timer/DeadlockTrackChecker.java
Dohbedoh: good point

src/main/java/com/cloudbees/jenkins/support/filter/ContentFilters.java
Dohbedoh: I think that the initialization of the `final ContentFilter` field in `DeadlockTrackChecker` was causing the problem.

But my findings were that `ExtensionList.lookupSingleton` is always returning an instance of the `ContentFilters` and may create the `ExtensionList` if it does not exist yet. In that case, and in the scenario of dynamic loading of the plugin, it seems to potentially cause the same extension to be loaded twice. By using `GlobalConfiguration.all()` we go through a different cause path that seems less prone to this error. I'd love that someone to confirm this.

> Have you tested that scenario?

I have only tested:
* the fix of `ContentFilters#get` alone: OK
* the fix of `ContentFilters#get` and `DeadlockTrackChecker` together: OK

I can test the fix for `DeadlockTrackChecker`  in isolation quickly.

src/main/java/com/cloudbees/jenkins/support/timer/DeadlockTrackChecker.java
varyvol: Not a blocker, but you could have added the orElse() at the beginning as well.

src/main/java/com/cloudbees/jenkins/support/filter/ContentFilters.java
varyvol: I don't know it exactly either but if you have the theory and it's not hurting, sounds good.

MRamonLeon: I'm a bit concern that it may not be the fix and the tests were not failing because it's hard to reproçduce. Maybe you can try to reproduce the problem deterministically in a test even in Jenkins core, instead of the plugin itself, you could take ideas from `PluginManagerTest` (`dynamicLoad`) or `InstallPluginCommandTest`.

PR #200, opened 2019-11-05
Fix the checking of env vars using an existing env var instead of path

MRamonLeon:
[JENKINS-59455](https://issues.jenkins-ci.org/browse/JENKINS-59455)

- Use the first available env variable, and not path, to be sure it's available.
- Wait for the created agent to be online.
- Consider '\0' (NUL character) as a character that appears to mark the end of a line, and so it must not make files containing it to be considered as binary files.

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/filter/FilteredConstants.java
+++ b/src/main/java/com/cloudbees/jenkins/support/filter/FilteredConstants.java
@@ -28,7 +28,7 @@ import java.util.regex.Pattern;
 
 final class FilteredConstants {
 
-    static final Pattern EOL = Pattern.compile("\r?\n");
+    static final Pattern EOL = Pattern.compile("\r?\n|" + '\0');
     static final int DEFAULT_DECODER_CAPACITY = 1024;
 
     private FilteredConstants() {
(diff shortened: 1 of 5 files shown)


batmat:
Still failing the exact same way for me AFAICT.

MRamonLeon:
> Still failing the exact same way for me AFAICT.

The CI is running, which is the important point, I guess.

Anyway, do you mind @batmat to add:
```
System.getenv().entrySet().forEach(entry -> System.out.println(entry.getKey() + "=" + entry.getValue()));
```

on the line 211 above: https://github.com/jenkinsci/support-core-plugin/pull/200/files#diff-35a14f6121db5e87b98228af3c8c66a1R211

To see what your environment looks like

MRamonLeon:
@alecharp was it failing in your laptop? if so, could you please check this PR?

`mvn test -Dtest=CheckFilterTest`

alecharp:
It is still failing.

varyvol:
@alecharp @batmat I've done some changes, could you please verify it's working for you now?

varyvol:
@jvz you might want to review this, given the streams were implemented by you (if I'm not wrong).

varyvol:
@jvz you are more than right. I've updated both the description in this PR and the ticket itself.

varyvol:
@jvz I'm merging this. If you think some more details are necessary, please tell me and I will change the descriptions then.

PR #201, opened 2019-11-18
[JENKINS-59498] Improve computation with DirectoryStream

Dohbedoh:
[JENKINS-59498](https://issues.jenkins-ci.org/browse/JENKINS-59498): this methods is currently using `getBuild` in an `allItems` loop that is really expensive in terms of computation. 

Proposing another approach similar to the old one that was [scanning the build directory](https://github.com/jenkinsci/support-core-plugin/blob/support-core-2.55/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java#L621-L672) but using `java.nio.DirectoryStream` and improving the loop by [moving conditional](https://github.com/jenkinsci/support-core-plugin/blob/support-core-2.55/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java#L636) outside the loop / stream.

Did some benchmarking using different approach in a Jenkins instance that has ~8400 jobs / ~130000 build. The instance has 4GB memory / 1.0 CPU in a containerized environment and tested different approach:

* current implementation: not able to get a support bundle within 30 seconds, even with ONLY the `ItemsContent` enabled and no activity on the instance.
* `java.nio.Files.list` approach: Average of **3.28s** over 5 bundle generations
* `java.nio.Files.newDirectorStream` approach: Average of **1.48s** over 5 bundle generations

I also generated some bundle while the instance had ~40 builds in queue and 30 kubernetes agents connected / building. But the time were only slightly impacted by less than 0.5s.

@varyvol @fbelzunc @MRamonLeon


Dohbedoh:
I hitting a known issue with spotbug / JDK 11: https://github.com/spotbugs/spotbugs/issues/756 :(

Dohbedoh:
Refactor the code with the try with resources in  a single function and added the warning suppression on that function.

src/main/java/com/cloudbees/jenkins/support/impl/ItemsContent.java
```diff
@@ -65,97 +69,134 @@ public Set<Permission> getRequiredPermissions() {
...
+                jenkins.allItems().forEach(item -> {
+                    String key = item.getClass().getName();
+                    Integer cnt = containerCounts.get(key);
+                    containerCounts.put(key, cnt == null ? 1 : cnt + 1);
+                    if (item instanceof Job) {
+                        Job<?, ?> j = (Job) item;
+                        // too expensive: int builds = j.getBuilds().size();
+                        int builds = 0;
+                        File buildDir = jenkins.getBuildDirFor(j);
+                        if(new File(buildDir, "legacyIds").isFile()) {
```
jglick, line 98: #639

PR #202, opened 2019-11-29
JENKINS-60327 Enable tests on windows

aheritier:
@escoem does it makes sense firstly to do such change to detect this kind of issue ?
cc @MRamonLeon @Dohbedoh 
REF: https://issues.jenkins-ci.org/browse/JENKINS-60327

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,1 +1,7 @@
-buildPlugin()
+#!/usr/bin/env groovy
+
+/* `buildPlugin` step provided by: https://github.com/jenkins-infra/pipeline-library */
+buildPlugin(configurations: [
+  [ platform: "linux", jdk: "8" ],
+  [ platform: "windows", jdk: "8" ]
+])


aheritier:
you are right @varyvol 
Windows is tested by default (I thought it was linux only).
Why don't we catch JENKINS-60327 ?

varyvol:
@aheritier I didn't mean you to close it. Maybe it was a good opportunity to start building in JDK11 as well.

aheritier:
@varyvol my PR didn't do more than the default behaviour? java 11 could make sense but it is probably something we should also configure by default ?

varyvol:
@aheritier I think using [the recommended configurations|https://github.com/jenkins-infra/pipeline-library/blob/master/vars/buildPlugin.groovy#L252] would be nice.

aheritier:
@varyvol good idea. I wasn't aware of this recommended config. Let's do it !

PR #203, opened 2019-11-29
Build/Test using the recommended config

aheritier:
Ref: https://github.com/jenkins-infra/pipeline-library/blob/master/vars/buildPlugin.groovy#L252

FYI @varyvol 

(I didn't create a Jira task for such update - I don't think it is useful)

WDYT ?

PR #204, opened 2019-12-02
Build/Test using the recommended config

varyvol:
Utility PR to validate #203


varyvol:
Building fine after disabling the warning due to Spotbugs issue.

PR #208, opened 2020-01-20
[JENKINS-60009] Add support for Java 9 and later for GCLogs

Dohbedoh:
[JENKINS-60009](https://issues.jenkins-ci.org/projects/JENKINS/issues/JENKINS-60009): The java arguments to enable GC logs are different since Java9. This PR adds support for it, see also https://docs.oracle.com/javase/9/tools/java.htm#GUID-BE93ABDC-999C-4CB5-A88B-1994AAAC74D5__CONVERTGCLOGGINGFLAGSTOXLOG-A5046BD1.

In order to write tests for this, we need to have tests that execute in JDK 8 or older and tests that execute in JDK 9 or later. So I used the `org.junit.Assume` for this.

Note: Since the 2.164.1 LTS, there is a class [jenkins.utils.java.JavaUtils](https://github.com/jenkinsci/jenkins/commit/412da254f2bbac386e9ab736684140cf107a3ffd#diff-f820412d63e84d362e303aeaa8be7729) that can be used to check if the runtime version is older than 8 or not. But I did not want to raise the version in the pom as this could prevent people that are running older version of core to upgrade the support core plugin. I wonder what are your thoughts about it ?

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/GCLogs.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/GCLogs.java
@@ -149,27 +149,27 @@ public class GCLogs extends Component {
 
         String fileLocation;
         
-        if(!isJava8OrBelow()) {
-            String gcLogSwitch = vmArgumentFinder.findVmArgument(GCLOGS_JRE9_SWITCH);
-            if (gcLogSwitch == null) {
-                LOGGER.fine("No GC Logging switch found, cannot collect gc logging files.");
-                fileLocation = null;
-            } else {
-                Matcher fileLocationMatcher = Pattern.compile(GCLOGS_JRE9_LOCATION).matcher(gcLogSwitch);
-                if(fileLocationMatcher.matches()) { 
-                    fileLocation = fileLocationMatcher.group("location");
-                } else {
-                    LOGGER.fine("No GC Logging custom file location found.");
-                    fileLocation = null;
-                }
-            }
-        } else {
+        if(isJava8OrBelow()) {
             String gcLogSwitch = vmArgumentFinder.findVmArgument(GCLOGS_JRE_SWITCH);
             if (gcLogSwitch == null) {
                 LOGGER.fine("No GC Logging switch found, cannot collect gc logging files.");
                 fileLocation = null;
             } else {
                 fileLocation = gcLogSwitch.substring(GCLOGS_JRE_SWITCH.length());
+            }
+        } else {
+            String gcLogSwitch = vmArgumentFinder.findVmArgument(GCLOGS_JRE9_SWITCH);
+            if (gcLogSwitch == null) {
+                LOGGER.fine("No GC Logging switch found, cannot collect gc logging files.");
+                fileLocation = null;
+            } else {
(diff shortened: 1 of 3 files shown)


src/main/java/com/cloudbees/jenkins/support/impl/GCLogs.java
```diff
@@ -145,16 +147,36 @@ private void handleRotatedLogs(@Nonnull final String gcLogFileLocation, Containe
     @CheckForNull
     private String getGcLogFileLocation() {
 
-        String gcLogSwitch = vmArgumentFinder.findVmArgument(GCLOGS_JRE_SWITCH);
-        if (gcLogSwitch == null) {
-            LOGGER.fine("No GC Logging switch found, cannot collect gc logging files.");
-            return null;
+        String fileLocation;
+        
+        if(!isJava8OrBelow()) {
```
varyvol: I tend to prefer the main branches for the "positive" value, I think it's less confusing. WDYT?

src/main/java/com/cloudbees/jenkins/support/impl/GCLogs.java
```diff
@@ -145,16 +147,36 @@ private void handleRotatedLogs(@Nonnull final String gcLogFileLocation, Containe
...
+                fileLocation = gcLogSwitch.substring(GCLOGS_JRE_SWITCH.length());
+            }
         }
-        return gcLogSwitch.substring(GCLOGS_JRE_SWITCH.length());
+        return fileLocation;
     }
     
     public boolean isFileLocationParameterized(String fileLocation) {
-        return Pattern.compile(".*%[tnp].*").matcher(fileLocation).find();
+        return Pattern.compile(".*%[tp].*").matcher(fileLocation).find();
```
varyvol: Is this related to the change?

varyvol:
@Dohbedoh tests seem to be failing for Windows with Java 11.

src/main/java/com/cloudbees/jenkins/support/impl/GCLogs.java
Dohbedoh: This was a leftover of a previous PR https://github.com/jenkinsci/support-core-plugin/pull/182/commits that I missed: https://github.com/jenkinsci/support-core-plugin/pull/182/commits/66970ac42f0f1e7223bad7c38c951446ec8dc68d. But did not affect the feature.

src/main/java/com/cloudbees/jenkins/support/impl/GCLogs.java
Dohbedoh: good point, will change this.

Dohbedoh:
I thought this was a file separator problem but I think that this is a quoting problem: http://mail.openjdk.java.net/pipermail/jdk9-dev/2017-August/005950.html. The file location in Windows must be quoted: `-Xlog:gc*:file=\"C:\example\gc.%t.log.%p\"`.

Dohbedoh:
So the quote is not the problem, although I believe that we should handle the case `file="<location>"`. The problem is that I use `:` to mark the end of file location in the system property. Because the syntax of the `Xlog` property that uses colon, see [Enable Logging with the JVM Unified Logging Framework](https://docs.oracle.com/javase/9/tools/java.htm#GUID-BE93ABDC-999C-4CB5-A88B-1994AAAC74D5).

PR #209, opened 2020-02-12
Improve RunningBuildsTest

darxriggs:
Make sure that builds are really running when calling the support tooling.

So far only the tests for pipelines made sure with a semaphore that the build is still running. The tests for FreeStyle jobs worked by chance.

Scheduling a build now is also a bit simplified.

### DIFF
--- a/src/test/java/com/cloudbees/jenkins/support/impl/RunningBuildsTest.java
+++ b/src/test/java/com/cloudbees/jenkins/support/impl/RunningBuildsTest.java
@@ -6,9 +6,12 @@ import com.cloudbees.jenkins.support.filter.ContentFilter;
 import com.cloudbees.jenkins.support.filter.ContentFilters;
 import com.cloudbees.jenkins.support.filter.ContentMapping;
 import com.cloudbees.jenkins.support.filter.ContentMappings;
+import hudson.Launcher;
+import hudson.model.AbstractBuild;
+import hudson.model.BuildListener;
 import hudson.model.FreeStyleBuild;
 import hudson.model.FreeStyleProject;
-import hudson.model.queue.QueueTaskFuture;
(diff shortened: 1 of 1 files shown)


darxriggs:
If there is something like `SemaphoreStep` but for FreeStyle jobs already available let me know.

src/test/java/com/cloudbees/jenkins/support/impl/RunningBuildsTest.java
```diff
@@ -101,4 +109,20 @@ public void addContentsPipeline() throws Exception {
...
         assertThat(output, containsString(String.format(EXPECTED_OUTPUT_FORMAT, p.getName(), workflowRun.getNumber())));
     }
+
+    private static class SemaphoreBuilder extends TestBuilder {
+
+        private final OneShotEvent event = new OneShotEvent();
+
+        @Override
+        public boolean perform(AbstractBuild<?, ?> build, Launcher launcher, BuildListener listener) throws InterruptedException {
+            System.err.println("Waiting for semaphore...");
```
varyvol: Why the error log?

darxriggs: It resembles the logging in [SemaphoreStep](https://github.com/jenkinsci/workflow-support-plugin/blob/workflow-support-3.4/src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java). When using this builder in a test, this output is easily visible next to the normal output from Jenkins. If `BuildListener` would be used, this would be harder to access.

darxriggs:
@varyvol friendly reminder

varyvol:
Tested locally and tests pass fine. Merging.

PR #211, opened 2020-03-18
Name of the category for this management link

aheritier:
Used only by Jenkins core 2.226+

Ref:
* https://jenkins.io/changelog/#v2.226
* jenkinsci/jenkins#4546

I hesitated between SECURITY and TROUBLESHOOTING categories but I think that the security concern of this feature is more important

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
+++ b/src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
@@ -246,4 +246,12 @@ public class ContentMappings extends ManagementLink implements Saveable, Iterabl
     public @Nonnull String getUrlName() {
         return "anonymizedMappings";
     }
+    
+    /**
+     * Name of the category for this management link.
+     * TBD: Use getCategory when core requirement is greater or equal to 2.226 
+     */
+    public @Nonnull String getCategoryName() {
+        return "SECURITY";
+    }    
 }


aheritier:
FYI @daniel-beck @Dohbedoh

varyvol:
@aheritier I tend to think TROUBLESHOOTING is a better option.

aheritier:
I hesitated a lot @varyvol :-)
I let others giving their feedbacks.

daniel-beck:
I don't understand security at all. It's Status or Troubleshooting (and based on its predominant use case IIUC, Troubleshooting).

aheritier:
@daniel-beck because this the configuration of the security settings (anonymization) for the support plugin.

daniel-beck:
Good point, had forgotten about https://github.com/jenkinsci/jenkins/pull/4578#issuecomment-599991359 already.

Security seems weird to me for this, but it's not completely wrong.

aheritier:
@daniel-beck cc @MarkEWaite  @oleg-nenashev I have no problem to put it under TROUBLESHOOTING like advisor ( https://github.com/jenkinsci/cloudbees-jenkins-advisor-plugin/pull/59 ) 

@daniel-beck about https://github.com/jenkinsci/jenkins/pull/4578

> It would be more straightforward if this was part of the "Support" action 

It's a specific configuration entry outside of the support because it applies to bundles generated in the background and not only from the UI (Support action).

But AFAIR now the activation of the anonymization isn't in this screen but in the global settings (this screen just give the mapping between the anonymised data and the real ones. thus not sure if it should be TROUBLESHOOTING or something else

daniel-beck:
@aheritier 

> It's a specific configuration entry outside of the support because it applies to bundles generated in the background and not only from the UI (Support action).

Right, hence "it _would_ be more straightforward". As is, it's clearly not Troubleshooting. If we count Privacy in Security, it's clearly there.

MarkEWaite:
Based on your description @daniel-beck and the comments from @aheritier , I agree that it is not Troubleshooting.  I prefer Security because it adjusts the amount of information shared with others.

aheritier:
ok thus I don't touch to my pull-request. 
SECURITY seems to be the most appropriate category for this management link.
I let the plugins owner merging it, etc ..

varyvol:
Happy to merge it when it gets some more approvals (which seems straight given the consensus).

PR #212, opened 2020-03-23
[JENKINS-61659] Allow to generate custom support bundle through api

Vlatombe:
See [JENKINS-61659](https://issues.jenkins-ci.org/browse/JENKINS-61659)

* Expose components through api
* New endpoint generating a support bundle only with included components

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/SupportAction.java
+++ b/src/main/java/com/cloudbees/jenkins/support/SupportAction.java
@@ -28,6 +28,7 @@ import com.cloudbees.jenkins.support.api.Component;
 import com.cloudbees.jenkins.support.api.SupportProvider;
 import com.cloudbees.jenkins.support.filter.ContentFilters;
 import hudson.Extension;
+import hudson.model.Api;
 import hudson.model.RootAction;
 import hudson.security.ACL;
 import hudson.security.ACLContext;
@@ -38,9 +39,13 @@ import org.jvnet.localizer.Localizable;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.QueryParameter;
 import org.kohsuke.stapler.StaplerProxy;
 import org.kohsuke.stapler.StaplerRequest;
 import org.kohsuke.stapler.StaplerResponse;
+import org.kohsuke.stapler.WebMethod;
+import org.kohsuke.stapler.export.Exported;
+import org.kohsuke.stapler.export.ExportedBean;
 import org.kohsuke.stapler.interceptor.RequirePOST;
 
 import javax.servlet.ServletException;
@@ -49,16 +54,19 @@ import javax.servlet.http.HttpServletResponse;
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 
 /**
  * Main root action for generating support.
  */
 @Extension
+@ExportedBean
 public class SupportAction implements RootAction, StaplerProxy {
     public static final Permission CREATE_BUNDLE = SupportPlugin.CREATE_BUNDLE;
     /**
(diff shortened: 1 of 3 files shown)


src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -202,9 +226,29 @@ public void doGenerateAllBundles(StaplerRequest req, StaplerResponse rsp) throws
...
+    }
+
+    /**
+     * Generates a support bundle with only requested components.
+     * @param components component names separated by comma.
+     * @param rsp
+     * @throws IOException
+     */
+    @RequirePOST
+    public void doGenerateBundle(@QueryParameter("components") String components, StaplerResponse rsp) throws IOException {
```
varyvol: I'm missing some permission check here?

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -202,9 +226,29 @@ public void doGenerateAllBundles(StaplerRequest req, StaplerResponse rsp) throws
...
+        if (components == null) {
+            rsp.sendError(HttpServletResponse.SC_BAD_REQUEST, "components parameter is mandatory");
+            return;
+        }
+        List<String> componentNames = Arrays.asList(components.split(","));
+        logger.fine("Selecting components...");
+        prepareBundle(rsp, getComponents().stream().filter(c -> componentNames.contains(c.getId())).collect(Collectors.toList()));
```
varyvol: It would be a user error but maybe it should check if the list is not empty?

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
Vlatombe: no, there is already a global check [here](https://github.com/jenkinsci/support-core-plugin/blob/5312de361286eef67f867a039dd33cc48915d2d4/src/main/java/com/cloudbees/jenkins/support/SupportAction.java#L80)

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
Vlatombe: Done.

Vlatombe:
@varyvol My use case is an automated support bundle generation through the Rest API following an ATH run, where I'm collecting only specific logs relevant to the ATH.

varyvol:
@Vlatombe your use case seems safe, but if this is opened, somebody else could have another use case where the performance impact could cause consequences.

Vlatombe:
@varyvol how so? Anything you can generate with the new endpoint can already be generated with the existing one.

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -202,9 +226,34 @@ public void doGenerateAllBundles(StaplerRequest req, StaplerResponse rsp) throws
...
+    }
+
+    /**
+     * Generates a support bundle with only requested components.
+     * @param components component names separated by comma.
+     * @param rsp
+     * @throws IOException
+     */
+    @RequirePOST
+    public void doGenerateBundle(@QueryParameter("components") String components, StaplerResponse rsp) throws IOException {
```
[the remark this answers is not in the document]
Vlatombe: No, it includes the components flagged with `isSelectedByDefault`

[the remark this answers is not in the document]
MRamonLeon: Sorry @Vlatombe, I cannot see in the code what you say. I mean, `doGenerateAllBundles` calls `getComponents` which gets all defined components, default or not. After that, it removes the components added to `request->json->components` element with `isSelected=false`. The UI shows the components checked or not based on whether they're selected by default or not. But if you call `doGenerateAllBundles` with `components` empty, you get all the components, not only the ones by default. So you can use the method for your purpose, saving a new method. But you still needs a way to get all existing components, the `getComponents` _webmethod_, that is mandatory.  I'm not strictly against adding a new method, but it's not needed, if I analyzed it well. Maybe I missed something.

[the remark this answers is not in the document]
Vlatombe: Ok, I agree with your statement that all components are picked up by default. In any case it is painful and complicates the assembly of a bundle when you know what you want to get.

The other thing that is a problem reusing `doGenerateAllBundles` is that it has a side-effect on the plugin configuration https://github.com/jenkinsci/support-core-plugin/blob/8724c737b0a76b6eb7848eb443a772f3f7e1958e/src/main/java/com/cloudbees/jenkins/support/SupportAction.java#L225-L228 This is not desirable when not calling from UI.

varyvol:
@Vlatombe can you maybe add the test that @jeffret-b mentions, and I think this will be ready to be merged?

Vlatombe:
@varyvol Seems like a nice to have and unrelated to my change. I think this could be tackled separately.

Vlatombe:
@varyvol thanks!

PR #214, opened 2020-04-03
Use Computer.threadPoolForRemoting rather than creating a new Executors.newFixedThreadPool

jglick:
Noticed a support bundle that had a ton of anonymous pool threads coming from this class.


varyvol:
Tested locally. Merging.

PR #218, opened 2020-04-11
Test improvement - Consistent static imports for assertions

darxriggs:


src/test/java/com/cloudbees/jenkins/support/filter/ContentMappingsTest.java
```diff
@@ -95,7 +96,7 @@ public void contentMappingsOrderedByLengthDescending() throws IOException {
         List<String> originals = StreamSupport.stream(mappings.spliterator(), false)
                 .map(ContentMapping::getOriginal)
                 .collect(toList());
-        Assertions.assertThat(originals).containsExactly("LongestNameHere", "LongerName", "ShortName");
+        assertThat(originals, contains("LongestNameHere", "LongerName", "ShortName"));
```
varyvol, line 99: Then it should verify the size as well?

src/test/java/com/cloudbees/jenkins/support/impl/SlaveCommandStatisticsTest.java
```diff
@@ -36,15 +36,16 @@
...
-import static org.hamcrest.Matchers.*;
 import org.junit.Test;
-import static org.junit.Assert.*;
 import org.junit.Rule;
 import org.jvnet.hudson.test.Issue;
 import org.jvnet.hudson.test.JenkinsRule;
 import org.jvnet.hudson.test.LoggerRule;
 import org.jvnet.hudson.test.TestBuilder;
 
+import static org.hamcrest.Matchers.*;
```
varyvol, line 46: Why this one if the actual doesn't change?

src/test/java/com/cloudbees/jenkins/support/filter/ContentMappingsTest.java
darxriggs: The `contains` Hamcrest matcher already does check the size, see the [Javadoc](http://hamcrest.org/JavaHamcrest/javadoc/2.2/org/hamcrest/Matchers.html#contains-E...-). It's just named differently than the AssertJ matcher which was used before. The Hamcrest matcher `hasItems` is the one that only checks for the specified items without the size, see [Javadoc](http://hamcrest.org/JavaHamcrest/javadoc/2.2/org/hamcrest/Matchers.html#hasItems-T...-).

src/test/java/com/cloudbees/jenkins/support/impl/SlaveCommandStatisticsTest.java
darxriggs: Because now in all test classes the block of static imports is placed below non-static ones.

varyvol: But these are not movements but additions? I mean, if they were not necessary before, why are they now?

src/test/java/com/cloudbees/jenkins/support/filter/ContentMappingsTest.java
varyvol: Thanks for the clarification.

src/test/java/com/cloudbees/jenkins/support/impl/SlaveCommandStatisticsTest.java
darxriggs: `import static org.hamcrest.Matchers.*;` is moved from line 39 to line 46.

darxriggs: `import static org.junit.Assert.*;` is moved from line 41 to line 46 and the not required star import was simplified to importing only the single required method.

varyvol: Sorry, I didn't see the other ones :(

PR #220, opened 2020-04-25
[JENKINS-62054] Ensure the Support action is not displayed if the use…

PierreBtz:
…r does not have the proper rights

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/actions/SupportAbstractItemAction.java
+++ b/src/main/java/com/cloudbees/jenkins/support/actions/SupportAbstractItemAction.java
@@ -22,7 +22,7 @@ public class SupportAbstractItemAction extends SupportObjectAction<AbstractItem>
 
     @Override
     public String getDisplayName() {
-        return Messages.SupportItemAction_DisplayName();
+        return shouldDisplay() ? Messages.SupportItemAction_DisplayName() : null;
     }
     
     @Extension
--- a/src/main/java/com/cloudbees/jenkins/support/actions/SupportObjectAction.java
+++ b/src/main/java/com/cloudbees/jenkins/support/actions/SupportObjectAction.java
@@ -59,12 +59,16 @@ public abstract class SupportObjectAction<T extends AbstractModelObject> impleme
 
     @Override
     public String getUrlName() {
-        return "support";
+        return shouldDisplay() ? "support" : null;
     }
 
     @Override
     public String getIconFileName() {
-        return "/plugin/support-core/images/24x24/support.png";
+        return shouldDisplay() ? "/plugin/support-core/images/24x24/support.png" : null;
+    }
+
+    boolean shouldDisplay() {
+        return Jenkins.get().hasPermission(Jenkins.ADMINISTER);
     }
 
     @DataBoundSetter
--- a/src/main/java/com/cloudbees/jenkins/support/actions/SupportComputerAction.java
+++ b/src/main/java/com/cloudbees/jenkins/support/actions/SupportComputerAction.java
@@ -23,7 +23,7 @@ public class SupportComputerAction extends SupportObjectAction<Computer> {
 
     @Override
     public String getDisplayName() {
-        return Messages.SupportComputerAction_DisplayName();
+        return shouldDisplay() ? Messages.SupportComputerAction_DisplayName() : null;
     }
 
     @Extension
--- a/src/main/java/com/cloudbees/jenkins/support/actions/SupportRunAction.java
+++ b/src/main/java/com/cloudbees/jenkins/support/actions/SupportRunAction.java
@@ -22,7 +22,7 @@ public class SupportRunAction extends SupportObjectAction<Run> {
 
     @Override
     public String getDisplayName() {
-        return Messages.SupportRunAction_DisplayName();
+        return shouldDisplay() ? Messages.SupportRunAction_DisplayName() : null;
     }
 
     @Extension


src/main/java/com/cloudbees/jenkins/support/actions/SupportObjectAction.java
```diff
@@ -59,12 +59,16 @@ public final T getObject() {
...
     }
 
     @Override
     public String getIconFileName() {
-        return "/plugin/support-core/images/24x24/support.png";
+        return shouldDisplay() ? "/plugin/support-core/images/24x24/support.png" : null;
+    }
+
+    boolean shouldDisplay() {
+        return Jenkins.get().hasPermission(Jenkins.ADMINISTER);
```
aheritier, line 71: Shouldn't you use `SupportAction.CREATE_BUNDLE` instead ?

https://github.com/jenkinsci/support-core-plugin/blob/master/src/main/java/com/cloudbees/jenkins/support/SupportAction.java#L71

PierreBtz: I reused the same as https://github.com/jenkinsci/support-core-plugin/pull/220/files/747dc40b634af540503515b01efc4ca8a2ded2c0#diff-018a199824d332a24f6b86f5c65bdf37R127

varyvol: This is pointing to your same commit?

src/main/java/com/cloudbees/jenkins/support/actions/SupportAbstractItemAction.java
```diff
@@ -22,7 +22,7 @@ public SupportAbstractItemAction(AbstractItem target) {
 
     @Override
     public String getDisplayName() {
-        return Messages.SupportItemAction_DisplayName();
+        return shouldDisplay() ? Messages.SupportItemAction_DisplayName() : null;
```
varyvol, line 25: Wouldn't it be possible to implement methods in the super class to avoid having to change every method in the subclasses?

src/main/java/com/cloudbees/jenkins/support/actions/SupportObjectAction.java
PierreBtz: Yeah sorry link was bad, I meant: https://github.com/jenkinsci/support-core-plugin/blob/747dc40b634af540503515b01efc4ca8a2ded2c0/src/main/java/com/cloudbees/jenkins/support/actions/SupportObjectAction.java#L127

This is an interesting point, because `SupportAction.CREATE_BUNDLE` defines a permission (Download Bundle) but the action I linked checks for ADMINISTER :)

I managed to reproduce the issue:
* have a user with Overall/Read + Job/read + Download Permission
* go to a job page
* click on generate bundle (for the job)
* you won't be allowed to generate the bundle

Question is: is it by design, ie only an admin can retrieve a build directory? Or is it a bug?
If it's a bug, happy to fix it while fixing the action visibility issue.

src/main/java/com/cloudbees/jenkins/support/actions/SupportAbstractItemAction.java
PierreBtz: Functionally what you suggest is possible (all action display names are the same) but they are in fact referring to a different message, eg:

https://github.com/jenkinsci/support-core-plugin/blob/09dbc25ed1f15000477fc62fd6b730085ca70049/src/main/resources/com/cloudbees/jenkins/support/Messages.properties#L25

https://github.com/jenkinsci/support-core-plugin/blob/4a3c12403766f8e28386df1d2bd9757e3ec2fae7/src/main/resources/com/cloudbees/jenkins/support/actions/Messages.properties#L25

I suppose I could build the key to retrieve the message from the class name, but maybe it's worth considering removing the duplication of messages instead. AFAIU, the name will always be Support and the icon will always be the same too. Wdyt?

varyvol: If the user result is the same, +1 for removing the duplication.

src/main/java/com/cloudbees/jenkins/support/actions/SupportObjectAction.java
varyvol: Maybe @Dohbedoh can better illustrate us.

src/main/java/com/cloudbees/jenkins/support/actions/SupportAbstractItemAction.java
duemir: Messages are made unique by #239

Dohbedoh:
@PierreBtz I think that the mistake is in jelly / stapler at:
* https://github.com/jenkinsci/support-core-plugin/blob/support-core-2.70/src/main/resources/com/cloudbees/jenkins/support/actions/SupportRunAction/action.jelly#L5
* https://github.com/jenkinsci/support-core-plugin/blob/support-core-2.70/src/main/resources/com/cloudbees/jenkins/support/actions/SupportComputerAction/action.jelly#L5
* https://github.com/jenkinsci/support-core-plugin/blob/support-core-2.70/src/main/resources/com/cloudbees/jenkins/support/actions/SupportAbstractItemAction/action.jelly#L5

That does not work and should be `${app.ADMINISTER}`.

Dohbedoh:
@PierreBtz I proposed https://github.com/jenkinsci/support-core-plugin/pull/252

PierreBtz:
yeah, my bad for not following up on this...

PR #221, opened 2020-05-15
[JENKINS-62297] Collect User counts

Dohbedoh:
[JENKINS-62297](https://issues.jenkins-ci.org/browse/JENKINS-62297) Add a component that provide User count to the support bundle.

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/UsersContent.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/UsersContent.java
@@ -1,0 +1,90 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2019, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package com.cloudbees.jenkins.support.impl;
+
+import com.cloudbees.jenkins.support.api.Container;
+import com.cloudbees.jenkins.support.api.PrintedContent;
+import com.cloudbees.jenkins.support.timer.UnfilteredFileListCapComponent;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import hudson.Extension;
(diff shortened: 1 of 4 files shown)


src/main/java/com/cloudbees/jenkins/support/impl/UsersContent.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+
+@Extension
+public class UsersContent extends UnfilteredFileListCapComponent {
+
+    @NonNull
+    @Override
+    public Set<Permission> getRequiredPermissions() {
+        return Collections.emptySet();
```
varyvol, line 48: This is my only doubt. Even if we are only displaying a number, is it something we want to do without checking permissions? 

@daniel-beck maybe you can provide your opinion.

src/main/java/com/cloudbees/jenkins/support/impl/UsersContent.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    @NonNull
+    @Override
+    public String getDisplayName() {
+        return "Users Count";
```
rsandell, line 54: A bit of strange English here. Should either be "User count" or "Number of users" if you want to use the plural form.

src/main/java/com/cloudbees/jenkins/support/impl/UsersContent.java
```diff
@@ -0,0 +1,90 @@
...
+import jenkins.security.LastGrantedAuthoritiesProperty;
+
+import java.io.PrintWriter;
+import java.util.Collections;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+@Extension
+public class UsersContent extends UnfilteredFileListCapComponent {
```
rsandell, line 43:
```suggestion
public class UserCount extends UnfilteredFileListCapComponent {
```

src/main/java/com/cloudbees/jenkins/support/impl/UsersContent.java
Dohbedoh: Agreed. I think this needs Administer permissions. I'll let Daniel confirm.

src/main/java/com/cloudbees/jenkins/support/impl/UsersContent.java
```diff
@@ -0,0 +1,90 @@
...
+
+                User.getAll().stream()
+                    .collect(
+                        Collectors.groupingBy(
+                            (user) -> Optional.ofNullable(user.getProperty(LastGrantedAuthoritiesProperty.class)).isPresent(),
+                            Collectors.counting()))
+                        .forEach((lastGrantedAuthoritiesProperty, aLong) ->
+                            out.println(" * "
+                                    + (lastGrantedAuthoritiesProperty ? "Authenticated" : "Non Authenticated")
+                                    + " Users count: " + aLong)
```
daniel-beck, line 75:
```suggestion
                                    + " Users Count: " + aLong)
```

The rest of this is title cased.

src/main/java/com/cloudbees/jenkins/support/impl/UsersContent.java
```diff
@@ -0,0 +1,90 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2019, CloudBees, Inc.
```
daniel-beck, line 4:
```suggestion
 * Copyright (c) 2020, CloudBees, Inc.
```

src/main/java/com/cloudbees/jenkins/support/impl/UsersContent.java
daniel-beck: It is a _Component_ adding a _Content_ to a bundle, so the current name seems misleading. (I would argue that `ItemsContent` has a misleading name too.)

src/main/java/com/cloudbees/jenkins/support/impl/UsersContent.java
varyvol: It should actually "User Count" to match the component name.

daniel-beck: The point here is the uppercasing.

varyvol: Yes, I know, I was just adding something else to your proposal.

daniel-beck:
```suggestion
                                    + " User Count: " + aLong)
```

To make it official 😄

PR #223, opened 2020-06-17
Upgrade parent pom to 4.2

alecharp:
This removes some dependencies from the hpi binaries, which shouldn't have been there in the first place

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,1 +1,5 @@
-buildPlugin(configurations: buildPlugin.recommendedConfigurations())
+buildPlugin(configurations: [
+    [ platform: "linux", jdk: "8", jenkins: null ],
+    [ platform: "linux", jdk: "11", jenkins: null, javaLevel: "8" ],
+    [ platform: "windows", jdk: "8", jenkins: null, javaLevel: "8" ],
+])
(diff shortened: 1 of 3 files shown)


MRamonLeon:
We may specify the Jenkins versions to use in the CI:

```
Caused by: java.io.IOException: Script Security Plugin version 1.73 failed to load.
 - You must update Jenkins from version 2.164.1 to version 2.176.4 or later to run this plugin.
```

alecharp:
This is because the Java 11 build branch is not correctly seeing that the new minimal Jenkins version is greater than the minimal version supporting Java 11. So, it shouldn't try to build with `2.164.x` but keep the version set on the `pom.xml`.

Either we change https://github.com/jenkins-infra/pipeline-library/blob/master/vars/buildPlugin.groovy#L269 or we add some condition to not override the `jenkins.version` of the project if it's newer than the one set in the lib.

MRamonLeon:
We ca specify the versions to run on in the https://github.com/jenkinsci/support-core-plugin/blob/master/Jenkinsfile to get it works

alecharp:
@MRamonLeon it won't be used in this PR as I don't have any permission on this plugin, so Jenkins will not use any modification I could do on `Jenkinsfile`.

varyvol:
@alecharp if you push the changes, I can create a branch from your changes and test it.

alecharp:
Normally, this should work @varyvol

varyvol:
#224

varyvol:
Merging given #224 worked.

PR #225, opened 2020-06-29
[JENKINS-62844] Update the link to the bundle anonymization setting

PierreBtz:


varyvol:
@PierreBtz I'm not sure I understand the relationshio between the title and the actual changes?

PierreBtz:
@varyvol there was none it was the commit message for another issue... Fixed

aheritier:
@MRamonLeon 

When you use

```
    /**
     * 
     * @return the global configuration category for CasC where this config lands
     */
    @Override
    public @Nonnull GlobalConfigurationCategory getCategory() {
        return GlobalConfigurationCategory.get(GlobalConfigurationCategory.Security.class);
    }
```

In ContentFilters you are moving it under the global security panel and the side-effect is to move the CASC entry under security.*

MRamonLeon:
> @MRamonLeon
> 
> When you use
> 
> ```
>     /**
>      * 
>      * @return the global configuration category for CasC where this config lands
>      */
>     @Override
>     public @Nonnull GlobalConfigurationCategory getCategory() {
>         return GlobalConfigurationCategory.get(GlobalConfigurationCategory.Security.class);
>     }
> ```
> 
> In ContentFilters you are moving it under the global security panel and the side-effect is to move the CASC entry under security.*

@aheritier I would say the opposite. We wanted to put it under this category on CasC which is the purpose of this method and we got the side-effect of having the url changed. Not expected from my PoV.

PR #227, opened 2020-08-27
[JENKINS-62925] Rename records of slave to agent

Dohbedoh:
[JENKINS-62925](https://issues.jenkins-ci.org/browse/JENKINS-62925): renaming "slave" to "agent" in the data produced, comments, class and variables names.

The more "breaking" changes are:

1) renaming of public component classes as those names may be used by clients such as the Jenkins CLI to generate a bundle:

* `SlaveCommandStatistics` to `AgentCommandStatistics`
* `SlaveLaunchLogs` to `AgentLaunchLogs`
* `SlaveLogs` to `AgentLogs`

2) renaming of directories in the bundle:

* `nodes/slave/` becomes `nodes/agent/`

3) In the `nodes.md` details:

* `Slave Version` becomes `Agent Version`
* we retrieve the digest or `agent.jar` and not `slave.jar` (we require `2.204.6` as of today so I think this is safe)

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
@@ -84,11 +84,11 @@ public class AboutJenkins extends Component {
 
     private static final Logger logger = Logger.getLogger(AboutJenkins.class.getName());
 
-    private final WeakHashMap<Node,String> slaveVersionCache = new WeakHashMap<Node, String>();
+    private final WeakHashMap<Node,String> agentVersionCache = new WeakHashMap<Node, String>();
 
     private final WeakHashMap<Node,String> javaInfoCache = new WeakHashMap<Node, String>();
 
-    private final WeakHashMap<Node,String> slaveDigestCache = new WeakHashMap<Node, String>();
+    private final WeakHashMap<Node,String> agentDigestCache = new WeakHashMap<Node, String>();
 
     @NonNull
     @Override
@@ -146,11 +146,11 @@ public class AboutJenkins extends Component {
         out.println("      - 99th percentile:    " + snapshot.get99thPercentile());
     }
 
-    private static final class GetSlaveDigest extends MasterToSlaveCallable<String, RuntimeException> {
+    private static final class GetAgentDigest extends MasterToSlaveCallable<String, RuntimeException> {
         private static final long serialVersionUID = 1L;
         private final String rootPathName;
 
-        public GetSlaveDigest(FilePath rootPath) {
+        public GetAgentDigest(FilePath rootPath) {
             this.rootPathName = rootPath.getRemote();
         }
 
@@ -172,7 +172,7 @@ public class AboutJenkins extends Component {
         }
     }
 
-    private static class GetSlaveVersion extends MasterToSlaveCallable<String, RuntimeException> {
+    private static class GetAgentVersion extends MasterToSlaveCallable<String, RuntimeException> {
         private static final long serialVersionUID = 1L;
 
         @SuppressFBWarnings(
@@ -639,7 +639,7 @@ public class AboutJenkins extends Component {
                     Markdown.escapeBacktick(ContentFilter.filter(filter, jenkins.getRootDir().getAbsolutePath())) + "`");
             out.println("      - Labels:         " + ContentFilter.filter(filter, getLabelString(jenkins)));
             out.println("      - Usage:          `" + jenkins.getMode() + "`");
-            out.println("      - Slave Version:  " + Launcher.VERSION);
+            out.println("      - Agent Version:  " + Launcher.VERSION);
             out.print(new GetJavaInfo("      -", "          +").getInfo(filter));
             out.println();
             for (Node node : jenkins.getNodes()) {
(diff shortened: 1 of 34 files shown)


varyvol:
@Dohbedoh A bit unsure about this. Obviously in favor of renaming, but I believe we should try something not so disturbing regarding possible breaking changes. Maybe something as simple as deprecating and give some time to people to execute the necessary changes.

Dohbedoh:
Agreed. I will take a look into what's possible for the Component classes.
What about the renaming of the zip entry `nodes/agents` and the markdown "Agent Version" ? Should we keep them and mark them as deprecated as well ? (maybe for the zip entry, a symlink for `nodes/slaves -> nodes/agents`)

varyvol:
I would vote for maintaining everything for some time, and then removing the whole thing at once. Symlinks might be problematic, so better doubling them, and maybe for the markdown adding both values?

Dohbedoh:
@varyvol I am bit worried about duplicating the `nodes/slaves` in `nodes/agents`. This directory may contain a lot of files and I have seen environment where it accounts for most of the content of the support bundle.

varyvol:
@Dohbedoh yes, might be. Let's go with the symlinks option then?

Dohbedoh:
I will try that asap.

aheritier:
If `slaves` folders are renamed we should rename `master` folder too but we should anticipate and update all tooling using them

@Dohbedoh from my POV in very short term we should just rename the terms visible in the UI

Dohbedoh:
I have rebased this.

@aheritier so we don't touch what is generated in the bundle yet then ?

aheritier:
@Dohbedoh for now it's probably better. we have to update our support tools to be able to handle both names for folders (master|controller) and (agents|slaves) - and for any other change we might do.

Dohbedoh:
I reverted the renaming of public Classes and the renaming of fields and directories in the bundle.

src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
```diff
@@ -739,7 +739,7 @@ private String getLabelString(Node n) {
                     }
                 }
                 for (String resourcePath : Arrays.asList(
-                        "/WEB-INF/slave.jar", // note that as of 2.33 this will not be present (anyway it is the same as war/WEB-INF/lib/remoting-*.jar, printed above)
+                        "/WEB-INF/agent.jar", // note that as of 2.33 this will not be present (anyway it is the same as war/WEB-INF/lib/remoting-*.jar, printed above)
```
aheritier, line 742: Not sure what we are talking about here @Dohbedoh 
This is/was a jar within jenkins WAR ?
Maybe we can just remove the line if the minimum version of jenkins the plugin support do not find any of them?

src/test/java/com/cloudbees/jenkins/support/CheckFilterTest.java
```diff
@@ -17,6 +17,7 @@
 import com.cloudbees.jenkins.support.impl.SystemProperties;
 import com.cloudbees.jenkins.support.impl.ThreadDumps;
 import com.cloudbees.jenkins.support.impl.UpdateCenter;
+import hudson.EnvVars;
```
aheritier, line 20: ??

aheritier (approved):
Few comments but globally good to me @Dohbedoh 
Your PR also makes me think that our tools (advisor) looking for logs with these words are probably broken if they were updated in Jenkins code (I think we already have a task about this)

src/test/java/com/cloudbees/jenkins/support/SupportActionTest.java
```diff
@@ -101,10 +101,10 @@ public void generateAllBundles() throws IOException, SAXException {
     }
 
     @Test
-    @Issue("JENKINS-63722")
+    @Issue({"JENKINS-63722","JENKINS-62925"})
     public void generateAllBundlesBackwardCompatibility() throws Exception {
         Assume.assumeTrue(!Functions.isWindows());
-        Assume.assumeTrue(SystemPlatform.LINUX == SystemPlatform.current());
+//        Assume.assumeTrue(SystemPlatform.LINUX == SystemPlatform.current());
```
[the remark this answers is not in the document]
Dohbedoh: some changes I did not revert properly. Thanks for spotting this!

src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
Dohbedoh: yeah we can remove this and `remoting.jar` I think: https://github.com/jenkinsci/jenkins/pull/2633/files

aHenryJard:
Sorry to jump on this PR, you are covering a lots thanks ! @Dohbedoh Do you think it's possible to change this 2 lines too ?

https://github.com/jenkinsci/support-core-plugin/blob/27acdf1c60d2bb461d17317c0f059a2662c2f096/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java#L642

https://github.com/jenkinsci/support-core-plugin/blob/27acdf1c60d2bb461d17317c0f059a2662c2f096/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java#L634

Without breaking ?

aheritier:
it is possible @aHenryJard but sadly it will break our internal parsers. we have to update them first :/

Dohbedoh:
@aHenryJard what is the new name of the controller built in node ? Is it `blub` ? https://github.com/jenkinsci/jenkins/blob/6c1c7fb26bf848143a2c71662eeea96fd5f00197/core/src/main/resources/jenkins/model/BlubMigration/message.properties#L24

aHenryJard:
> @aHenryJard what is the new name of the controller built in node ? Is it `blub` ? https://github.com/jenkinsci/jenkins/blob/6c1c7fb26bf848143a2c71662eeea96fd5f00197/core/src/main/resources/jenkins/model/BlubMigration/message.properties#L24

@Dohbedoh  It's **_not decided yet_**, ongoing discussion here (feel free to participate too :) ) https://groups.google.com/u/1/g/jenkinsci-dev/c/x5vdlJDvntw
The PR you mention, is there to re-open discussion and move forward.

aHenryJard:
@Dohbedoh decision is made, Master node => "Built-in Node" as per https://groups.google.com/g/jenkinsci-dev/c/x5vdlJDvntw/m/o5OGmRrpAAAJ

Dohbedoh:
They would definitely impact some our tools.

@aHenryJard Can we do the changes you suggested as part of next iterations for breaking changes ?

aHenryJard:
> They would definitely impact some our tools.
> 
> @aHenryJard Can we do the changes you suggested as part of next iterations for breaking changes ?

Yes sure the first iteration is end user facing texts, with no code change and no breaking changes.

Dohbedoh:
thanks! @MRamonLeon Are we good to merge this ?

PR #228, opened 2020-08-27
[JENKINS-62746] Add remoting version details to about.md

Dohbedoh:
[JENKINS-62746](https://issues.jenkins-ci.org/browse/JENKINS-62746) Adding Master Remoting version details to the bundle.

cc @aheritier

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
@@ -36,6 +36,7 @@ import jenkins.model.Jenkins;
 import jenkins.model.JenkinsLocationConfiguration;
 import jenkins.model.identity.IdentityRootAction;
 import jenkins.security.MasterToSlaveCallable;
+import jenkins.slaves.RemotingVersionInfo;
 import org.apache.commons.io.FileUtils;
 import org.kohsuke.stapler.Stapler;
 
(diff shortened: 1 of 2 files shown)


aheritier (approved):
LGTM. In nodes.md we also have the agent version ( https://github.com/jenkinsci/support-core-plugin/blob/3cf5a71d726501ba59b478ab5b1c43ec8d5bebae/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java#L664 ) for every agent. Do you think we should add the same in the master definition ( https://github.com/jenkinsci/support-core-plugin/blob/3cf5a71d726501ba59b478ab5b1c43ec8d5bebae/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java#L626 ) ?
WDYT ?

varyvol (approved):

Dohbedoh:
> Do you think we should add the same in the master definition

I thought it might be redundant to put it in the `nodes.md` too. The master node in `nodes.md` shows the launcher version: https://github.com/jenkinsci/support-core-plugin/pull/228/files#diff-f2c7ab60306f05af2fab348ce75750cfR640.

aheritier:
I agree @Dohbedoh 
That's also why I would prefer to have one day a data model properly serialised that we could read
it's just confusing to have some info in different files (sometimes the same) and probably available from different options (I don't remember if nodes.md can be enabled/disabled with a different option than about.md)

PR #231, opened 2020-09-18
[JENKINS-63674] Capture full thread-dumps when there is a high CPU usage

fbelzunc:
Dear maintainer of the support-core plugin,

the following implementation intends to capture thread dumps when the CPU consumption is very high. Under my experience as Jenkins Support Engineer, slow-requests are not enough to troubleshoot performance issues and I am intending with this implementation to capture the full picture of the JVM when the metrics plugin captures a very high usage of the CPU.

I am concerned about two things:

* Is to look at the metrics plugin the right approach? I think so because I usually do the same through Groovy Scripting with success.
* `slow-requests` are currently stored under `$JENKINS_HOME/slow-requests`, so I am doing something similar here using `$JENKINS_HOME/high-load/cpu`. However, tomorrow I might want to capture more things... so I am wondering if everything should go under a folder called support-data-collection or something like this.

CC @Dohbedoh

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/threaddump/HighLoadChecker.java
+++ b/src/main/java/com/cloudbees/jenkins/support/threaddump/HighLoadChecker.java
@@ -1,0 +1,103 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2020 CloudBees, Inc
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.cloudbees.jenkins.support.threaddump;
+
+import com.cloudbees.jenkins.support.impl.ThreadDumps;
+import com.cloudbees.jenkins.support.timer.FileListCap;
+import com.codahale.metrics.Gauge;
+import com.codahale.metrics.MetricSet;
+import hudson.Extension;
+import hudson.ExtensionList;
+import hudson.model.PeriodicWork;
+import jenkins.metrics.api.MetricProvider;
+import jenkins.metrics.impl.VMMetricProviderImpl;
+import jenkins.model.Jenkins;
+
+import java.io.File;
+import java.io.FileOutputStream;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.TimeZone;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * PeriodicWork to check when there is a high load in the instance.
+ *
+ * Only checking high CPU usage for the moment, but it can be used
+ * to generate thread dumps in high heap memory consumption.
+ */
+@Extension
+public class HighLoadChecker extends PeriodicWork {
+
+    /**
+     * Recurrence period to check high load consumption and take the corresponded actions
+     * like generating a thread dump
+     */
+    public static final int RECURRENCE_PERIOD_SEC =
+            Integer.getInteger(HighLoadChecker.class.getName()+ ".RECURRENCE_PERIOD_SEC", 900);
+
+    /**
+     * This is the CPU usage threshold
+     */
+    public static final Double CPU_USAGE_THRESHOLD =
+            new Double(System.getProperty(HighLoadChecker.class.getName() + ".CPU_USAGE_THRESHOLD", "0.95"));
+
+    /**
+     * Limit the number of thread dumps to retain on high cpu
+     */
+    public static final int HIGH_CPU_THEAD_DUMPS_TO_RETAIN =
+            Integer.getInteger(HighLoadChecker.class.getName()+ ".HIGH_CPU_THEAD_DUMPS_TO_RETAIN", 10);
+
+    /**
+     * Thread dumps generated on high CPU load are stored in $JENKINS_HOME/high-load/cpu
+     **/
+    final FileListCap logs = new FileListCap(new File(Jenkins.getInstanceOrNull().getRootDir(),"high-load/cpu"), HIGH_CPU_THEAD_DUMPS_TO_RETAIN);
+
+    final SimpleDateFormat format = new SimpleDateFormat("yyyyMMdd-HHmmss.SSS");
+    {
+        format.setTimeZone(TimeZone.getTimeZone("UTC"));
+    }
+
+    @Override
+    public long getRecurrencePeriod() {
+        return TimeUnit.SECONDS.toMillis(RECURRENCE_PERIOD_SEC);
+    }
+
+    @Override
+    protected void doRun() throws Exception {
+        VMMetricProviderImpl vmMetricProvider = ExtensionList.lookup(MetricProvider.class).get(VMMetricProviderImpl.class);
+        MetricSet vmProviderMetricSet = vmMetricProvider.getMetricSet();
+        Gauge cpu = (Gauge) vmProviderMetricSet.getMetrics().get("vm.cpu.load");
+        Double cpuLoad = new Double(cpu.getValue().toString());
+
+        if (Double.compare(CPU_USAGE_THRESHOLD, cpuLoad) < 0) {
+            File threadDumpFile = logs.file(format.format(new Date()) + ".txt");
+            logs.add(threadDumpFile);
+            FileOutputStream fileOutputStream = new FileOutputStream(threadDumpFile);
+            ThreadDumps.threadDump(fileOutputStream);
+            fileOutputStream.close();
+        }
+    }
+}
--- a/src/main/java/com/cloudbees/jenkins/support/threaddump/HighLoadChecker.java
+++ b/src/main/java/com/cloudbees/jenkins/support/threaddump/HighLoadChecker.java
@@ -98,16 +98,16 @@ public class HighLoadChecker extends PeriodicWork {
             Gauge cpu = (Gauge) vmProviderMetricSet.getMetrics().get("vm.cpu.load");
             cpuLoad = new Double(cpu.getValue().toString());
         } catch (NullPointerException nullPointerException) {
-            LOGGER.log(WARNING, "Metrics plugin does not seem to be available", nullPointerException);
+            LOGGER.log(WARNING, "Support Core plugin can't generate thread dumps. Metrics plugin does not seem to be available", nullPointerException);
             return;
         }
 
         if (Double.compare(Runtime.getRuntime().availableProcessors() * CPU_USAGE_THRESHOLD, cpuLoad) < 0) {
             File threadDumpFile = logs.file(format.format(new Date()) + ".txt");
-            FileOutputStream fileOutputStream = new FileOutputStream(threadDumpFile);
-            fileOutputStream.close();
-            ThreadDumps.threadDump(fileOutputStream);
-            logs.add(threadDumpFile);
+            try (FileOutputStream fileOutputStream = new FileOutputStream(threadDumpFile)){
+                ThreadDumps.threadDump(fileOutputStream);
+                logs.add(threadDumpFile);
+            }
         }
     }
 
--- a/src/main/java/com/cloudbees/jenkins/support/threaddump/HighLoadChecker.java
+++ b/src/main/java/com/cloudbees/jenkins/support/threaddump/HighLoadChecker.java
@@ -41,6 +41,9 @@ import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.TimeZone;
 import java.util.concurrent.TimeUnit;
+import java.util.logging.Logger;
+
+import static java.util.logging.Level.WARNING;
 
 /**
  * PeriodicWork to check when there is a high load in the instance.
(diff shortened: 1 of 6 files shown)


Dohbedoh:
@fbelzunc I am a bit worried about the 0.95. This is not a percentage (well it would be, if the equivalent of exactly **1** CPU core is available to the JVM). Per the Metrics plugin:

```
The rate of CPU time usage by the JVM per unit time on the Jenkins master. This is equivalent to the number of CPU cores being used by the Jenkins master JVM.
```

In a container for example, you may want to set CPU quota / period to give an amount of 2 allocatable CPU in which case we probably want something like 0.95 * 2 (in general 95% of what the JVM is "*allocated*").

It can be argued whether this is a good default or not. I think that what we want is something more aware of the environment, such as `runtime.availableProcessors()*0.95` ?

src/main/java/com/cloudbees/jenkins/support/threaddump/HighLoadChecker.java
```diff
@@ -0,0 +1,103 @@
...
+    /**
+     * Limit the number of thread dumps to retain on high cpu
+     */
+    public static final int HIGH_CPU_THEAD_DUMPS_TO_RETAIN =
+            Integer.getInteger(HighLoadChecker.class.getName()+ ".HIGH_CPU_THEAD_DUMPS_TO_RETAIN", 10);
+
+    /**
+     * Thread dumps generated on high CPU load are stored in $JENKINS_HOME/high-load/cpu
+     **/
+    final FileListCap logs = new FileListCap(new File(Jenkins.getInstanceOrNull().getRootDir(),"high-load/cpu"), HIGH_CPU_THEAD_DUMPS_TO_RETAIN);
```
varyvol, line 76:
```suggestion
    final FileListCap logs = new FileListCap(new File(Jenkins.getInstance().getRootDir(),"high-load/cpu"), HIGH_CPU_THEAD_DUMPS_TO_RETAIN);
```

src/main/java/com/cloudbees/jenkins/support/threaddump/HighLoadChecker.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    @Override
+    public long getRecurrencePeriod() {
+        return TimeUnit.SECONDS.toMillis(RECURRENCE_PERIOD_SEC);
+    }
+
+    @Override
+    protected void doRun() throws Exception {
+        VMMetricProviderImpl vmMetricProvider = ExtensionList.lookup(MetricProvider.class).get(VMMetricProviderImpl.class);
```
varyvol, line 90: Is this provider always going to be there? (I'm not sure metrics can be deactivated or something like that.

src/main/java/com/cloudbees/jenkins/support/threaddump/HighLoadChecker.java
```diff
@@ -0,0 +1,103 @@
...
+        MetricSet vmProviderMetricSet = vmMetricProvider.getMetricSet();
+        Gauge cpu = (Gauge) vmProviderMetricSet.getMetrics().get("vm.cpu.load");
+        Double cpuLoad = new Double(cpu.getValue().toString());
+
+        if (Double.compare(CPU_USAGE_THRESHOLD, cpuLoad) < 0) {
```
varyvol, line 95: Wouldn't this be the way around? If cpuLoad is bigger than the threshold?

src/main/java/com/cloudbees/jenkins/support/threaddump/HighLoadChecker.java
```diff
@@ -0,0 +1,103 @@
...
+            File threadDumpFile = logs.file(format.format(new Date()) + ".txt");
+            logs.add(threadDumpFile);
```
varyvol, line 97: Maybe this should be at the end, so that we don't save files that cannot be correctly generated?

src/main/java/com/cloudbees/jenkins/support/threaddump/HighLoadChecker.java
```diff
@@ -0,0 +1,103 @@
...
+            FileOutputStream fileOutputStream = new FileOutputStream(threadDumpFile);
```
varyvol, line 98: try with resources?

varyvol (commented):
Some comments

src/main/java/com/cloudbees/jenkins/support/threaddump/HighLoadChecker.java
```diff
@@ -0,0 +1,116 @@
...
+            VMMetricProviderImpl vmMetricProvider = ExtensionList.lookup(MetricProvider.class).get(VMMetricProviderImpl.class);
+            MetricSet vmProviderMetricSet = vmMetricProvider.getMetricSet();
+            Gauge cpu = (Gauge) vmProviderMetricSet.getMetrics().get("vm.cpu.load");
+            cpuLoad = new Double(cpu.getValue().toString());
+        } catch (NullPointerException nullPointerException) {
+            LOGGER.log(WARNING, "Support Core plugin can't generate thread dumps. Metrics plugin does not seem to be available", nullPointerException);
+            return;
+        }
+
+        if (Double.compare(Runtime.getRuntime().availableProcessors() * CPU_USAGE_THRESHOLD, cpuLoad) < 0) {
```
fbelzunc, line 105: @varyvol AFAIK this is correct, doesn't it? If `cpuLoad > Runtime.getRuntime().availableProcessors() * CPU_USAGE_THRESHOLD`, then  it will return 

```
     * @param   d1        the first {@code double} to compare
     * @param   d2        the second {@code double} to compare
     * @return  the value {@code 0} if {@code d1} is
     *          numerically equal to {@code d2}; a value less than
     *          {@code 0} if {@code d1} is numerically less than
     *          {@code d2}; and a value greater than {@code 0}
     *          if {@code d1} is numerically greater than
     *          {@code d2}.
     * @since 1.4
     */
```

src/main/java/com/cloudbees/jenkins/support/threaddump/ThreadDumpHighCPU.java
```diff
@@ -0,0 +1,50 @@
...
+    @Inject
+    HighLoadChecker checker;
+
+    public String getDisplayName() {
+        return Messages.ThreadDumpHighCPU_DisplayName();
+    }
+
+    @Override
+    public void addContents(@NonNull Container container) {
+        super.addContents(container, checker.logs);
```
varyvol, line 48: It the logs are empty, what should happen?

src/main/java/com/cloudbees/jenkins/support/threaddump/HighLoadChecker.java
varyvol: Yes, I believe it's valid but it's really confusing to me. At least my thinking is: I've calculated the cpuLoad, great! Now I'm going to compare it to the limit. Is it bigger? Then I do whatever.

But not against it if you think it's fine.

Dohbedoh: Looks correct to me. 
Another thing I want to point out is that if the cpuLoad is `Double#NaN`, then it will satisfy this condition and generate a thread dump and we may not want this. I think this is a possibility and maybe we should check that the `cpuLoad` is not `NaN` to avoid false positive ?

fbelzunc:
@varyvol @Dohbedoh  I modified a bit the code in order to be less aggressive in terms of the default settings.

By default, I am ONLY taking a thread dump in case there are 3 consecutive High CPU (>80% of average) during 30 minutes - so in the worst case it is just 1 thread dump each 30 minutes.

What do you think? I just test it these bits and I am happy with the way it currently works.

WDYT?

src/main/java/com/cloudbees/jenkins/support/threaddump/HighLoadCpuChecker.java
```diff
@@ -0,0 +1,128 @@
...
+            Gauge cpu = (Gauge) vmProviderMetricSet.getMetrics().get("vm.cpu.load");
+            cpuLoad = new Double(cpu.getValue().toString());
+        } catch (NullPointerException nullPointerException) {
+            LOGGER.log(WARNING, "Support Core plugin can't generate thread dumps. Metrics plugin does not seem to be available", nullPointerException);
+            return;
+        }
+        if (Double.compare(Runtime.getRuntime().availableProcessors() * CPU_USAGE_THRESHOLD, cpuLoad) < 0) {
+            countConsequtivePositives++;
+            if (countConsequtivePositives >= HIGH_CPU_CONSECUTIVE_TIMES) {
+                    countConsequtivePositives = 0;
```
varyvol, line 115: Note this would happen at the 4th time, I think.

fbelzunc: @varyvol Why do you say that? I am using `>= HIGH_CPU_CONSECUTIVE_TIMES `. I just debugged it and it happens at the 3rd time. Am I missing something?

varyvol: Ah right, because you are increasing it before asking. Forget about my comment!

varyvol (approved):

src/main/java/com/cloudbees/jenkins/support/threaddump/HighLoadComponent.java
```diff
@@ -0,0 +1,52 @@
...
+import com.cloudbees.jenkins.support.timer.UnfilteredFileListCapComponent;
+import com.google.inject.Inject;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import hudson.Extension;
+
+/**
+ * ThreadDumpHighCPU component
+ */
+@Extension
+public class HighLoadComponent extends UnfilteredFileListCapComponent {
```
recampbell, line 37: I think the thread dump needs to be filtered because it may contain job names.

src/main/java/com/cloudbees/jenkins/support/threaddump/HighLoadCpuChecker.java
```diff
@@ -0,0 +1,128 @@
...
+            VMMetricProviderImpl vmMetricProvider = ExtensionList.lookup(MetricProvider.class).get(VMMetricProviderImpl.class);
+            MetricSet vmProviderMetricSet = vmMetricProvider.getMetricSet();
+            Gauge cpu = (Gauge) vmProviderMetricSet.getMetrics().get("vm.cpu.load");
+            cpuLoad = new Double(cpu.getValue().toString());
+        } catch (NullPointerException nullPointerException) {
+            LOGGER.log(WARNING, "Support Core plugin can't generate thread dumps. Metrics plugin does not seem to be available", nullPointerException);
+            return;
+        }
+        if (cpuLoad != null && Double.compare(Runtime.getRuntime().availableProcessors() * CPU_USAGE_THRESHOLD, cpuLoad) < 0) {
+            countConsequtivePositives++;
```
recampbell, line 113: "consecutive", not "consequtive"

src/main/java/com/cloudbees/jenkins/support/threaddump/HighLoadComponent.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Inject
```
recampbell, line 39: Suggest avoiding usage of @Inject and instead use static component lookup. In my experience, when it fails, it's incredibly hard to debug. Just my advice, not a hard block.

recampbell (changes requested):
Need to anonymize/filter the thread dumps I think, since they can contain confidential data.

src/main/java/com/cloudbees/jenkins/support/threaddump/HighLoadCpuChecker.java
```diff
@@ -0,0 +1,128 @@
...
+     * cores available
+     */
+    public static final Double CPU_USAGE_THRESHOLD =
+            new Double(System.getProperty(HighLoadCpuChecker.class.getName() + ".CPU_USAGE_THRESHOLD", "0.80"));
+
+    /**
+     * Limit the number of thread dumps to retain on high cpu
+     */
+    public static final int HIGH_CPU_THEAD_DUMPS_TO_RETAIN =
+            Integer.getInteger(HighLoadCpuChecker.class.getName()+ ".HIGH_CPU_THEAD_DUMPS_TO_RETAIN", 5);
```
Dohbedoh, line 81:
```suggestion
            Integer.getInteger(HighLoadCpuChecker.class.getName()+ ".HIGH_CPU_THREAD_DUMPS_TO_RETAIN", 5);
```

src/main/java/com/cloudbees/jenkins/support/threaddump/HighLoadCpuChecker.java
```diff
@@ -0,0 +1,128 @@
...
+     * This is the CPU usage threshold. Determinate de percentage of the total CPU used across all the
+     * cores available
+     */
+    public static final Double CPU_USAGE_THRESHOLD =
+            new Double(System.getProperty(HighLoadCpuChecker.class.getName() + ".CPU_USAGE_THRESHOLD", "0.80"));
+
+    /**
+     * Limit the number of thread dumps to retain on high cpu
+     */
+    public static final int HIGH_CPU_THEAD_DUMPS_TO_RETAIN =
```
Dohbedoh, line 80:
```suggestion
    public static final int HIGH_CPU_THREAD_DUMPS_TO_RETAIN =
```

src/main/java/com/cloudbees/jenkins/support/threaddump/HighLoadCpuChecker.java
```diff
@@ -0,0 +1,128 @@
...
+            Integer.getInteger(HighLoadCpuChecker.class.getName()+ ".HIGH_CPU_THEAD_DUMPS_TO_RETAIN", 5);
+
+    /**
+     * Thread dumps generated on high CPU load are stored in $JENKINS_HOME/high-load/cpu
+     **/
+    protected final FileListCap logs = new FileListCap(new File(Jenkins.getInstance().getRootDir(),"high-load/cpu"), HIGH_CPU_THEAD_DUMPS_TO_RETAIN);
```
Dohbedoh, line 86:
```suggestion
    protected final FileListCap logs = new FileListCap(new File(Jenkins.getInstance().getRootDir(),"high-load/cpu"), HIGH_CPU_THREAD_DUMPS_TO_RETAIN);
```

src/main/java/com/cloudbees/jenkins/support/threaddump/HighLoadCpuChecker.java
```diff
@@ -0,0 +1,128 @@
...
+     * in the RECURRENCE_PERIOD_SEC
+     */
+    public static final int RECURRENCE_PERIOD_SEC =
+            Integer.getInteger(HighLoadCpuChecker.class.getName()+ ".RECURRENCE_PERIOD_SEC", 600);
+
+    /**
+     * Consecutive high CPUs to take a thread dump
+     */
+    public static final int HIGH_CPU_CONSECUTIVE_TIMES =
+            Integer.getInteger(HighLoadCpuChecker.class.getName()+ ".HIGH_CPU_CONSECUTIVE_TIMES", 3);;
```
Dohbedoh, line 68:
```suggestion
            Integer.getInteger(HighLoadCpuChecker.class.getName()+ ".HIGH_CPU_CONSECUTIVE_TIMES", 3);
```

src/main/java/com/cloudbees/jenkins/support/threaddump/HighLoadComponent.java
Dohbedoh: @fbelzunc pointed out that it uses `ThreadDumps#threadDump`. That should take care of the filtering AFAIU, and we should not need to use a filtered content here or it would be redundant.

varyvol (approved):

Dohbedoh (approved):

src/main/java/com/cloudbees/jenkins/support/threaddump/HighLoadChecker.java
Dohbedoh: Metrics plugin is not an optional dependency to support core and that `VMMetricProviderImpl` is around for a long time. So I guess that so far it is acceptable to use the `VMMetricProviderImpl` right ?

PR #232, opened 2020-09-21
[JENKINS-50010] Add Global Configuration and Category

Dohbedoh:
Add the ability to configure the Automated Bundle Generation globally. This also introduce the first ManagementLink for Support Core. Therefore I implemented the `GlobalConfigurationCategory` solution proposed since 2.222.x to allow addition of GlobalConfiguration under that category in the future from this plugin or its extensions.

* A new Management Link "Support Core" under **Manage Jenkins**

support-core-category-link

* Under that management link, a section "Automated Support Bundle Generation" with the following configuration:
    * **Enable**: Whether to enable / disable automated bundle generation (default `true`)
    * **Period**: The recurrence period in hours. Possible values are integer between `1` and `24`, as per the [current system property](https://github.com/jenkinsci/support-core-plugin/blob/support-core-2.70/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java#L132-L137) (default `1`)
    * **Components**: Advanced layout to select the components for automated generation (default to the list of `selectedByDefault` components)

automated-bundle-generation-folded

* CasC compatibility. Example:

    ```
    support:
      automatedBundleConfiguration:
        componentIds:
          - "AboutBrowser"
          - "AboutJenkins"
          - "AboutUser"
        enabled: true
        period: 2
    ```
* Added documentation

Couple of notes:

* The System Property `com.cloudbees.jenkins.support.SupportPlugin.AUTO_BUNDLE_PERIOD_HOURS` is still supported. It **enforces** the enablement if set to 0 or less. Or enforces the period otherwise.
* The current automatic bundle generation captures all components. And I am proposing to change this to **only include the ones selected by default**. WDYT @varyvol ? If not part of this PR I really want to create a ticket to review the list of components to include in the background generation. We have been taking all of them, with a period of 1 hour since the beginning. But the list of components have grown since and this might not be the best strategy (especially in terms of performance vs. value).

cc @fbelzunc


src/main/java/com/cloudbees/jenkins/support/config/SupportAutomatedBundleConfiguration.java
```diff
@@ -0,0 +1,252 @@
...
+
+    @DataBoundSetter
+    public void setComponentIds(List<String> componentIds) {
+        this.componentIds = componentIds;
+    }
+
+    @DataBoundSetter
+    public void setEnabled(boolean enabled) {
+        if (isEnforcedDisabled()) {
+            this.enabled = SupportPlugin.AUTO_BUNDLE_PERIOD_HOURS >= 1;
```
varyvol, line 123: This is never going to happen because the if is checking that the period hours is <= 0.

src/main/resources/com/cloudbees/jenkins/support/config/SupportAutomatedBundleConfiguration/components.jelly
```diff
@@ -0,0 +1,26 @@
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" 
+         xmlns:f="/lib/form">
+    <j:forEach var="component" items="${instance.applicableComponents}">
+        <f:entry field="component">
+            <div name="components">
```
varyvol, line 6: Shouldn't this be in an upper level?

src/main/resources/com/cloudbees/jenkins/support/config/SupportAutomatedBundleConfiguration/help-period.html
```diff
@@ -0,0 +1,4 @@
+<div>
```
varyvol, line 1: I'm wondering if there should be some kind of message here clarifying how this works with the system property.

varyvol:
@Dohbedoh it sounds good to me that we revisit the list of selected components, but I prefer to do so on another PR.

pom.xml
```diff
@@ -71,7 +71,7 @@
   <properties>
     <revision>2.71</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.204.6</jenkins.version>
+    <jenkins.version>2.222.1</jenkins.version>
```
aheritier, line 74: We will have to take care to highlight the pre-requisite dump (For example I will have to do it in Advisor too)

src/main/java/com/cloudbees/jenkins/support/config/SupportAutomatedBundleConfiguration.java
Dohbedoh: The value 0 can happen. This is the current way that users have to disable the automatic generation. After double checking, values less than 0 cannot happen because of the `min` and `max` used when we initialize from the system property: https://github.com/jenkinsci/support-core-plugin/blob/support-core-2.70/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java#L132-L137.

src/main/resources/com/cloudbees/jenkins/support/config/SupportAutomatedBundleConfiguration/help-period.html
Dohbedoh: Good point! Will definitely add that.

src/main/resources/com/cloudbees/jenkins/support/config/SupportAutomatedBundleConfiguration/components.jelly
Dohbedoh: This is very similar to what we have in the `SupportAction`. I just wanted to split the "components" part in a jelly file.

Note: IMO we should create an improvement ticket about creating a reusable jelly component selector. And improve the UX as well.

src/main/java/com/cloudbees/jenkins/support/config/SupportAutomatedBundleConfiguration.java
varyvol: What I mean is: if "isEnforceDisabled()" return true, then the value is 0 and so this.enabled is always going to be false.

src/main/resources/com/cloudbees/jenkins/support/config/SupportAutomatedBundleConfiguration/components.jelly
varyvol: If it works fine, I'm OK at this point.

src/main/java/com/cloudbees/jenkins/support/config/SupportAutomatedBundleConfiguration.java
Dohbedoh: Oh I see. You're right!

pom.xml
Dohbedoh: The was the only way to leverage the `GlobalConfigurationCategory`. What else need to be done to highlight the pre req ?

src/main/resources/com/cloudbees/jenkins/support/config/SupportAutomatedBundleConfiguration/help-period.html
Dohbedoh: done

varyvol (approved):

Dohbedoh:
@aheritier @varyvol @MRamonLeon any objection before we merge this ?

Dohbedoh:
@varyvol @aheritier any objection before I merge this ?

src/test/java/com/cloudbees/jenkins/support/CasCTest.java
```diff
@@ -1,16 +1,61 @@
...
+                "AboutUser",
+                "AdministrativeMonitors",
+                "AgentProtocols",
+                "BuildQueue",
+                "CustomLogs",
+                "DumpExportTable",
+                "EnvironmentVariables",
+                "FileDescriptorLimit",
+                "ItemsContent",
+                "MasterJVMProcessSystemMetricsContents",
```
amuniz, line 38:
```suggestion
                "ControllerJVMProcessSystemMetricsContents",
```

src/test/java/com/cloudbees/jenkins/support/CasCTest.java
```diff
@@ -1,16 +1,61 @@
...
+                "JenkinsLogs",
+                "LoadStats",
+                "LoggerManager",
+                "Metrics",
+                "NetworkInterfaces",
+                "NodeMonitors",
+                "ReverseProxy",
+                "RunningBuilds",
+                "SlaveCommandStatistics",
+                "MasterSystemConfiguration",
```
amuniz, line 48:
```suggestion
                "ControllerSystemConfiguration",
```

src/test/resources/com/cloudbees/jenkins/support/configuration-as-code.yaml
```diff
@@ -1,3 +1,39 @@
...
+      - "MasterJVMProcessSystemMetricsContents"
+      - "JenkinsLogs"
+      - "LoadStats"
+      - "LoggerManager"
+      - "Metrics"
+      - "NetworkInterfaces"
+      - "NodeMonitors"
+      - "ReverseProxy"
+      - "RunningBuilds"
+      - "SlaveCommandStatistics"
```
amuniz, line 27:
```suggestion
      - "AgentCommandStatistics"
```

src/test/resources/com/cloudbees/jenkins/support/configuration-as-code.yaml
```diff
@@ -1,3 +1,39 @@
...
+      - "AboutUser"
+      - "AdministrativeMonitors"
+      - "AgentProtocols"
+      - "BuildQueue"
+      - "CustomLogs"
+      - "DumpExportTable"
+      - "EnvironmentVariables"
+      - "FileDescriptorLimit"
+      - "ItemsContent"
+      - "MasterJVMProcessSystemMetricsContents"
```
amuniz, line 18:
```suggestion
      - "ControllerJVMProcessSystemMetricsContents"
```

src/test/resources/com/cloudbees/jenkins/support/configuration-as-code.yaml
```diff
@@ -1,3 +1,39 @@
...
+      - "JenkinsLogs"
+      - "LoadStats"
+      - "LoggerManager"
+      - "Metrics"
+      - "NetworkInterfaces"
+      - "NodeMonitors"
+      - "ReverseProxy"
+      - "RunningBuilds"
+      - "SlaveCommandStatistics"
+      - "MasterSystemConfiguration"
```
amuniz, line 28:
```suggestion
      - "ControllerSystemConfiguration"
```

src/test/java/com/cloudbees/jenkins/support/CasCTest.java
```diff
@@ -1,16 +1,61 @@
...
+                "MasterJVMProcessSystemMetricsContents",
+                "JenkinsLogs",
+                "LoadStats",
+                "LoggerManager",
+                "Metrics",
+                "NetworkInterfaces",
+                "NodeMonitors",
+                "ReverseProxy",
+                "RunningBuilds",
+                "SlaveCommandStatistics",
```
amuniz, line 47:
```suggestion
                "AgentCommandStatistics",
```

amuniz (commented):
Made a few suggestions regarding terminology changes in the overall Jenkins community.

PR #233, opened 2020-09-21
[JENKINS-63722] Fix getId() of nested components

Dohbedoh:
[JENKINS-63722](https://issues.jenkins-ci.org/browse/JENKINS-63722) Fixing this by overriding `getId()`. We are also adding a `WARNING` for the different REST APIs and make sure that the old IDs `Master` and `Agents` are still working identically.

### DIFF
--- a/src/test/java/com/cloudbees/jenkins/support/SupportPluginTest.java
+++ b/src/test/java/com/cloudbees/jenkins/support/SupportPluginTest.java
@@ -25,8 +25,10 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
 import java.util.zip.ZipFile;
 
+import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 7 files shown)


src/test/java/com/cloudbees/jenkins/support/SupportPluginTest.java
```diff
@@ -37,6 +39,16 @@
 
     @Rule
     public TemporaryFolder temp = new TemporaryFolder();
+    
+    @Test
+    @Issue("JENKINS-63722")
+    public void testComponentIdsAreUnique() {
```
varyvol, line 45: Could you add another test validating that old component IDs are still working?

varyvol (approved):

src/test/java/com/cloudbees/jenkins/support/SupportPluginTest.java
varyvol: Thanks for adding them!

PR #239, opened 2020-10-13
JENKINS-64172: Support Action names are made distinct

duemir:
To avoid confusion include the name of the Object for which support bundle is generated into the action naem.

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/BundleFileName.java
+++ b/src/main/java/com/cloudbees/jenkins/support/BundleFileName.java
@@ -1,0 +1,85 @@
+package com.cloudbees.jenkins.support;
+
+import com.cloudbees.jenkins.support.api.SupportProvider;
+import org.apache.commons.lang.StringUtils;
+
+import javax.annotation.Nonnull;
+import java.time.Clock;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.Objects;
+
+/**
+ * <p>Generate the support bundle names.</p>
+ *
+ * <p>Format: {@code "${support-provider-name}[_${qualifier}][_${instance_type}]_${date_time}}</p>
+ * <dl>
+ *     <dt>support-provider-name</dt>
+ *     <dd>"support" or the value of {@link SupportProvider#getName()}</dd>
+ *     <dt>qualifier</dt>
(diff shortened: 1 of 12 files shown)


varyvol:
Is this only for the UI or does it affect the generate bundle name?

duemir:
> Is this only for the UI or does it affect the generate bundle name?

Only UI, as far as I am aware.

duemir:
Making bundle names distinct is a good idea as well. I will look into it.

Dohbedoh (approved):

duemir:
So here is how it will look.
For Items the pronoun is used:
Screen Shot 2020-10-22 at 15 30 42
For Runs the task pronoun is used:
Screen Shot 2020-10-22 at 15 30 56
For Nodes it is static "Agent" to match "Delete Agent":
Screen Shot 2020-10-22 at 15 30 17

And here are the bundle names:

support_agent_2020-10-22_05.30.06.zip
support_build_2020-10-22_05.30.57.zip
support_item_2020-10-22_05.30.30.zip

Dohbedoh (approved):

duemir:
@varyvol can you please have a second look. Thank you!

src/main/java/com/cloudbees/jenkins/support/actions/SupportObjectAction.java
```diff
@@ -67,6 +68,8 @@ public String getIconFileName() {
         return "/plugin/support-core/images/24x24/support.png";
     }
 
+    protected abstract String getBundleNameQualifier();
```
varyvol, line 71: I think this would break if any other components is implementing an Action. Maybe providing a default value would be a better option.

Dohbedoh (approved):

src/main/java/com/cloudbees/jenkins/support/BundleFileName.java
```diff
@@ -0,0 +1,85 @@
+package com.cloudbees.jenkins.support;
+
+import com.cloudbees.jenkins.support.api.SupportProvider;
+import org.apache.commons.lang.StringUtils;
+
+import javax.annotation.Nonnull;
```
MRamonLeon, line 6: We should use findbugs/spotbugs annotations

PR #240, opened 2020-10-16
[JENKINS-35342] Allow to download a bundle

dariver:
Initial approach which add a new `Download` button in the `Existing support bundles` section which allows:

- Download one single file
- Download multiple files grouping all of the them in a single zip file (multiBundle(n).zip) where n is the number of files zipped)

This is how it looks the UI

JENKINS-35342

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/SupportAction.java
+++ b/src/main/java/com/cloudbees/jenkins/support/SupportAction.java
@@ -35,6 +35,8 @@ import hudson.security.ACLContext;
 import hudson.security.Permission;
 import jenkins.model.Jenkins;
 import net.sf.json.JSONObject;
+
+import org.apache.commons.io.FileUtils;
 import org.jvnet.localizer.Localizable;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
@@ -52,6 +54,8 @@ import javax.servlet.ServletException;
 import javax.servlet.ServletOutputStream;
 import javax.servlet.http.HttpServletResponse;
 import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -61,6 +65,8 @@ import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipOutputStream;
 
 /**
  * Main root action for generating support.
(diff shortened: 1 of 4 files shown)


src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -185,6 +181,78 @@ public void doDeleteBundles(StaplerRequest req, StaplerResponse rsp) throws Serv
...
+            rsp.setContentType("application/zip");
+            rsp.addHeader("Content-Disposition", "inline; filename=" + fileToDownload.getName() + ";");
+            FileUtils.copyFile(fileToDownload, rsp.getOutputStream());
+        } catch (RuntimeException e) {
+            logger.log(Level.SEVERE, "Unable to download file " + fileToDownload.getAbsolutePath(), e);
+        } finally {
+            if (bundlesToDownload.size() > 1 && fileToDownload.delete()) {
+                logger.log(Level.FINE, "Temporary multiBundle file deleted: " + fileToDownload.getAbsolutePath());
+            } else {
+                logger.log(Level.SEVERE, "Unable to delete temporary multiBundle file: " + fileToDownload.getAbsolutePath());
```
Dohbedoh, line 211: If there is a `RuntimeException` and `bundlesToDownload` is 1, then we will see `Unable to delete temporary multiBundle file: ...` that is wrong since there was not temporary bundle file created in that case.

IIUC, this shouldbe:

```
if (bundlesToDownload.size() > 1) {
    if (! fileToDownload.delete()) {
        logger.log(Level.FINE, "Temporary multiBundle file deleted: " + fileToDownload.getAbsolutePath());
    } else {
        logger.log(Level.SEVERE, "Unable to delete temporary multiBundle file: " + fileToDownload.getAbsolutePath());
    }
}
```

https://docs.oracle.com/javase/8/docs/api/java/io/File.html#createTempFile-java.lang.String-java.lang.String-

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -185,6 +181,78 @@ public void doDeleteBundles(StaplerRequest req, StaplerResponse rsp) throws Serv
...
+                    logger.log(Level.FINE, "The bundle selected {0} does not exist, so it will not be processed", s.getName());
+                }
+            }
+        }
+        return bundles;
+    }
+
+    private File createZipFile(Set<String> bundles) throws IOException {
+        File rootDirectory = SupportPlugin.getRootDirectory();
+        File zipFile = File.createTempFile(
```
Dohbedoh, line 234: I would also mark the deletion of the temp file when Jenkins exits with [zipFile.deleteOnExit()](https://docs.oracle.com/javase/8/docs/api/java/io/File.html#deleteOnExit--).

Dohbedoh (commented):
We should add a test for the `downloadBundles` action.

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
dariver: Done

dariver:
Added tests for downloadBundles

Dohbedoh (approved):

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -185,6 +181,81 @@ public void doDeleteBundles(StaplerRequest req, StaplerResponse rsp) throws Serv
...
+                    int length;
+                    while ((length = fis.read(buffer)) > 0) {
+                        zos.write(buffer, 0, length);
+                    }
+                }
+            }
+        } catch (IOException e) {
+            logger.log(Level.SEVERE, "Error creating zip file: " + zipFile.getAbsolutePath(), e);
+        }
+        zipFile.deleteOnExit();
```
Dohbedoh, line 255: @dariver you should do this right after creating the file, before the `try {} catch {}`

Dohbedoh (changes requested):

Dohbedoh (approved):

varyvol (approved):

PR #246, opened 2020-11-20
[JENKINS-42012] Wrapper to make Channel.callAsync really async

fbelzunc:
Create a wrapper to make `Channel.callAsync` really async. The problem is that `callAsync` is performing a remoting call before returning the Future, so the call is not really async - see https://github.com/jenkinsci/remoting/blob/remoting-4.6/src/main/java/hudson/remoting/Request.java#L238 

This PR should fix both Jira issues below:

* https://issues.jenkins-ci.org/browse/JENKINS-42011 
* https://issues.jenkins.io/browse/JENKINS-42012

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/util/CallAsyncWrapper.java
+++ b/src/main/java/com/cloudbees/jenkins/support/util/CallAsyncWrapper.java
@@ -14,10 +14,9 @@ import java.util.concurrent.TimeoutException;
 public final class CallAsyncWrapper {
 
     public static <V,T extends Throwable> hudson.remoting.Future<V> callAsync(final VirtualChannel channel, final Callable<V,T> callable) throws IOException {
-
         Future<hudson.remoting.Future<V>> future = Computer.threadPoolForRemoting.submit(new java.util.concurrent.Callable<hudson.remoting.Future<V>>() {
             @Override
-            public hudson.remoting.Future<V> call() throws Exception {
+            public hudson.remoting.Future<V> call() throws IOException {
                 return channel.callAsync(callable);
             }
         });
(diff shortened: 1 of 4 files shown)


Dohbedoh (approved):

src/main/java/com/cloudbees/jenkins/support/util/CallAsyncWrapper.java
```diff
@@ -0,0 +1,30 @@
...
+    public static <V,T extends Throwable> hudson.remoting.Future<V> callAsync(final VirtualChannel channel, final Callable<V,T> callable) throws IOException {
+        Future<hudson.remoting.Future<V>> future = Computer.threadPoolForRemoting.submit(new java.util.concurrent.Callable<hudson.remoting.Future<V>>() {
+            @Override
+            public hudson.remoting.Future<V> call() throws IOException {
+                return channel.callAsync(callable);
+            }
+        });
+        try {
+            return future.get(SupportPlugin.REMOTE_OPERATION_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+        } catch (InterruptedException | ExecutionException | TimeoutException e) {
```
Dohbedoh, line 25: We may want to want to cancel the future on `TimeoutException` as nobody else is interested in the future after the exception is raised. Otherwise the task would still be executing:

```
        } catch (InterruptedException | ExecutionException e) {
            throw new IOException(e);
        }  catch (TimeoutException te) {
            future.cancel(true);
            throw new IOException(e);
        }
```

Would like maybe @MRamonLeon to confirm

MRamonLeon:
@oleg-nenashev, @jeffret-b I would love to see your thoughts on this? It seems like a so-called async method of **remoting** is not so. Isn't this _feature/bug_ affecting everywhere? 
https://github.com/jenkinsci/remoting/blob/remoting-4.6/src/main/java/hudson/remoting/Request.java#L238

jeffret-b:
I'll take a look when I get a chance within the next couple of days.

jeffret-b:
The `Channel.callAsync()` method works as expected in the tests. I've tried a number of them to see that the method returns without waiting for the callable to finish.

jeffret-b:
Let me know if you need any further assistance with this. Possibly you've got a scenario with different behavior than what I see in the tests.

PR #251, opened 2020-12-09
[FIXED JENKINS-64601] Improve write performance for big files under $JENKINS_HOME/logs/tasks/ by using BufferedOutputStream instead of OutputStream

fbelzunc:
See the Jira for all the details https://issues.jenkins.io/browse/JENKINS-64401

Basically, there is a performance issue which happens for all support bundles generated when {{$JENKINS_HOME/logs/tasks/}} have big text files (you can reproduce the issue with around 6 files bigger than 50 MB each of them).

The current proposal is to use {{BufferedOutputStream}} instead of {{FilterOutputStream}} as stream so we can write bytes to the underlying output stream without necessarily causing a call to the underlying system for each byte written. 

In my example, without the fix, a support bundle with the issue exposed is taking around 15 minutes to be generated. With the current implementation, the support bundle is generated in less than 20 seconds.

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
+++ b/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
@@ -81,7 +81,6 @@ import java.io.BufferedOutputStream;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
-import java.io.FilterOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.io.PrintStream;
--- a/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
+++ b/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
@@ -81,6 +81,7 @@ import java.io.BufferedOutputStream;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
+import java.io.FilterOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.io.PrintStream;
@@ -329,7 +330,7 @@ public class SupportPlugin extends Plugin {
                 OutputStream textOut = maybeFilteredOut.map(OutputStream.class::cast).orElse(binaryOut);
                 OutputStreamSelector selector = new OutputStreamSelector(() -> binaryOut, () -> textOut);
                 IgnoreCloseOutputStream unfilteredOut = new IgnoreCloseOutputStream(binaryOut);
-                IgnoreCloseOutputStream filteredOut = new IgnoreCloseOutputStream(selector);
+                IgnoreCloseOutputStream filteredOut = new IgnoreCloseOutputStream(new FilterOutputStream(selector));
                 boolean entryCreated = false;
                 for (Content content : contents) {
                     if (content == null) {
--- a/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
+++ b/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
@@ -330,7 +330,7 @@ public class SupportPlugin extends Plugin {
                 OutputStream textOut = maybeFilteredOut.map(OutputStream.class::cast).orElse(binaryOut);
                 OutputStreamSelector selector = new OutputStreamSelector(() -> binaryOut, () -> textOut);
                 IgnoreCloseOutputStream unfilteredOut = new IgnoreCloseOutputStream(binaryOut);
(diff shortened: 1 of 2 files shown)


src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -329,7 +330,7 @@ public static void writeBundle(OutputStream outputStream, final List<? extends C
                 OutputStream textOut = maybeFilteredOut.map(OutputStream.class::cast).orElse(binaryOut);
                 OutputStreamSelector selector = new OutputStreamSelector(() -> binaryOut, () -> textOut);
                 IgnoreCloseOutputStream unfilteredOut = new IgnoreCloseOutputStream(binaryOut);
-                IgnoreCloseOutputStream filteredOut = new IgnoreCloseOutputStream(selector);
+                IgnoreCloseOutputStream filteredOut = new IgnoreCloseOutputStream(new FilterOutputStream(selector));
```
Dohbedoh, line 333: @fbelzunc do we actually still need to wrap this up with a `FilterOutputStream` ? Seems to me that this is already done if needed (in case of anonymization) at https://github.com/jenkinsci/support-core-plugin/blob/master/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java#L328. @MRamonLeon could confirm.

Dohbedoh (commented):
Glad you spotted this. Looks good to me, just a comment about the

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
fbelzunc: I think you are right. I just made a commit to fix this. I also tried to generate an anonymous bundle and it worked without any problem after I perform the chage.

src/main/java/com/cloudbees/jenkins/support/util/IgnoreCloseOutputStream.java
```diff
@@ -28,21 +28,36 @@
...
 @Restricted(NoExternalUse.class)
-public final class IgnoreCloseOutputStream extends FilterOutputStream implements WrapperOutputStream {
-    public IgnoreCloseOutputStream(@Nonnull OutputStream out) {
+public final class IgnoreCloseOutputStream extends BufferedOutputStream implements WrapperOutputStream {
+    /**
+     * The underlying output stream to be filtered.
+     *
+     * @param out the underlying output stream.
+     */
+    public IgnoreCloseOutputStream(OutputStream out) {
```
MRamonLeon, line 47: Any reason to remove the `@Nonnull`? You can use the findbugs version (`@NonNull`).

src/main/java/com/cloudbees/jenkins/support/util/IgnoreCloseOutputStream.java
```diff
@@ -28,21 +28,36 @@
...
         super(out);
     }
 
+    /**
+     * The underlying output stream to be filtered.
+     *
+     * @param   out    the underlying output stream.
```
MRamonLeon, line 54:
```suggestion
     * @param out the underlying output stream.
```

src/main/java/com/cloudbees/jenkins/support/util/IgnoreCloseOutputStream.java
```diff
@@ -28,21 +28,36 @@
...
+     * @param   size   the buffer size.
```
MRamonLeon, line 55:
```suggestion
     * @param size the buffer size.
```

src/main/java/com/cloudbees/jenkins/support/util/IgnoreCloseOutputStream.java
```diff
@@ -52,4 +67,4 @@ public void close() throws IOException {
     public @Nonnull OutputStream unwrap() {
         return out;
     }
-}
+}
```
MRamonLeon, line 70:
```suggestion
}

```

src/main/java/com/cloudbees/jenkins/support/util/IgnoreCloseOutputStream.java
```diff
@@ -28,21 +28,36 @@
...
         super(out);
     }
 
+    /**
+     * The underlying output stream to be filtered.
+     *
+     * @param   out    the underlying output stream.
+     * @param   size   the buffer size.
+     */
+    public IgnoreCloseOutputStream(OutputStream out, int size) {
```
MRamonLeon, line 57: Is it created to be used some way? If not, better to keep it simple. We can create this one when the need comes

MRamonLeon (approved):
Just minor comments.

Dohbedoh (approved):

PR #259, opened 2020-12-29
Update year format for BundleNamePrefixTest

bmunozm:
The format is wrong. It should be "yyyy" instead of "YYYY". "YYYY" stands for the week year, not the year, which makes the latest week of the year this test is failing (see https://docs.oracle.com/javase/8/docs/api/java/text/SimpleDateFormat.html)

Thanks Thierry Wasylczenko for getting the root cause of the issue.


batmat (approved):

batmat:
Infra failure, retriggering

Dohbedoh (approved):

PR #268, opened 2021-01-20
JENKINS-64667 - Check for more reverse proxy headers

duemir:
Presence of the standard `Forwarded` header and the rest of non-standard headers are added:

- `X-Forwarded-Proto`
- `X-Forwarded-Host`
- `X-Forwarded-Port`


Dohbedoh (commented):
Can we add these to the `ReverseProxyTest`

Dohbedoh (approved):

PR #304, opened 2021-08-23
Fix test which will fail when upgrading to remoting 4.8 (i.e core 2.294)

olamy:
Signed-off-by: Olivier Lamy <olamy@apache.org>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -211,4 +211,20 @@
     </dependency>
   </dependencies>
 
+  <build>
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-surefire-plugin</artifactId>
+          <configuration>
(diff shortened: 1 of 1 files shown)


car-roll (approved):

bmunozm (approved):

Dohbedoh (commented):
So starting with Remoting 4.8, the [DumpExportTable](https://github.com/jenkinsci/support-core-plugin/blob/master/src/main/java/com/cloudbees/jenkins/support/impl/DumpExportTable.java) will always collect an empty output. Is that correct ?

Pinging @res0nance as he worked on https://github.com/jenkinsci/remoting/pull/441.

olamy:
not empty but less verbose :) see https://github.com/jenkinsci/remoting/pull/441

Dohbedoh (approved):

alecharp (approved):

res0nance (approved):

PR #310, opened 2021-10-14
[JENKINS-66445] Only print a warning if file from system property doe…

Dohbedoh:
…s not exist

[JENKINS-66445](https://issues.jenkins.io/browse/JENKINS-66445)

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
+++ b/src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
@@ -162,14 +162,15 @@ public class ContentMappings extends ManagementLink implements Saveable, Iterabl
 
     private static Set<String> getAdditionalStopWords() {
         Set<String> words = new HashSet<>();
-        String fileLocation = System.getProperty(ADDITIONAL_STOP_WORDS_PROPERTY, SupportPlugin.getRootDirectory()
-            + "/" + ADDITIONAL_STOP_WORDS_FILENAME);
-        if (fileLocation != null) {
-            LOGGER.log(Level.FINE, "Attempting to load user provided stop words from ''{0}''.", fileLocation);
-            File f = new File(fileLocation);
-            if (!f.exists() || !f.canRead()) {
-                LOGGER.log(Level.WARNING, "Could not load user provided stop words as " + fileLocation 
-                    + " does not exist or is not readable.");
+        String fileLocationFromProperty = System.getProperty(ADDITIONAL_STOP_WORDS_PROPERTY);
+        String fileLocation = fileLocationFromProperty == null
+            ? SupportPlugin.getRootDirectory() + "/" + ADDITIONAL_STOP_WORDS_FILENAME
+            : fileLocationFromProperty;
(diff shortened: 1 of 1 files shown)


src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
```diff
@@ -162,14 +162,15 @@ private ContentMappings(@Nonnull XmlProxy proxy) {
...
         Set<String> words = new HashSet<>();
-        String fileLocation = System.getProperty(ADDITIONAL_STOP_WORDS_PROPERTY, SupportPlugin.getRootDirectory()
-            + "/" + ADDITIONAL_STOP_WORDS_FILENAME);
-        if (fileLocation != null) {
-            LOGGER.log(Level.FINE, "Attempting to load user provided stop words from ''{0}''.", fileLocation);
-            File f = new File(fileLocation);
-            if (!f.exists() || !f.canRead()) {
-                LOGGER.log(Level.WARNING, "Could not load user provided stop words as " + fileLocation 
-                    + " does not exist or is not readable.");
+        String fileLocationFromProperty = System.getProperty(ADDITIONAL_STOP_WORDS_PROPERTY);
```
MRamonLeon, line 165: Why did you split this into two more complex sentences? I prefer the simplest way if there is no need to change it.

Dohbedoh: If the system property is used, but the path provided does not exist, we want to log a warning.
If the default file location does not exist, we do not want to log a warning.
In both cases those, the logic will be the same. There is just a particular scenario (system property used but file not exist where we want to log something).
Does that make sense ?

src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
```diff
@@ -179,10 +180,12 @@ private ContentMappings(@Nonnull XmlProxy proxy) {
                     }
                     return words;
                 } catch (IOException ex) {
-                    LOGGER.log(Level.WARNING, "Could not load user provided stop words as " + fileLocation 
+                    LOGGER.log(Level.WARNING, "Could not load user provided stop words as " + fileLocation
```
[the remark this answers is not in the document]
Dohbedoh: I see and I agree.

src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
```diff
@@ -179,10 +180,12 @@ private ContentMappings(@Nonnull XmlProxy proxy) {
...
                         + " does not exist or is not readable.", ex);
```
Dohbedoh, line 184:
```suggestion
                    LOGGER.log(Level.WARNING, "Could not load user provided stop words. there was an error reading " + fileLocation, ex);
```

src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
MRamonLeon: I see now, thanks

PR #324, opened 2021-12-03
`SupportLogHandler.records` could leak memory

jglick:
As discovered in https://github.com/jenkinsci/bom/pull/644#issuecomment-985156639 it seems that the ring buffer could hold a strong ref to `LogRecord.parameters` for recently logged messages, including things like

```
FINER	o.j.p.w.cps.CpsFlowExecution#cleanUpLoader: ignoring org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.SandboxResolvingClassLoader@…
```

which really ought to have been collected.

Eagerly formatting the records is not a good option for these sub-`INFO` messages which might otherwise never be formatted, so the most straightforward fix seems to be to use `SoftReference` to offer these records most of the time to https://github.com/jenkinsci/support-core-plugin/blob/bb61a87442c7d5ea391214bd97cd71edb422e8d7/src/main/java/com/cloudbees/jenkins/support/SupportLogHandler.java#L199 callers, if any, without forcing them to be held.

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/SupportLogHandler.java
+++ b/src/main/java/com/cloudbees/jenkins/support/SupportLogHandler.java
@@ -38,6 +38,8 @@ import java.io.IOException;
 import java.io.OutputStreamWriter;
 import java.io.UnsupportedEncodingException;
 import java.io.Writer;
+import java.lang.ref.Reference;
+import java.lang.ref.SoftReference;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


PierreBtz (approved):
Maybe a naive question but I wonder why we do not use a `MemoryHandler` to handle this for us.
On another side, looking into Jenkins' codeline, I see a `RingBufferLogHandler` that might suffer from the same issue.

Anyway, lgtm, thanks a lot for analyzing https://github.com/jenkinsci/bom/pull/644!

jglick:
> why we do not use a `MemoryHandler` to handle this for us

WDYM? That uses strong references as well.

> looking into Jenkins' codeline, I see a `RingBufferLogHandler` that might suffer from the same issue

Indeed. That would not show up during functional tests, because `LoggerRule` directly sets log levels as well as activating its own handler, but in production usage it is possible `LogRecorder` would have the same temporary leak. I will look into a core patch for it.

Dohbedoh (approved):

PR #350, opened 2022-03-17
Use `systemPropertyVariables` rather than `argLine`

basil:
In https://github.com/jenkinsci/bom/pull/935 I have been trying to test this plugin with Java 17, which requires setting `argLine` to include some `--add-opens` directives, but this plugin rather unhelpfully wipes them out to set its own `argLine`. This PR changes the POM to use `systemPropertyVariables` rather than wiping out `argLine`. With this change I can successfully test on Java 11.

As a bonus, I also incrementalified the plugin and updated the Jenkins version and plugin BOM. This should help me more easily consume this build from https://github.com/jenkinsci/bom/pull/935.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -41,7 +41,7 @@
   <description>
     This plugin provides a common set of classes to assist in generating support bundles.
   </description>
-  <url>https://github.com/jenkinsci/support-core-plugin</url>
+  <url>https://github.com/jenkinsci/${project.artifactId}-plugin</url>
   <licenses>
     <license>
       <name>The MIT license</name>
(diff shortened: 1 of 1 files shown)


timja (approved):

basil:
Both the first and second commits of this PR have been successfully tested with PCT and Java 17 in jenkinsci/bom#935, so I think we are ready to move forward with this. A merge and release of this PR would be helpful in Java 17 compatibility testing efforts. CC @Dohbedoh

Dohbedoh (approved):
Thanks for that effort @timja

basil:
Thank you!

Dohbedoh:
np @basil I was able to finally get this this released: https://github.com/jenkinsci/support-core-plugin/releases/tag/1140.vb_b_3b_7d866b_a_8.

PR #358, opened 2022-04-16
Pin jackson2-api temporarily

jetersen:
Pinned jackson2-api to avoid cyclic dependency

We need to pin this to unblock https://github.com/jenkinsci/bom/pull/1011#issuecomment-1100675121

Can be removed once BOM and metrics plugin is updated to using that BOM

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -87,7 +87,7 @@
       <url>https://repo.jenkins-ci.org/public/</url>
     </pluginRepository>
   </pluginRepositories>
-  
+
   <dependencyManagement>
     <dependencies>
       <dependency>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -191,6 +192,13 @@
       <groupId>io.jenkins.configuration-as-code</groupId>
       <artifactId>test-harness</artifactId>
       <scope>test</scope>
+      <!-- TODO https://github.com/jenkinsci/bom/pull/1011 -->
+      <exclusions>
+        <exclusion>
+          <groupId>org.jenkins-ci.plugins</groupId>
+          <artifactId>jackson2-api</artifactId>
+        </exclusion>
+      </exclusions>
```
jetersen, line 201: Only temporary until https://github.com/jenkinsci/bom/pull/1011 is resolved.

I could go to test-harness and attempt to make another PR but it would be yet another PR that blocks bom...

jetersen:
@Dohbedoh would appreciate a review / release to unblock the BOM.

basil (approved):

basil:
Now that BOM has been released I think we should close this in favor of #360.

PR #362, opened 2022-04-23
Fix test on Java 17

basil:
While running PCT on Java 17 I noticed [this test failure](https://ci.jenkins.io/job/Tools/job/bom/job/PR-935/36/testReport/junit/com.cloudbees.jenkins.support.impl/FileDescriptorLimitTest/pct_support_core_weekly___addContentsFiltered/), which I could also reproduce locally:

```
java.lang.AssertionError: 

Expected: a string containing "core file size"
     but: was "Jenkins
======

N/A: Either no connection to node or no cached result
"
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:6)
	at com.cloudbees.jenkins.support.impl.FileDescriptorLimitTest.addContentsFiltered(FileDescriptorLimitTest.java:66)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.jvnet.hudson.test.JenkinsRule$1.evaluate(JenkinsRule.java:606)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:299)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:293)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.lang.Thread.run(Thread.java:833)
```

Debugging the test locally, I saw that the code was working correctly but we were timing out before we ever got the result. It seems this code path has gotten a few cycles slower in Java 17, but by increasing the timeout the test is fully functional on Java 17.

PR #366, opened 2022-05-09
[JENKINS-68339] Skip remoting stack on built-in node and improve IPv6 Regex

Dohbedoh:
[JENKINS-68339](https://issues.jenkins.io/browse/JENKINS-68339)

* Only use `channel.callAsync` (through the `CallAsyncWrapper`) if retrieving data on remote node. When retrieving data on the built-in node, just execute the callable.
* updated CI to build use containers, and build on JDK 11 and JDK 17.

I am hoping that this fix the flakyness with JDK 17 as we remove the remoting overhead for this use case. See the following for reference:

* https://github.com/jenkinsci/support-core-plugin/pull/363
* https://github.com/jenkinsci/support-core-plugin/pull/362

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/AsyncResultCache.java
+++ b/src/main/java/com/cloudbees/jenkins/support/AsyncResultCache.java
@@ -40,8 +40,7 @@ public class AsyncResultCache<T> implements Runnable {
         
         if (node == null) return null;
         Future<V> future;
-        // If running in on the built-in node, no need to use the CallAsyncWrapper or be subjected 
-        // to the REMOTE_OPERATION_TIMEOUT_MS
+        // If launching execution on the built-in node, no need to use the CallAsyncWrapper
         if (node instanceof Jenkins) {
             future = Computer.threadPoolForRemoting.submit(() -> {
                 try {
--- a/src/main/java/com/cloudbees/jenkins/support/AsyncResultCache.java
+++ b/src/main/java/com/cloudbees/jenkins/support/AsyncResultCache.java
@@ -4,13 +4,13 @@ import com.cloudbees.jenkins.support.util.CallAsyncWrapper;
 import hudson.model.Computer;
 import hudson.model.Node;
 import hudson.remoting.Callable;
-import hudson.remoting.Future;
 import hudson.remoting.VirtualChannel;
 import jenkins.model.Jenkins;
 
 import java.io.IOException;
 import java.util.WeakHashMap;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.logging.Level;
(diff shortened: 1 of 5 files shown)


src/main/java/com/cloudbees/jenkins/support/AsyncResultCache.java
```diff
@@ -37,29 +37,35 @@ public static <V, T extends java.lang.Throwable> V get(Node node, WeakHashMap<No
...
+        
         if (node == null) return null;
-        VirtualChannel channel = node.getChannel();
-        if (channel == null) {
-            synchronized (cache) {
-                return cache.get(node);
+        Future<V> future;
+        // If running in on the built-in node, no need to use the CallAsyncWrapper or be subjected 
+        // to the REMOTE_OPERATION_TIMEOUT_MS
+        if (node instanceof Jenkins) {
```
basil, line 45: Might be safer to call `JenkinsJVM.isJenkinsJVM()` if this might run on an agent? Not sure if `node instanceof Jenkins` will trigger classloading of the main `Jenkins` class on agents (which would be undesirable).

basil (approved):
Very nice!

src/main/java/com/cloudbees/jenkins/support/AsyncResultCache.java
Dohbedoh: The AsyncResultCache actually always executes on the controller JVM. It accepts the node and the callable and handle the asynchronous execution. So here we check if the node in which **we are going to** execute the callable is the controller or not.
Maybe I can rephrase my comment.

Dohbedoh:
Still failing with the timeout though 🤔

Dohbedoh:
That regexp for IPv6 sure is taking a toll.. I have tested others like the one from the Regulare Expression Cookbook:

```
private static final String IPv6 = "(?:(?:(?:[A-Fa-f0-9]{1,4}:){6}|(?=(?:[A-Fa-f0-9]{0,4}:){0,6}(?:[0-9]{1,3}\\.){3}[0-9]{1,3}(?![:.\\w]))(([0-9A-Fa-f]{1,4}:){0,5}|:)((:[0-9A-Fa-f]{1,4}){1,5}:|:))(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)|(?:[A-Fa-f0-9]{1,4}:){7}[A-Fa-f0-9]{1,4}|(?=(?:[A-Fa-f0-9]{0,4}:){0,7}[A-Fa-f0-9]{0,4}(?![:.\\w]))(([0-9A-Fa-f]{1,4}:){1,7}|:)((:[0-9A-Fa-f]{1,4}){1,7}|:))(?![:.\\w])";
```

But it is also kind of slow. 

The one from http://www.java2s.com/example/java/java.util.regex/is-ipv6-address-by-regex.html seems promising and simpler:

```
    private static final String IPv6_STANDARD = "(?:[0-9a-fA-F]{1,4}:){7}[0-9a-fA-F]{1,4}";
    private static final String IPv6_COMPRESSED = "((?:[0-9A-Fa-f]{1,4}(?::[0-9A-Fa-f]{1,4})*)?)::((?:[0-9A-Fa-f]{1,4}(?::[0-9A-Fa-f]{1,4})*)?)";
```

Will do more test. but this is most likely the better solution to the problem we are trying to solve.

Dohbedoh:
I realized that regexp for IPv6 addresses at https://github.com/jenkinsci/support-core-plugin/blob/1162.vb_b_e5198c6b_22/src/main/java/com/cloudbees/jenkins/support/filter/InetAddressContentFilter.java#L61 is not even capturing all IPv6. It misses the compressed and mixed addresses... And above all, it is terribly slow...

I am updating tests that reflect this. And will then update a solution that seem promising.

Dohbedoh:
Those tests shows that the current regex does not find valid IPv6 compressed addresses like:

```
Other found falsifying value(s) :- 
::0:0:0:0:0:1:0
::0:0:0:0:2:1:0
::0:0:0:0:2:4:0
::0:0:0:0:d:4:0
::0:0:0:0:d:9:0
::0:0:0:0:d:f:0
::0:0:0:0:d:16:0
::0:0:0:0:e:16:0
::0:0:0:0:f:16:0
::0:0:0:0:f:3c:0
```

But also notice how the test take seconds in JDK 11 and more than a minute on JDK 17.. Now pushing an improved regexp.

Dohbedoh:
Now all the tests run in <1s.

Dohbedoh:
I also tested this in different environments and that looks good. And seems to be reducing the generation time when anonymization is on.

PR #377, opened 2022-06-13
[JENKINS-68743] Report if a node is marked offline

Dohbedoh:
[JENKINS-68743](https://issues.jenkins.io/browse/JENKINS-68743) Report if a node is marked offline. 

Note [Computer.java#isTemporarilyOffline](https://github.com/jenkinsci/jenkins/blob/jenkins-2.332.3/core/src/main/java/hudson/model/Computer.java#L667-L684) is marked as deprecated to warn people that they should use `Computer#isOffline` to truly know about a node status. What we want to know here is if that `temporarilyOffline` attribute is set to `true` for troubleshooting purposes.

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
@@ -64,6 +64,7 @@ import java.util.Date;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
 import java.util.TimeZone;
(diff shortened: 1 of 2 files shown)


src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
```diff
@@ -643,6 +644,13 @@ private String getLabelString(Node n) {
                     Markdown.escapeBacktick(ContentFilter.filter(filter, jenkins.getRootDir().getAbsolutePath())) + "`");
             out.println("      - Labels:         " + ContentFilter.filter(filter, getLabelString(jenkins)));
             out.println("      - Usage:          `" + jenkins.getMode() + "`");
+            Optional.ofNullable(jenkins.toComputer()).ifPresent(computer ->
+                out.println("      - Marked Offline: " + computer.isTemporarilyOffline()));
+            if(jenkins.getChannel() == null) {
```
Dohbedoh, line 649: Added also the `- Status` for the built-in node to bring parity with other nodes. Not sure if it is possible to have a `null` channel for the built-in node ? unless maybe a plugin or a script does this explicitly.

PR #380, opened 2022-06-23
Improve support bundle generation page

alecharp:
I'm trying to see how we can improve the UX of the plugin. 
First in my list is the huge page to select what to include in the bundle to be generated.

<details>
<summary>
Here the current display with <code>2.346.1</code>
</summary>

Screenshot 2022-06-23 at 17-18-32 Support Jenkins

</details>

Do you think this is worth the effort to continue? 

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/api/Component.java
+++ b/src/main/java/com/cloudbees/jenkins/support/api/Component.java
@@ -24,6 +24,7 @@
 
 package com.cloudbees.jenkins.support.api;
 
+import com.cloudbees.jenkins.support.Messages;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.ExtensionPoint;
 import hudson.model.AbstractModelObject;
@@ -31,6 +32,7 @@ import hudson.security.ACL;
 import hudson.security.Permission;
 import jenkins.model.Jenkins;
 import org.acegisecurity.Authentication;
+import org.jvnet.localizer.Localizable;
 import org.kohsuke.stapler.export.Exported;
 import org.kohsuke.stapler.export.ExportedBean;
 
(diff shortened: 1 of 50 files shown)


aheritier (commented):
I think it's an important feature to group the items by categories
I think we have the RFE in our Jira backlog.
It was discussed several times with @Dohbedoh and @PierreBtz

alecharp:
I updated the PR with more component with specific categories.

alecharp:
We can discuss the category, I had no real idea how to divides the current components.

aheritier:
Yes not sure about the granularity
I think the categories should guide the users to grab everything for a kind of issue

Something to add somewhere in the plugin backlog: It shouldn't be possible to not bundle the `About Jenkins` file. A bundle without this file is 99.9999% useless

alecharp:
> I think the categories should guide the users to grab everything for a kind of issue

The problem with this is that some components are required for different type of problems. I'd like to see a selector to say "Slowness" and automatically select a bunch of components. But that should be different from those categories imo.

> Something to add somewhere in the plugin backlog: It shouldn't be possible to not bundle the About Jenkins file. A bundle without this file is 99.9999% useless

True. We should also revisit some components where they have a too wide of a scope. The list of plugins should on its own component for example, and selected by default maybe.

src/main/java/com/cloudbees/jenkins/support/api/Component.java
```diff
@@ -140,6 +142,59 @@ public <C extends AbstractModelObject> boolean isApplicable(Class<C> clazz) {
...
+        /**
+         * For components related to the logging system or the logs themself.
+         */
+        LOGS(Messages._SupportPlugin_Category_Logs()),
+        /**
+         * Anything that doesn't fit any other categories.
+         */
+        MISC(Messages._SupportPlugin_Category_Misc()),
+        /**
+         * For components related to how the controller is running, how it is running, where it is running.
```
Dohbedoh, line 182:
```suggestion
         * For components related to how the controller is running and where it is running.
```

Dohbedoh (approved):

Dohbedoh:
That looks good to me thanks 👍 

> The problem with this is that some components are required for different type of problems. I'd like to see a selector to say "Slowness" and automatically select a bunch of components. But that should be different from those categories imo.

Like a profile. Yes that's something that we can add later on. 

I thought about a whole overhaul of the UI for components, such as having a jelly component for the selection, so that plugins like for example Advisor can use and leverage any improvement that we make. Was also thinking about splitting up the SupportAction to have different "sub actions" below it. One for generating bundles, one for managing existing bundles and data, etc... Maybe I can put this into JIRA

> We should also revisit some components where they have a too wide of a scope. The list of plugins should on its own component for example, and selected by default maybe.

Right, AboutJenkins needs to be split up in a Plugins component and Nodes components https://issues.jenkins.io/browse/JENKINS-54920. ThreadDumps too, actually take dumps from controller and connected agents..

Dohbedoh:
We can document this in issues.jenkins.io. In the meantime, I think that we can merge this.

PR #465, opened 2023-05-09
More gracefully handle empty `bundlesToDownload`

jglick:
In a support bundle (ironically enough) I noticed a stack trace

```
java.util.NoSuchElementException
	at java.base/java.util.HashMap$HashIterator.nextNode(HashMap.java:1513)
	at java.base/java.util.HashMap$KeyIterator.next(HashMap.java:1534)
	at com.cloudbees.jenkins.support.SupportAction.doDownloadBundles(SupportAction.java:227)
```

which looks like it could come from an attempt to download no bundles. Should be handled more clearly.

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/SupportAction.java
+++ b/src/main/java/com/cloudbees/jenkins/support/SupportAction.java
@@ -29,6 +29,7 @@ import com.cloudbees.jenkins.support.api.SupportProvider;
 import com.cloudbees.jenkins.support.filter.ContentFilters;
 import hudson.Extension;
 import hudson.model.Api;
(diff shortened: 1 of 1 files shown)


jglick:
Tried to reproduce actually, but without any existing bundles, the button is offered and if you click it you get a 400 browser error, which is not too friendly either https://github.com/jenkinsci/support-core-plugin/blob/5a4a2287ce97233cf9ef06cab67e7a57b8fc1822/src/main/java/com/cloudbees/jenkins/support/SupportAction.java#L192-L195

Dohbedoh (approved):

PR #473, opened 2023-06-12
[JENKINS-57138] Address anonymization performance issues

Dohbedoh:
[JENKINS-57138](https://issues.jenkins.io/browse/JENKINS-57138) Improved the Anonymization resource usage and process times. The anonymization of Sensitive Content (Jenkins item, users, labels, etc) is extremely slow:

* It regenerates search and replacement lists for every single line
* the ignore case scenario requires to change the casing of both the input and content mapping word, redundantly

After addressing some of those issues, the timings were improved a bit. But I then try to use a general `Pattern` compiled with all content mappings and apply it for every line which considerably improved the efficiency!

Then reading through Regex optimization techniques, I extracted the first characters of content mapping words to reduce backtracking. For example instead of `a|abc|abcd|bcd)` use `a(?:bc|bcd)?|bcd`. And that improved it even more.

So then I pushed further trying to apply the same recursively. I started working on a structure to create a regexp following this strategy. And found out about the `Trie` structure. And after testing a Trie-based regexp I end up with the very good times and resource usage.

Note: I thought a further improvement could be to use case insensitive mapping for the replacement. Here, we convert mapping to lowercase to create a replacement map. And then during the processing of a file, we convert the line to lowercase when passing it to the matcher. That help us achieve case insensitive word replacement. I thought that maybe creating a "case insensitive replacement map" could help save even more time during processing... but maybe at a cost of more resource (memory) usage. Actually when using some of the case insensitive implementation, we don't save time. It's quite similar (or worse in case of the TreeMap with comparator).

### Testing done

In a controller with 16k items, here is the comparison of background bundle generation (150 MB uncompressed):

* anonymization with previous method: **> 6 hours**
* anonymization with proposed change: **~40s**
* without anonymization: **~8/9s**

And the Memory (and especially CPU consumption) is reduced considerably.

### Submitter checklist

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilter.java
+++ b/src/main/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilter.java
@@ -24,18 +24,24 @@
 
 package com.cloudbees.jenkins.support.filter;
 
-import com.cloudbees.jenkins.support.util.WordReplacer;
+import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.Extension;
 import hudson.ExtensionList;
 import org.apache.commons.lang.StringUtils;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
 
-import edu.umd.cs.findbugs.annotations.NonNull;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
+import java.util.Map;
 import java.util.Set;
+import java.util.TreeMap;
+import java.util.function.Function;
+import java.util.regex.MatchResult;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 /**
  * Filters contents based on names provided by all {@linkplain NameProvider known sources}.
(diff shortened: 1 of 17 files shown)


src/main/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilter.java
```diff
@@ -47,37 +50,44 @@
...
+                    // But the reload is already quite fast anyway. (~1s for 10^4 items with 1 CPU / 2 GB memory
+                    // container)
+                    if(!stopWords.contains(lowerCaseOriginal)) {
+                        ContentMapping mapping = mappings.getMappingOrCreate(name, original -> ContentMapping.of(original, provider.generateFake()));
+                        replacementsMap.put(lowerCaseOriginal, mapping.getReplacement());
+                        trie.add(lowerCaseOriginal);
+                    }
+                }));
+        this.mappingsPattern.set(Pattern.compile("\\b" + trie.getRegex() + "\\b", Pattern.CASE_INSENSITIVE));
+        this.replacementsMap.set(replacementsMap);
```
jglick, line 90: I guess this is somehow dealing with concurrent support bundle generation? Suggests an API mistake somewhere—a context should be passed into methods which require it. Maybe tricky to fix now due to use of the API by other plugins?

src/main/java/com/cloudbees/jenkins/support/filter/WordsTrie.java
```diff
@@ -0,0 +1,171 @@
...
+import java.util.regex.Pattern;
+
+/**
+ * Trie implementation to generate a "Trie regex". A regex that reduces backtracking by following a Trie structure.
+ * <p>
+ * When searching for a match within a list of words, for example ["go", "goes", "going", "gone", "goose"],  a simple
+ * regex that matches any word would typically look like {code}\b(?:(go|goes|going|gone|goose))\b{code}.
+ * <p>
+ * While this works, Such a pattern can be optimized significantly by following a Trie structure in the prefixes such
+ * as {code}\b(?:go(?:(?:es|ing|ne|ose))?)\b{code}.
```
jglick, line 19:
```suggestion
 * When searching for a match within a list of words, for example {@code ["go", "goes", "going", "gone", "goose"]},  a simple
 * regex that matches any word would typically look like {@code\b(?:(go|goes|going|gone|goose))\b}.
 * <p>
 * While this works, Such a pattern can be optimized significantly by following a Trie structure in the prefixes such
 * as {@code\b(?:go(?:(?:es|ing|ne|ose))?)\b}.
```
?

src/main/java/com/cloudbees/jenkins/support/filter/WordsTrie.java
```diff
@@ -0,0 +1,171 @@
...
+ */
+public class WordsTrie {
+
+    /*
+     * As per https://docs.oracle.com/javase/tutorial/essential/regex/literals.html, metacharacters need to be escaped
+     * in the regex String. The regular expression matches any metacharacter occurrence in a String, in which case
+     * the String will need to be quoted.
+     *
+     * NOTE: We could instead choose to escape everything and not do that check. This would increase the size of the
```
jglick, line 28:
```suggestion
     * NOTE: We could instead choose to {@link Pattern#quote} and not do that check. This would increase the size of the
```
?

src/main/java/com/cloudbees/jenkins/support/filter/WordsTrie.java
jglick, line 1: Unfortunate that `Pattern.compile` does not do this sort of optimization internally when given a simple `go|goes|going|gone|goose` pattern. Do you know if that is a filed `OPENJDK-nnnnnnn` RFE?

src/main/java/com/cloudbees/jenkins/support/filter/WordsTrie.java
```diff
@@ -0,0 +1,171 @@
...
+    public void add(String word) {
+        TrieNode ref = root;
+        int i = 0;
+        while (i < word.length() - 1) {
+            // Fill in down the Trie
+            ref = ref.getOrCreate(word.charAt(i), s -> new TrieNode(false));
+            i++;
+        }
+        // We need to mark the last node as an end node
+        ref.getOrCreate(word.charAt(i), s -> new TrieNode(true)).end = true;
```
jglick, line 54: I doubt it matters in this context, but beware emojis which would have multiple `char`s per code point.

src/main/java/com/cloudbees/jenkins/support/filter/WordsTrie.java
```diff
@@ -0,0 +1,171 @@
...
+            if (characters.size() == 1) {
+                // Need to escape special / metacharacters
+                childPatterns.add(quote(String.valueOf(characters.get(0))));
+            } else if (characters.size() > 0) {
+                // Chunking is necessary here to prevent StackOverFlow in pattern matching unions
+                final StringBuilder buf = new StringBuilder();
+                if (characters.size() < 1024) {
+                    buf.append("[");
+                    characters.forEach(character -> buf.append(quote(String.valueOf(character))));
+                    buf.append("]");
```
jglick, line 122: I do not think you can use `Pattern.quote` inside a character class. That would be something like `[\Q?\E]` right? Which I think is a synonym for `[EQ?]` and you want just `[?]`. IIRC only `]` and `\` need to be escaped.

src/main/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilter.java
Dohbedoh: Correct. To deal with concurrent bundle generation. 

The [ContentFilter#reload](https://github.com/jenkinsci/support-core-plugin/blob/master/src/main/java/com/cloudbees/jenkins/support/filter/ContentFilter.java#L62-L66) should be used prior to using the filters. It's always done before starting the bundle generation. So I think this is fine here ?

However, `ThreadLocal` is is a problem when using the filter over remoting. I first thought that a `ThreadLocal` would be well suited, given the ability to process several bundle generation request concurrently. But maybe we should just actually use an `AtomicReference` here. We don't truly need a thread context..

src/main/java/com/cloudbees/jenkins/support/filter/WordsTrie.java
Dohbedoh: Maybe I can write a test for all chars. Like in https://github.com/jenkinsci/support-core-plugin/blob/1283.v9ddb_0284a_00c/src/test/java/com/cloudbees/jenkins/support/util/WordReplacerTest.java#L76-L81.

src/main/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilter.java
jglick: > It's always done before starting the bundle generation.

OK. I guess if it takes no args then it is harmless for it to be called again in the middle of bundle generation from another thread starting another bundle?

src/main/java/com/cloudbees/jenkins/support/filter/WordsTrie.java
Dohbedoh: Good point. Testing this and checking the `Pattern` implementation around Character classes, apparently ` / - [ ] ^ \` must be escaped.

src/main/java/com/cloudbees/jenkins/support/filter/WordsTrie.java
Dohbedoh: I did not find anything about this. But maybe it most effective in specific use case (union of a large list of words).

src/main/java/com/cloudbees/jenkins/support/filter/WordsTrie.java
Dohbedoh: Interesting. Java Pattern with CASE_INSENSITIVE does not hanlde some of those chars. For example the following fails:

```
assertTrue(Pattern.compile("\\baİz\\b", Pattern.CASE_INSENSITIVE).matcher("ai̇z").matches());
```

`İ` has a single char but its lowercase value `i̇` has 2 chars..

Dohbedoh: Eventually found a way to fix this with `UNICODE_CASE` and using the simpler `Matcher#replaceAll`.

src/main/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilter.java
Dohbedoh: Tested multiple generation in parallel and it works. Should I maybe create a test for the SensitiveFilter with parallel threads ?

src/main/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilter.java
```diff
@@ -47,37 +50,44 @@
 @Restricted(NoExternalUse.class)
 public class SensitiveContentFilter implements ContentFilter {
 
+    private static final Logger LOGGER = Logger.getLogger(SensitiveContentFilter.class.getName());
+
+    private final transient ThreadLocal<Pattern> mappingsPattern = new ThreadLocal<>();
+    private final transient ThreadLocal<Map<String, String>> replacementsMap = new ThreadLocal<>();
```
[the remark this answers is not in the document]
Dohbedoh: It is not a persisted configuration. Though it seems to be used over remoting for example at https://github.com/jenkinsci/support-core-plugin/blob/1283.v9ddb_0284a_00c/src/main/java/com/cloudbees/jenkins/support/impl/FileDescriptorLimit.java#L122-L128.

@jglick I am not sure what that would mean then. We should remove transient because of the serialization over remoting ? I have tested with and without and that FileDescriptorLimit component seems to word 🤔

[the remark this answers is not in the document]
jglick: Maybe your testing missed some aspect. #474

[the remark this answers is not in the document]
Dohbedoh: Oh I just realized, all this time I was looking at the file descriptor of the master nodes 🤦 
So yeah apparently this was never working... Can be fixed by either making ContentFilters `Serializable` or `SerializableOnlyOverRemoting`, or doing filtering on the controller side (on the result, but before caching - in `AyncResultCache`).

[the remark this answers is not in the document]
Dohbedoh: Removed the `transient` as suggested.

PR #474, opened 2023-06-15
[JENKINS-71464] Fix `GetUlimit` anonymization

jglick:
If you are unsure how something works, sometimes the answer is that it does not. https://github.com/jenkinsci/support-core-plugin/pull/473#discussion_r1230861903

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/FileDescriptorLimit.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/FileDescriptorLimit.java
@@ -96,8 +96,11 @@ public class FileDescriptorLimit extends Component {
                     out.println("======");
                     out.println();
                     try {
-                        out.println(AsyncResultCache.get(node, fileDescriptorCache, new GetUlimit(filter),
-                            "file descriptor info", "N/A: Either no connection to node or no cached result"));
+                        out.println(ContentFilter.filter( filter,
+                            AsyncResultCache.get(node, fileDescriptorCache, new GetUlimit(),
+                            "file descriptor info",
+                            "N/A: Either no connection to node or no cached result")
+                        ));
                     } catch (IOException e) {
                         Functions.printStackTrace(e, out);
                     } finally {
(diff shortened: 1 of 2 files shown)


Dohbedoh:
So we could either filter on the controller side rather than the agent. Or make the `ContentFilter` implement `SerializableOnlyOverRemoting`. Or both..

Note that [CommandOutputContent](https://github.com/jenkinsci/support-core-plugin/blob/1283.v9ddb_0284a_00c/src/main/java/com/cloudbees/jenkins/support/api/CommandOutputContent.java) is an example of API that help run commands over remoting and the filtering is done on the controller side (see [StringContent](https://github.com/jenkinsci/support-core-plugin/blob/1283.v9ddb_0284a_00c/src/main/java/com/cloudbees/jenkins/support/api/StringContent.java):

* https://github.com/jenkinsci/support-core-plugin/blob/1283.v9ddb_0284a_00c/src/main/java/com/cloudbees/jenkins/support/api/CommandOutputContent.java#L30-L31
* https://github.com/jenkinsci/support-core-plugin/blob/1283.v9ddb_0284a_00c/src/main/java/com/cloudbees/jenkins/support/api/StringContent.java#L59-L60

I think that in most - if not all - other cases, we filter the result on the controller side.

Dohbedoh:
We could also make `ContentFilter` implements `SerializableOnlyOverRemoting` but I am not sure if we really have a need for it - i.e. anonymize the data sent over the channel. 
Both method work. I did notice a performance toll of doing it over remoting (maybe due to serialization, not sure) compare to doing it on the controller side that is much faster.

@jglick let me know what you think.

jglick:
No particular opinion.

src/main/java/com/cloudbees/jenkins/support/impl/FileDescriptorLimit.java
```diff
@@ -96,8 +96,11 @@ protected void printTo(PrintWriter out, ContentFilter filter) {
                     out.println("======");
                     out.println();
                     try {
-                        out.println(AsyncResultCache.get(node, fileDescriptorCache, new GetUlimit(filter),
-                            "file descriptor info", "N/A: Either no connection to node or no cached result"));
+                        out.println(ContentFilter.filter( filter,
+                            AsyncResultCache.get(node, fileDescriptorCache, new GetUlimit(),
```
jglick, line 100: Seems fine.

src/main/java/com/cloudbees/jenkins/support/impl/FileDescriptorLimit.java
```diff
@@ -113,18 +116,16 @@ public static String getUlimit(VirtualChannel channel) throws IOException, Inter
...
+        return ContentFilter.filter(SupportPlugin.getContentFilter().orElse(null),
+            channel.call(new GetUlimit()));
     }
 
     /**
      * * For agent machines.
      */
     private static final class GetUlimit extends MasterToSlaveCallable<String, RuntimeException> {
-        
-        private final ContentFilter filter;
```
jglick, line 124: BTW I wonder why SpotBugs was not flagging this long ago.

PR #485, opened 2023-08-02
[JENKINS-71722][JENKINS-71758] Respect globally configured period for Periodic generation

Dohbedoh:
[JENKINS-71722](https://issues.jenkins.io/browse/JENKINS-71722): global configuration (disablement and period) was not respected in the periodic generation. 

### Testing done

* Start Jenkins
* Configure a recurrence period of 2 hours
* Validate that periodic generation happens every 2 hours

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/config/SupportAutomatedBundleConfiguration.java
+++ b/src/main/java/com/cloudbees/jenkins/support/config/SupportAutomatedBundleConfiguration.java
@@ -22,7 +22,6 @@ import org.kohsuke.stapler.QueryParameter;
 import org.kohsuke.stapler.StaplerRequest;
 import org.kohsuke.stapler.verb.POST;
 
-import edu.umd.cs.findbugs.annotations.NonNull;
 import java.io.IOException;
 import java.util.HashSet;
 import java.util.List;
@@ -94,18 +93,19 @@ public class SupportAutomatedBundleConfiguration extends GlobalConfiguration {
     }
 
     public boolean isEnabled() {
-        return enabled;
+        return !isEnforcedDisabled() && (isEnforcedPeriod() || enabled);
     }
 
     public int getPeriod() {
-        return period;
+        return isEnforcedDisabled() ? 0
+            : Math.max(Math.min(24, isEnforcedPeriod() ? SupportPlugin.AUTO_BUNDLE_PERIOD_HOURS : period), 1);
     }
 
     /**
      * Return if the {@link SupportAutomatedBundleConfiguration#period} is enforced by the System Property
      * {@link SupportPlugin#AUTO_BUNDLE_PERIOD_HOURS}.
      *
-     * @return a list of {@link String}
+     * @return true if enforced by system property, false otherwise
      */
     public boolean isEnforcedPeriod() {
         return System.getProperty(SupportPlugin.class.getName() + ".AUTO_BUNDLE_PERIOD_HOURS") != null;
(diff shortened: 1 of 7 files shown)


jglick (commented):
Looks like it would make the system property not work: https://github.com/jenkinsci/support-core-plugin/blob/305a08ca979bebe14447065d32aa4c589c072f6d/src/main/java/com/cloudbees/jenkins/support/config/SupportAutomatedBundleConfiguration.java#L135-L139 would honor it only if you saved config. I think it is rather https://github.com/jenkinsci/support-core-plugin/blob/305a08ca979bebe14447065d32aa4c589c072f6d/src/main/java/com/cloudbees/jenkins/support/config/SupportAutomatedBundleConfiguration.java#L100-L102 which should load the property if defined?

Dohbedoh:
Right, testing this, the system property is not effective anymore. Maybe we should put the enforcement in the getter rather than the setter.

Dohbedoh:
Disabled that is enforced by `com.cloudbees.jenkins.support.SupportPlugin.AUTO_BUNDLE_PERIOD_HOURS=0`:

enforcedDisable

Period enforced by `com.cloudbees.jenkins.support.SupportPlugin.AUTO_BUNDLE_PERIOD_HOURS=2`:

enforcedPeriod

src/main/java/com/cloudbees/jenkins/support/config/SupportAutomatedBundleConfiguration.java
```diff
@@ -179,13 +175,22 @@ public static List<Component> getApplicableComponents() {
     @POST
     public boolean configure(StaplerRequest req, JSONObject json) throws FormException {
         Jenkins.get().checkPermission(Jenkins.ADMINISTER);
-        if (json.getBoolean("enabled") && !json.has("components")) {
+
+        if(isEnforcedDisabled()) {
+            return true;
```
jglick, line 180: This is going to suppress unrelated config like https://github.com/jenkinsci/support-core-plugin/blob/210c61fc96090e780143d0e55097373fe51be16f/src/main/java/com/cloudbees/jenkins/support/config/SupportAutomatedBundleConfiguration.java#L190?

src/test/java/com/cloudbees/jenkins/support/config/SupportAutomatedBundleConfigurationSystemPropertiesTest.java
```diff
@@ -0,0 +1,110 @@
...
+ */
+public class SupportAutomatedBundleConfigurationSystemPropertiesTest {
+
+    @Rule
+    public RealJenkinsRule rr = new RealJenkinsRule();
+
+    @Test
+    public void enforcePeriod() throws Throwable {
+        rr.javaOptions().javaOptions("-D" + SupportPlugin.class.getName()
+            + ".AUTO_BUNDLE_PERIOD_HOURS=2").then(r -> {
```
jglick, line 30: May not be safe to pass lambdas to `then`. (Serialized into a different JVM.) Generally suggest using a `static` nested class when parameters must be passed, and otherwise (normal case) using a `static` method reference.

Dohbedoh: That explains why I am not able to debug inside the lambda..

src/main/java/com/cloudbees/jenkins/support/config/SupportAutomatedBundleConfiguration.java
Dohbedoh: It leaves the config untouched actually. I tested this.

* If you set `-Dcom.cloudbees.jenkins.support.SupportPlugin.AUTO_BUNDLE_PERIOD_HOURS=0` and then go to the config page and save, it leaves the entire configuration as is (including `com.cloudbees.jenkins.support.config.SupportAutomatedBundleConfiguration.xml`). So that if you then remove the system property, you are back to your original configuration that you had.

* If you set `-Dcom.cloudbees.jenkins.support.SupportPlugin.AUTO_BUNDLE_PERIOD_HOURS=2` and then go to the config page and save, it leaves only the `period` configuration as is (including `com.cloudbees.jenkins.support.config.SupportAutomatedBundleConfiguration.xml`). If you later remove the system property, the generation is still enabled and you are back to your original `period` value.

Does that makes sense ?

jglick: Whatever you say—I am not familiar with the details of this form.

jglick (approved):

PR #496, opened 2023-11-15
Introducing `ComponentCategory.BUILDS`

jglick:
A good place for several Pipeline-specific components (downstream coming).


src/main/java/com/cloudbees/jenkins/support/impl/RunDirectoryComponent.java
```diff
@@ -98,7 +98,7 @@ public boolean understandsSymlink() {
     @NonNull
     @Override
     public ComponentCategory getCategory() {
-        return ComponentCategory.PLATFORM;
+        return ComponentCategory.BUILDS;
```
jglick, line 101: `category` not used in `ObjectComponent`s AFAICT, so this has no visible effect.

julieheard (approved):

Vlatombe (approved):

Dohbedoh (approved):

PR #502, opened 2024-01-15
Use a passthrough filter if Anonymization disabled

Dohbedoh:
* Remove the use of `Optional<ContentFilter>` passed in parameters through the API. Use a `NoneFilter` implementation instead that simply pass through the value. This should simplify the readability, also avoid NPE and remove the need for null checks in some part of the code.
* Deprecated the existing methods `SupportPlugin#getContentFilter` / `SupportPlugin#getContentFilter(ensureLoaded)` in favor of `SupportPlugin#getDefaultContentFilter` / `SupportPlugin#getDefaultContentFilter(ensureLoaded)` and make sure that `ensureLoaded` also reloads the mappings.
* Deprecate `ContentFilter#ensureLoaded` that does not seem useful in favor of just using the existing `ContentFilter#reload`. 
* Also fixed checkstyle in the classes part of the fix

### Testing done

Tested on a test instance with thousands of items and anonymization off then on.

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
+++ b/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
@@ -381,26 +381,20 @@ public class SupportPlugin extends Plugin {
                     CountingOutputStream countingOs = new CountingOutputStream(outputStream);
                     ZipArchiveOutputStream binaryOut =
                             new ZipArchiveOutputStream(new BufferedOutputStream(countingOs, 16384))) {
-                // Get the filter to be used
-                Optional<ContentFilter> maybeFilter = getContentFilter(false);
-                // Recalculate the mappings and stop words and save it to disk
-                maybeFilter.ifPresent(SupportPlugin::reloadAndSaveMappings);
+                ContentFilter filter = getDefaultContentFilter(true);
 
-                // Generate the content of the manifest.md going trough all the components which will be included. It
-                // also returns the contents to include. We pass maybeFilter to filter the names written in the manifest
+                // Generate the content of the manifest.md going through all the components which will be included. It
+                // also returns the contents to include. We pass a filter to filter the names written in the manifest
                 appendManifestHeader(manifest);
                 long startTime = System.currentTimeMillis();
                 List<Content> contents =
-                        appendManifestContents(manifest, errorWriter, components, componentConsumer, maybeFilter);
+                        appendManifestContents(manifest, errorWriter, components, componentConsumer, filter);
                 LOGGER.log(
                         Level.FINE,
                         "Took " + (System.currentTimeMillis() - startTime) + "ms to process all components");
                 contents.add(new UnfilteredStringContent("manifest.md", manifest.toString()));
 
-                Optional<FilteredOutputStream> maybeFilteredOut =
-                        maybeFilter.map(filter -> new FilteredOutputStream(binaryOut, filter));
-                OutputStream textOut =
-                        maybeFilteredOut.map(OutputStream.class::cast).orElse(binaryOut);
+                FilteredOutputStream textOut = new FilteredOutputStream(binaryOut, filter);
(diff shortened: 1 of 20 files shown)


jglick (approved):
Looks simpler.

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -381,26 +381,20 @@ public static void writeBundle(
...
+                        appendManifestContents(manifest, errorWriter, components, componentConsumer, filter);
                 LOGGER.log(
                         Level.FINE,
                         "Took " + (System.currentTimeMillis() - startTime) + "ms to process all components");
                 contents.add(new UnfilteredStringContent("manifest.md", manifest.toString()));
 
-                Optional<FilteredOutputStream> maybeFilteredOut =
-                        maybeFilter.map(filter -> new FilteredOutputStream(binaryOut, filter));
-                OutputStream textOut =
-                        maybeFilteredOut.map(OutputStream.class::cast).orElse(binaryOut);
```
dwnusbaum, line 403: @Dohbedoh For what it's worth, while debugging issues with a log file using a support bundle, after a bit of confusion I think that this PR had the side effect of converting Windows line endings in files in the support bundle to Unix line endings, even when anonymization is disabled, at least in some cases.

I think the key difference is that in the old code, if anonymization was disabled, we just passed everything through via binary streams with no processing. With the new code, even if anonymization is disabled, we pass everything through `FilteredOutputStream`, which decodes the binary data into text, applies the filter, and then re-encodes the text, and I guess somewhere along the way line endings get normalized. Beyond the line endings issue though, I suspect performance might also be worse due to the increased processing overhead when anonymization is disabled.

src/main/java/com/cloudbees/jenkins/support/api/BaseFileContent.java
```diff
@@ -122,8 +123,8 @@ protected void writeTo(OutputStream os) throws IOException {
         }
     }
 
-    protected void writeTo(OutputStream os, ContentFilter filter) throws IOException {
-        if (isBinary || filter == null) {
```
dwnusbaum, line 126: Dropping this condition was actually the reason for https://github.com/jenkinsci/support-core-plugin/pull/502/files#r1491768009, see https://github.com/jenkinsci/support-core-plugin/pull/509.

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
dwnusbaum: The specific issue was actually with `BaseFileContent`, see https://github.com/jenkinsci/support-core-plugin/pull/502/files#r1492687323 and https://github.com/jenkinsci/support-core-plugin/pull/509.

src/main/java/com/cloudbees/jenkins/support/api/BaseFileContent.java
dwnusbaum: Well, only partially. Previously we would have gone through the `writeTo(OutputStream)` overload above instead. If this was the only issue, then we would have noticed the lack of `return` causing most types of file content to be duplicated!

src/main/java/com/cloudbees/jenkins/support/filter/ContentFilter.java
```diff
@@ -76,8 +83,8 @@ default void reload() {}
      * @param text the text to filter
      * @return the text filtered if it is not empty and the filter is not null
      */
-    static String filter(@CheckForNull ContentFilter filter, @CheckForNull String text) {
-        if (filter != null && StringUtils.isNotEmpty(text)) {
+    static String filter(@NonNull ContentFilter filter, @CheckForNull String text) {
+        if (StringUtils.isNotEmpty(text)) {
```
jglick, line 87: https://github.com/jenkinsci/support-core-plugin/pull/522#discussion_r1521811757

PR #509, opened 2024-02-16
Preserve line endings in `FileContent` when anonymization is disabled and avoid duplicating binary content

dwnusbaum:
https://github.com/jenkinsci/support-core-plugin/pull/502 had the side effect of normalizing line endings for all instances of `FileContent` even when anonymization is disabled, which makes it difficult to use support bundles to analyze issues where exact file contents are significant (e.g. build logs and associated metadata).

While looking at that issue, I _also_ noticed that any kind of `FileContent` that was detected as binary was actually written twice, once as binary and then once as text. Oops! From a quick look I think this has been broken since https://github.com/jenkinsci/support-core-plugin/pull/174.

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/api/BaseFileContent.java
+++ b/src/main/java/com/cloudbees/jenkins/support/api/BaseFileContent.java
@@ -124,8 +124,9 @@ class BaseFileContent {
     }
 
     protected void writeTo(OutputStream os, @NonNull ContentFilter filter) throws IOException {
-        if (isBinary) {
+        if (isBinary || filter == ContentFilter.NONE) {
             writeTo(os);
+            return;
         }
 
         try {
--- a/src/test/java/com/cloudbees/jenkins/support/api/FileContentTest.java
+++ b/src/test/java/com/cloudbees/jenkins/support/api/FileContentTest.java
@@ -26,9 +26,12 @@ package com.cloudbees.jenkins.support.api;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 
+import com.cloudbees.jenkins.support.filter.ContentFilter;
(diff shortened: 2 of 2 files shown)


src/main/java/com/cloudbees/jenkins/support/api/BaseFileContent.java
```diff
@@ -124,8 +124,9 @@ protected void writeTo(OutputStream os) throws IOException {
     }
 
     protected void writeTo(OutputStream os, @NonNull ContentFilter filter) throws IOException {
-        if (isBinary) {
+        if (isBinary || filter == ContentFilter.NONE) {
             writeTo(os);
+            return;
```
dwnusbaum, line 129: Drop the `return` and `preservesLineEndingsWithContentFilterNone` starts failing because the content is written twice.

src/main/java/com/cloudbees/jenkins/support/api/BaseFileContent.java
```diff
@@ -124,8 +124,9 @@ protected void writeTo(OutputStream os) throws IOException {
     }
 
     protected void writeTo(OutputStream os, @NonNull ContentFilter filter) throws IOException {
-        if (isBinary) {
+        if (isBinary || filter == ContentFilter.NONE) {
```
dwnusbaum, line 127: If we are using `ContentFilter.NONE`, just write everything directly as binary.

It would be nice to preserve newlines in all cases, but the use of `BufferedReader.readLine` below makes that difficult to do, and anyways, once anonymization is involved, any debugging that cares about byte-level details is not going work in the first place so I don't think it's that important.

Dohbedoh (approved):
thanks @dwnusbaum

PR #511, opened 2024-02-28
Replace class restriction to method levels of ContentMappings

Dohbedoh:
Trying to leverage the existing Anonymization stack of Support Core from an external plugin, I noticed some problems. `ContentFilter`s may generate mappings when filtering or when reloading, so when generating the content with anonymization, one need access to the singleton to avoid performance problems:

* https://github.com/jenkinsci/support-core-plugin/blob/c7f0a46f9e6cf50656fc46d209e9aa886fc3880c/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java#L380
* https://github.com/jenkinsci/support-core-plugin/blob/c7f0a46f9e6cf50656fc46d209e9aa886fc3880c/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java#L555-L556

Therefore:

* I am proposing to relax access to the Singleton, while still restricting most methods.

Note: I also think that we would eventually externalize the anonymization stack in a module / plugin.

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/filter/ContentFilter.java
+++ b/src/main/java/com/cloudbees/jenkins/support/filter/ContentFilter.java
@@ -26,13 +26,34 @@ package com.cloudbees.jenkins.support.filter;
 
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
+import hudson.BulkChange;
 import hudson.ExtensionList;
 import hudson.ExtensionPoint;
+import hudson.model.Saveable;
+import java.io.IOException;
 import org.apache.commons.lang.StringUtils;
 
 /**
  * Provides a strategy to filter support bundle written contents. This is primarily useful to anonymize data written
  * to the bundle, though more complex filtering can be achieved.
+ * A {@link ContentFilter} may produce {@link ContentMappings} during {@link ContentFilter#reload()} and
+ * {@link ContentFilter#filter(String)} that are automatically persisted. It is therefore important to use
+ * {@link BulkChange} when reloading the filter as well as when launching a task that performs the filtering:
+ *
+ * <pre>
+ * {@code
+ *   ContentFilter filter = ContentFilter.ALL;
+ *   try {
+ *     ContentFilter.reloadAndSaveMappings(filter);
+ *   } catch (IOException e) {
+ *      ...
+ *   }
+ *   [...]
+ *   try (BulkChange change = new BulkChange(ContentFilter.bulkChangeTarget())) {
+ *     doSomething(filter);
+ *     change.commit();
+ *   }
+ * }</pre>
  *
  * @since TODO
  */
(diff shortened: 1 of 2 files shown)


jglick:
Maybe cleaner to add to `ContentFilter` something like:

```java
static Saveable bulkChangeTarget() {
    return ContentMappings.get();
}
```

Dohbedoh:
Makes sense to me. We'd also to externally be able to reload the mappings (`ContentMappings#reload`). Maybe this can also be provided through a static method in `ContentFilter`. Will propose something.

src/main/java/com/cloudbees/jenkins/support/filter/ContentFilter.java
```diff
@@ -26,13 +26,33 @@
...
+ * <p><blockquote><pre>
+ *   ContentFilter filter = ContentFilter.ALL;
+ *   try {
+ *     ContentFilter.reloadAndSaveMappings(filter);
+ *   } catch (IOException e) {
+ *      ...
+ *   }
+ *   [...]
+ *   try (BulkChange change = new BulkChange(ContentFilter.bulkChangeTarget())) {
+ *     doSomething(filter)
```
jglick, line 52:
```suggestion
 *     doSomething(filter);
```

src/main/java/com/cloudbees/jenkins/support/filter/ContentFilter.java
```diff
@@ -90,4 +110,33 @@ static String filter(@NonNull ContentFilter filter, @CheckForNull String text) {
             return text;
         }
     }
+
+    /**
+     * Access the {@link ContentMappings} as Saveable.
```
jglick, line 115: Not good form to link to something restricted, even though Javadoc will allow it since it knows nothing of `@Restricted` which is basically a workaround for lack of Java Module System restrictions. Better to discuss in terms of what this does.

src/main/java/com/cloudbees/jenkins/support/filter/ContentFilter.java
```diff
@@ -90,4 +110,33 @@ static String filter(@NonNull ContentFilter filter, @CheckForNull String text) {
...
+        return ContentMappings.get();
+    }
+
+    /**
+     * Reloads the state of this filter and commit .
+     *
+     * @throws IOException if reload or saving failed
+     */
+    static void reloadAndSaveMappings(@NonNull ContentFilter filter) throws IOException {
+        ContentMappings mappings = (ContentMappings) bulkChangeTarget();
```
jglick, line 129:
```suggestion
        ContentMappings mappings = ContentMappings.get();
```
seems clearer.

jglick (approved):
Looks fine. (I am not familiar with the API so I cannot comment on whether its usage would be appropriate.)

jglick (approved):

Dohbedoh:
I'm gonna rebase, push and merge.

PR #515, opened 2024-02-29
Update baseline to 2.414.x

jglick:
The oldest suggested baseline according to https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/#currently-recommended-versions

PR #517, opened 2024-02-29
Overhaul of `SlaveLaunchLogs`

jglick:
The collection of launch logs from agents is rather antiquated given the prevalence of `Cloud`s and is missing a lot of critical diagnostics. (CloudBees-internal reference: BEE-44099) Some observations:

`SlaveLaunchLogs` claims to include logs from ephemeral agents but does not. Perhaps it once did, but https://github.com/jenkinsci/jenkins/pull/1681 broke this: `SlaveComputer.kill` deletes `Computer.logDir` when the computer is removed. (Despite the appearance in the executor widget, this does not happen merely because an agent is disconnected, so `SlaveLaunchLogs` will still include an entry for that agent while offline.) Not so easy to work around because this happens from `Jenkins.updateComputerList`, before `NodeListener.fireOnDeleted` is fired, and potentially before or concurrently with `ComputerListener.onOffline`.

`SlaveLaunchLogs` does not honor `SafeTimerTask.getLogsRoot()`. (This is set by CloudBees CI HA controllers.)

These issues can be solved in a straightforward way by just collecting _all_ agent launch logs in one (rotated) log file, with every line prefixed by a timestamp and agent name. Downstream of https://github.com/jenkinsci/jenkins/pull/9009 (without which the new system for log collection ~cannot work~ cannot work so easily for inbound agents).

Needs to be squash-merged since things went back and forth quite a bit here.


Vlatombe (commented):
Recording review

src/main/java/com/cloudbees/jenkins/support/impl/SlaveLaunchLogs.java
```diff
@@ -73,128 +86,80 @@ public String getDisplayName() {
...
-        /**
-         * If the file is more than 7 days old, it is considered too old.
-         */
-        public boolean isTooOld() {
-            return time < System.currentTimeMillis() - TimeUnit.DAYS.toMillis(7);
+        protected void eol(byte[] b, int len) throws IOException {
+            synchronized (out) {
+                out.write('[');
+                out.write(DateTimeFormatter.ISO_INSTANT
+                        .format(Instant.now().truncatedTo(ChronoUnit.MILLIS))
```
jglick, line 155: See https://github.com/jenkinsci/lib-support-log-formatter/issues/64 for other logs.

pom.xml
```diff
@@ -67,7 +67,8 @@
 
   <properties>
     <changelist>999999-SNAPSHOT</changelist>
-    <jenkins.version>2.361.4</jenkins.version>
+    <!-- TODO https://github.com/jenkinsci/jenkins/pull/9009 -->
+    <jenkins.version>2.448-rc34683.5cd5ff72b_b_89</jenkins.version>
```
jglick, line 71: see also #515

Vlatombe (approved):

Vlatombe (approved):

Dohbedoh (approved):

PR #519, opened 2024-03-08
Define `Container.getComponents`

jglick:
A special component defined in CloudBees CI in high availability mode gathers information from other controller replicas. To make this properly reflect the admin’s selection it is necessary to expose that selection to other components. This information was not previously available, though it is easy to pass.


Vlatombe (approved):

Dohbedoh (approved):
I think it is a good addition indeed.

PR #522, opened 2024-03-12
Trying to deflake `CheckFilterTest`

jglick:
https://github.com/jenkinsci/support-core-plugin/runs/22561627825 is not reproducibe locally but has also been observed in PCT (@cloudbees reference BEE-10081). The failure does not make much sense because `AsyncResultCache.get` should always succeed on the built-in node (`Jenkins`). Locally:

```
FINER	c.c.j.support.AsyncResultCache#get: com.cloudbees.jenkins.support.impl.NetworkInterfaces$GetNetworkInterfaces@6ba48c4d on master succeeded
FINER	c.c.j.support.AsyncResultCache#get: com.cloudbees.jenkins.support.impl.NetworkInterfaces$GetNetworkInterfaces@35c553e2 on agent0 succeeded
```

PR #523, opened 2024-03-14
NPE in `AboutJenkins`

jglick:
Split from https://github.com/jenkinsci/support-core-plugin/pull/522#discussion_r1521707377 due to prolonged discussion with @jtnord & @Dohbedoh.

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
@@ -235,7 +235,7 @@ public class AboutJenkins extends Component {
          * @return the Java information.
          */
         @SuppressFBWarnings("DMI_HARDCODED_ABSOLUTE_FILENAME")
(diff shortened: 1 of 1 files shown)


src/main/java/com/cloudbees/jenkins/support/impl/AboutJenkins.java
```diff
@@ -439,7 +439,7 @@ private String getInfo(ContentFilter filter) {
                         .append(" arg[")
                         .append(count++)
                         .append("]: `")
-                        .append(Markdown.escapeBacktick(ContentFilter.filter(filter, arg)))
+                        .append(Markdown.escapeBacktick(filter != null ? ContentFilter.filter(filter, arg) : arg))
```
jtnord, line 442: previous discussion in https://github.com/jenkinsci/support-core-plugin/pull/522#discussion_r1521707377

Dohbedoh (approved):

PR #525, opened 2024-03-18
Fixes to `RunDirectoryComponent.excludes`

jglick:
* **Build Support** excludes needlessly looked in subdirectories for things like `stashes`.
* `artifacts` is not a directory name; it is `archive`. This particular important exclusion lacked test coverage. (Be careful not to mix up directory names with URL components, which are often subtly different: `archive` vs. `artifact`; `jobs` vs. `job`.)
* Default excludes were missing several kinds of directories that I have observed containing many (and often large) files in builds running on @cloudbees servers.
* **Project Support** excludes were mixing two sorts of entries: subitems; and standard excludes for builds. More clearly separate them and refine the patterns.
* Some help files displayed `{0}` which was apparently meant to be subtituted with something but was not.
* Links to Apache Ant references were light on detail.
* Switching to `xxx/` as a shortcut for `xxx/**`.
* `AbstractItemComponentTest` was not using a standard naming convention.

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/AbstractItemDirectoryComponent.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/AbstractItemDirectoryComponent.java
@@ -12,7 +12,10 @@ import hudson.util.FormValidation;
 import java.io.File;
 import java.io.IOException;
 import java.io.PrintWriter;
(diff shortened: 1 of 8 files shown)


Dohbedoh (approved):
thanks!

PR #542, opened 2024-05-20
[chore] No point for CustomLogs to be enabled by default

Dohbedoh:
### Submitter checklist
- [ ] Please describe what you did


jglick:
OK I guess? Not sure why particular components are marked to be enabled by default and others are not; I would tend to think that if you went to the trouble of setting up a custom log recorder, it is probably because these messages were important to review, and they are possibly the _most_ interesting things in a bundle.

BTW `chore` is the wrong label for a user-visible behavioral change. Use either `enhancement` or `bug`.

Dohbedoh:
It thought it could help to prevent scenario where someone generate bundle and share it somewhere without paying attention of currently enabled customer loggers. Though they should since it is enabled by default. And some logger may log sensitive info - typically http client internal packages that print headers.
I know this is not solving the source of that problem but may as well alleviate it.

jglick (approved):

PR #338, opened 2022-02-23
Fix LogRecorder usage for newer core versions

raul-arabaolaza:
There is some issues on migration code from https://github.com/jenkinsci/jenkins/pull/4538 see https://github.com/jenkinsci/github-branch-source-plugin/pull/485 for full details.

This PR updates the plugin usages of affected code to make it work again and updates the baseline. Issues can be demonstrated by `mvn test -Djenkins.version=2.332.1-cb-3 -Dtest=CustomLogsTest#testCustomLogsContent -Denforcer.skip -Daccess-modifier-checker.skip` 

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #556, opened 2024-07-04
[JENKINS-73402] Load LogManager recorder afetr CasC loading

Dohbedoh:
Was probably introduced by https://github.com/jenkinsci/support-core-plugin/pull/338.

Propose to load the recorders after CasC init (see https://github.com/jenkinsci/configuration-as-code-plugin/blob/1810.v9b_c30a_249a_4c/plugin/src/main/java/io/jenkins/plugins/casc/ConfigurationAsCode.java#L337). 

I thought about just always targeting the reference `Jenkins.get().getLog().getRecorders()` from [CustomHandler#publish](https://github.com/jenkinsci/support-core-plugin/blob/1459.va_f527ce9a_d64/src/main/java/com/cloudbees/jenkins/support/impl/CustomLogs.java#L166) but wonder if that would be overkill ? We go through this for every log record.. cc @jglick I wonder WDYT about this ?

### Testing done

* Start Jenkins with CasC:
 
```
jenkins:
  log:
    recorders:
    - loggers:
      - level: "ALL"
        name: "jenkins.InitReactorRunner"
      name: "reactor"
```

* verify that `$JENKINS_HOME/logs/custom/reactor.log` exists

### Submitter checklist

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/CustomLogs.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/CustomLogs.java
@@ -133,7 +133,7 @@ public class CustomLogs extends Component {
         }
     }
 
-    @Initializer(after = InitMilestone.EXTENSIONS_AUGMENTED, before = InitMilestone.SYSTEM_CONFIG_LOADED)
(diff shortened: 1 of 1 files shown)


jglick (approved):
Seems OK.

PR #558, opened 2024-07-10
THREAD_DUMPS_ON_SLOW_REQUEST: adding the base of the development of t…

ironcerocloudbees:
…his feature

### Testing done

I tested this feature installing the snapshot version into a testing environment (in the last `2.452.2` LTS version) and run the following tests:

1. The way to trigger the slow request was to run `sleep(60000)` groovy code in the `Manage Jenkins -> Script Console` in multiples tabs at the same time (around 10).
2. I ran the above trigger a few times consecutively to confirm that we don't trigger the collecting operation while the `RECURRENCE_PERIOD_MIN` is working.
3. I run the above trigger each 12-15 minutes for more than 2 hours to confirm that the `SLOW_REQUEST_THREAD_DUMPS_TO_RETAIN` function works propertly.
4. I changed the parameters (`RECURRENCE_PERIOD_MIN`, `MINIMAL_SLOW_REQUEST_COUNT`, etc) in the Jenkins startup to confirm that we can modify any of those parameters and the change affected the behavior of the new feature in the correct way.
5. I generated a new support bundle and confirmed that the new files were included correctly.

Proof of the feature working:

image

UI changes:

image

JIRA:

https://issues.jenkins.io/browse/JENKINS-73431

### Submitter checklist
- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/slowrequest/SlowRequestChecker.java
+++ b/src/main/java/com/cloudbees/jenkins/support/slowrequest/SlowRequestChecker.java
@@ -1,5 +1,8 @@
 package com.cloudbees.jenkins.support.slowrequest;
 
+import static com.cloudbees.jenkins.support.slowrequest.SlowRequestThreadDumpsGenerator.MINIMAL_SLOW_REQUEST_COUNT;
+import static java.util.logging.Level.WARNING;
+
 import com.cloudbees.jenkins.support.SupportPlugin;
 import com.cloudbees.jenkins.support.filter.ContentFilter;
 import com.cloudbees.jenkins.support.timer.FileListCap;
@@ -18,6 +21,8 @@ import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.TimeZone;
 import java.util.concurrent.TimeUnit;
+import java.util.logging.Logger;
+
 import jenkins.model.Jenkins;
 
 /**
@@ -81,6 +86,8 @@ public class SlowRequestChecker extends PeriodicWork {
         // We filter the information written to the slow-requests files
         ContentFilter contentFilter = SupportPlugin.getDefaultContentFilter();
 
+        int slowRequestCount = 0;
+
         for (InflightRequest req : filter.tracker.values()) {
             long totalTime = now - req.startTime;
 
@@ -89,6 +96,9 @@ public class SlowRequestChecker extends PeriodicWork {
                 if (req.ended) continue;
 
                 boolean newRecord = req.record == null;
+
+                slowRequestCount++;
+
                 if (newRecord) {
                     req.record = logs.file(format.format(new Date(iota++)) + ".txt");
                     logs.add(req.record);
@@ -122,6 +132,22 @@ public class SlowRequestChecker extends PeriodicWork {
                 }
             }
         }
+
+        if (slowRequestCount >= MINIMAL_SLOW_REQUEST_COUNT) {
+            boolean newThreadDumps = SlowRequestThreadDumpsGenerator.checkThreadDumpsTrigger(iota);
+
+            if (newThreadDumps){
+                try {
+                    SlowRequestThreadDumpsGenerator slowRequestThreadDumpsGenerator = new SlowRequestThreadDumpsGenerator(iota);
+                    slowRequestThreadDumpsGenerator.start();
+                } catch(Exception e){
+                    LOGGER.log(WARNING,
+                            "Support Core plugin can't throw a new thread to collect thread dumps under SlowRequest scenario",
+                            e);
+                }
+            }
+        }
+
     }
 
     private void printThreadStackElements(ThreadInfo threadinfo, PrintWriter writer, ContentFilter contentFilter) {
@@ -129,4 +155,6 @@ public class SlowRequestChecker extends PeriodicWork {
             writer.println("    " + contentFilter.filter(element.toString()));
         }
     }
+
+    private static final Logger LOGGER = Logger.getLogger(SlowRequestChecker.class.getName());
 }
--- a/src/main/java/com/cloudbees/jenkins/support/slowrequest/SlowRequestChecker.java
+++ b/src/main/java/com/cloudbees/jenkins/support/slowrequest/SlowRequestChecker.java
@@ -22,7 +22,6 @@ import java.util.Date;
 import java.util.TimeZone;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Logger;
-
 import jenkins.model.Jenkins;
 
 /**
@@ -136,18 +135,19 @@ public class SlowRequestChecker extends PeriodicWork {
         if (slowRequestCount >= MINIMAL_SLOW_REQUEST_COUNT) {
             boolean newThreadDumps = SlowRequestThreadDumpsGenerator.checkThreadDumpsTrigger(iota);
 
-            if (newThreadDumps){
+            if (newThreadDumps) {
                 try {
-                    SlowRequestThreadDumpsGenerator slowRequestThreadDumpsGenerator = new SlowRequestThreadDumpsGenerator(iota);
+                    SlowRequestThreadDumpsGenerator slowRequestThreadDumpsGenerator =
+                            new SlowRequestThreadDumpsGenerator(iota);
                     slowRequestThreadDumpsGenerator.start();
-                } catch(Exception e){
-                    LOGGER.log(WARNING,
+                } catch (Exception e) {
+                    LOGGER.log(
+                            WARNING,
                             "Support Core plugin can't throw a new thread to collect thread dumps under SlowRequest scenario",
                             e);
                 }
             }
         }
-
     }
 
     private void printThreadStackElements(ThreadInfo threadinfo, PrintWriter writer, ContentFilter contentFilter) {
--- a/src/main/java/com/cloudbees/jenkins/support/slowrequest/SlowRequestThreadDumpsGenerator.java
+++ b/src/main/java/com/cloudbees/jenkins/support/slowrequest/SlowRequestThreadDumpsGenerator.java
@@ -9,6 +9,7 @@ import hudson.Extension;
 import java.io.*;
 import java.text.SimpleDateFormat;
 import java.util.Date;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Logger;
 
 /**
@@ -27,7 +28,7 @@ public class SlowRequestThreadDumpsGenerator extends Thread {
     /**
      * Semaphore to ensure that only one instance is collecting data at the same time.
      */
-    private static boolean running = false;
+    private static AtomicBoolean running = new AtomicBoolean(false);
 
     /**
      * How often (at minimum) we will capture the ThreadDump under a slowRequest scenario.
@@ -134,17 +135,11 @@ public class SlowRequestThreadDumpsGenerator extends Thread {
     }
 
     private static boolean isRunning() {
-        boolean runningLocal;
-        synchronized (SlowRequestThreadDumpsGenerator.class) {
-            runningLocal = running;
-        }
-        return runningLocal;
+        return running.get();
     }
 
     private static void setRunningStatus(boolean runningLocal) {
-        synchronized (SlowRequestThreadDumpsGenerator.class) {
-            running = runningLocal;
-        }
+        running.set(runningLocal);
     }
 
     private static final Logger LOGGER = Logger.getLogger(SlowRequestThreadDumpsGenerator.class.getName());
--- a/src/main/java/com/cloudbees/jenkins/support/slowrequest/SlowRequestThreadDumpsGenerator.java
+++ b/src/main/java/com/cloudbees/jenkins/support/slowrequest/SlowRequestThreadDumpsGenerator.java
@@ -22,7 +22,7 @@ public class SlowRequestThreadDumpsGenerator extends Thread {
     /**
      * When was the latest ThreadDumps generation in milliseconds
      */
-    public static long latestGeneratedSlowRequestThreadDump = 0l;
+    private static long latestGeneratedSlowRequestThreadDump = 0l;
 
     /**
      * How often (at minimum) we will capture the ThreadDump under a slowRequest scenario.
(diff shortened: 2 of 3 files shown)


jglick:
The slow request records component already includes thread dumps, at least for the thread handling the HTTP request (and IIRC for any thread it is blocked on as well); what additional value comes from this? I did not follow the motivation in Jira.

jglick:
Just curious. Leaving review to @Dohbedoh.

src/main/java/com/cloudbees/jenkins/support/slowrequest/SlowRequestThreadDumpsGenerator.java
```diff
@@ -0,0 +1,121 @@
...
+     * Limit the number of thread dumps to retain on slowRequest scenario
+     */
+    public static final int SLOW_REQUEST_THREAD_DUMPS_TO_RETAIN = Integer.getInteger(
+            SlowRequestThreadDumpsGenerator.class.getName() + ".SLOW_REQUEST_THREAD_DUMPS_TO_RETAIN", 40);
+
+    /**
+     * Thread dumps generated on slowRequest scenario are stored in $JENKINS_HOME/slow-request-threaddumps
+     */
+    protected final FileListCap logs = new FileListCap(
+            new File(Jenkins.get().getRootDir(), "slow-request-threaddumps"), SLOW_REQUEST_THREAD_DUMPS_TO_RETAIN);
```
Dohbedoh, line 67: I thought that maybe we should use the `$JENKINS_HOME/support` directory (I know that historically we added slow-requests and deadlocks files at `$JENKINS_HOME` which I ought to eventually move into the support directory instead):

```suggestion
            new File(SupportPlugin.getRootDirectory(), "slow-request-threaddumps"), SLOW_REQUEST_THREAD_DUMPS_TO_RETAIN);
```

src/main/java/com/cloudbees/jenkins/support/slowrequest/SlowRequestThreadDumpsGenerator.java
```diff
@@ -0,0 +1,121 @@
...
+            Integer.getInteger(SlowRequestThreadDumpsGenerator.class.getName() + ".FREQUENCY_SEC", 5);
+
+    /**
+     * Limit the number of thread dumps to retain on slowRequest scenario
+     */
+    public static final int SLOW_REQUEST_THREAD_DUMPS_TO_RETAIN = Integer.getInteger(
+            SlowRequestThreadDumpsGenerator.class.getName() + ".SLOW_REQUEST_THREAD_DUMPS_TO_RETAIN", 40);
+
+    /**
+     * Thread dumps generated on slowRequest scenario are stored in $JENKINS_HOME/slow-request-threaddumps
```
Dohbedoh, line 64:
```suggestion
     * Thread dumps generated on slowRequest scenario are stored in $JENKINS_HOME/support/slow-request-threaddumps
```

src/main/java/com/cloudbees/jenkins/support/slowrequest/SlowRequestThreadDumpsGenerator.java
```diff
@@ -0,0 +1,121 @@
...
+     */
+    protected final FileListCap logs = new FileListCap(
+            new File(Jenkins.get().getRootDir(), "slow-request-threaddumps"), SLOW_REQUEST_THREAD_DUMPS_TO_RETAIN);
+
+    /**
+     * Provide a means to disable the slow request thread dump checker. This is a volatile non-final field as if you run into
+     * issues in a running Jenkins you may need to disable without restarting Jenkins.
+     */
+    public static volatile boolean DISABLED =
```
Dohbedoh, line 73: Do we really need to make it `volatile` here ? Seems to me that only a single instance of this threqad shall be running but maybe I am missing something 🤔 .

src/main/java/com/cloudbees/jenkins/support/slowrequest/SlowRequestThreadDumpsGenerator.java
```diff
@@ -0,0 +1,121 @@
...
+                logs.add(threadDumpFile);
+            } catch (IOException ioe) {
+                LOGGER.log(
+                        WARNING,
+                        "Support Core plugin can't generate automatically thread dumps under SlowRequest scenario",
+                        ioe);
+            } finally {
+                fileNameDate += FREQUENCY_SEC * 1000;
+            }
+        }
```
Dohbedoh, line 107: Hmm maybe I am missing something here, but we never actually sleep or reschedule anything here, so we are actually generating `TOTAL_ITERATIONS` thread dumps with different file names in a loop, not every `FREQUENCY_SEC * 1000` seconds ?

src/main/java/com/cloudbees/jenkins/support/slowrequest/SlowRequestChecker.java
```diff
@@ -122,11 +131,30 @@ protected void doRun() throws Exception {
...
         }
+
+        if (slowRequestCount >= MINIMAL_SLOW_REQUEST_COUNT) {
+            boolean newThreadDumps = SlowRequestThreadDumpsGenerator.checkThreadDumpsTrigger(iota);
+
+            if (newThreadDumps) {
+                try {
+                    SlowRequestThreadDumpsGenerator slowRequestThreadDumpsGenerator =
+                            new SlowRequestThreadDumpsGenerator(iota);
+                    slowRequestThreadDumpsGenerator.start();
```
Dohbedoh, line 142: We are blocking the `SlowRequestChecker` here for the entire time that the `SlowRequestThreadDumpsGenerator` shall run. Which is by default 5s * 4, so 20s. Way over the 3s recurrence period of the checker. Maybe we need to launch this asynchronously or something..

src/main/java/com/cloudbees/jenkins/support/slowrequest/SlowRequestThreadDumpsGenerator.java
Dohbedoh: Yeah definitely generating dump without the 5 sec interval:

```
% ll /tmp/je-1-home/slow-request-threaddumps
total 436
drwxr-xr-x  6 jenkins wheel    192 Jul 19 19:58 .
drwxr-xr-x 53 jenkins wheel   1696 Jul 19 19:56 ..
-rw-r--r--  1 jenkins wheel 120211 Jul 19 19:58 20240719-195806.303.txt
-rw-r--r--  1 jenkins wheel 105633 Jul 19 19:58 20240719-195811.303.txt
-rw-r--r--  1 jenkins wheel 106723 Jul 19 19:58 20240719-195816.303.txt
-rw-r--r--  1 jenkins wheel 105673 Jul 19 19:58 20240719-195821.303.txt
% stat /tmp/je-1-home/slow-request-threaddumps/20240719-195806.303.txt
16777233 9837190 -rw-r--r-- 1 jenkins wheel 0 120211 "Jul 19 19:58:06 2024" "Jul 19 19:58:06 2024" "Jul 19 19:58:06 2024" "Jul 19 19:58:06 2024" 4096 240 0 /tmp/je-1-home/slow-request-threaddumps/20240719-195806.303.txt
% stat /tmp/je-1-home/slow-request-threaddumps/20240719-195811.303.txt
16777233 9837191 -rw-r--r-- 1 jenkins wheel 0 105633 "Jul 19 19:58:07 2024" "Jul 19 19:58:06 2024" "Jul 19 19:58:06 2024" "Jul 19 19:58:06 2024" 4096 208 0 /tmp/je-1-home/slow-request-threaddumps/20240719-195811.303.txt
% stat /tmp/je-1-home/slow-request-threaddumps/20240719-195816.303.txt
16777233 9837194 -rw-r--r-- 1 jenkins wheel 0 106723 "Jul 19 19:58:07 2024" "Jul 19 19:58:06 2024" "Jul 19 19:58:06 2024" "Jul 19 19:58:06 2024" 4096 216 0 /tmp/je-1-home/slow-request-threaddumps/20240719-195816.303.txt
% stat /tmp/je-1-home/slow-request-threaddumps/20240719-195821.303.txt
16777233 9837195 -rw-r--r-- 1 jenkins wheel 0 105673 "Jul 19 19:58:08 2024" "Jul 19 19:58:07 2024" "Jul 19 19:58:07 2024" "Jul 19 19:58:06 2024" 4096 208 0 /tmp/je-1-home/slow-request-threaddumps/20240719-195821.303.txt
```

Dohbedoh (changes requested):
Some need to be addressed:

* the `SlowRequestThreadDumpsGenerator` actually generates thread dumps consecutively without intervals and give a misleading name to the files.
* store the data under the `$JENKINS_HOME/support/` directory.

Another thing that bugs me is that if we tweak the settings, we could theoretically have several `SlowRequestThreadDumpsGenerator` threads running (when iterations x frequency > recurrence period) and they would generate dumps. redundantly. And that make me feel like the generator should probably be a singleton that manage and launch **the** task.

src/main/java/com/cloudbees/jenkins/support/slowrequest/SlowRequestThreadDumpsGenerator.java
ironcerocloudbees: You are totally right. I just revisited the code, and as you said, we don't need the `volatile` attribute here. The expectation is that only one thread will be running simultaneously. I will apply the change in my next commit.

src/main/java/com/cloudbees/jenkins/support/slowrequest/SlowRequestThreadDumpsGenerator.java
ironcerocloudbees: Holy sh.. You are right. I forgot to add the sleeping call to wait for this `FREQUENCY_SEC * 1000` seconds. As you said, currently, we are creating all those files at the same time. 

I will add the `sleep` instruction in the `finally` block.

ironcerocloudbees:
> Another thing that bugs me is that if we tweak the settings, we could theoretically have several SlowRequestThreadDumpsGenerator threads running (when iterations x frequency > recurrence period) and they would generate dumps. redundantly. And that make me feel like the generator should probably be a singleton that manage and launch the task.

First of all, thank @Dohbedoh so much for your review. Regarding this last question, I trust your criteria better than mine. However, I would like to talk about the matter briefly before making the change.

As you said, we could tune the variables to have two threads running simultaneously, and we should try to avoid it. However, I am afraid of making the class singleton. We are saving files during the thread execution and we are using the Java tool to do it. So, in case other threads affect the performance of those libraries (or it was blocking part of that code), would it not cause any kind of deadlock (or at least a performance degradation)? 

Knowing that the only benefit of the `singleton` here is to avoid the execution of two threads due to an incongruent variables setup, it would not be better if we check the assigned values of the variables and establish minimum values.

Kind regards,

Dohbedoh:
@ironcerocloudbees What I mean by singleton is a some kind of "Holder" for the generator Thread. To avoid creating new a `SlowRequestThreadDumpsGenerator` threads if one of them is actually still running. The condition whether we want to launch a generator thread (i.e. `checkThreadDumpsTrigger`) should not just be based on recurrence period but also on whether a dump generator is already running.

ironcerocloudbees:
Hi @Dohbedoh,

Thank you again for your time here. I pushed a new version using a semaphore to avoid concurrencies. I have been reviewing different implementations of the Singleton for threads, and the semaphore looks a good option for us. If you could review it it would be great.

Kind regards,

src/main/java/com/cloudbees/jenkins/support/slowrequest/SlowRequestThreadDumpsGenerator.java
```diff
@@ -110,17 +117,35 @@ public void run() {
...
+            runningLocal = running;
+        }
+        return runningLocal;
+    }
+
+    private static void setRunningStatus(boolean runningLocal) {
+        synchronized (SlowRequestThreadDumpsGenerator.class) {
+            running = runningLocal;
+        }
+    }
```
Dohbedoh, line 148: Could be simplified using `AtomicBoolean`.

Dohbedoh (approved):
Thanks. That looks good. Maybe an `AtomicBoolean` is better suited for the `running`. That'd be my only suggestion here.

src/main/java/com/cloudbees/jenkins/support/slowrequest/SlowRequestThreadDumpsGenerator.java
```diff
@@ -0,0 +1,151 @@
...
+     * Thread dumps generated on slowRequest scenario are stored in $JENKINS_HOME/support/slow-request-threaddumps
+     */
+    protected final FileListCap logs = new FileListCap(
+            new File(SupportPlugin.getRootDirectory(), "slow-request-threaddumps"),
+            SLOW_REQUEST_THREAD_DUMPS_TO_RETAIN);
+
+    /**
+     * Provide a means to disable the slow request thread dump checker.
+     */
+    public static boolean DISABLED = Boolean.getBoolean(SlowRequestThreadDumpsGenerator.class.getName() + ".DISABLED");
```
Dohbedoh, line 78: Per spotbugs:

```suggestion
    public static final boolean DISABLED = Boolean.getBoolean(SlowRequestThreadDumpsGenerator.class.getName() + ".DISABLED");
```

src/main/java/com/cloudbees/jenkins/support/slowrequest/SlowRequestThreadDumpsGenerator.java
ironcerocloudbees: Hi @Dohbedoh,

Thank you for the tip. I have never used AtomicBoolean, so I read the documentation about the topic, and I performed some changes to the PR. Could you please revisit it and confirm if the usage of AtomicBoolean is correct?

Kind regards,

src/main/java/com/cloudbees/jenkins/support/slowrequest/SlowRequestThreadDumpsGenerator.java
```diff
@@ -0,0 +1,147 @@
...
+     * Thread dumps generated on slowRequest scenario are stored in $JENKINS_HOME/support/slow-request-threaddumps
+     */
+    protected final FileListCap logs = new FileListCap(
+            new File(SupportPlugin.getRootDirectory(), "slow-request-threaddumps"),
+            SLOW_REQUEST_THREAD_DUMPS_TO_RETAIN);
+
+    /**
+     * Provide a means to disable the slow request thread dump checker.
+     */
+    public static final boolean DISABLED =
```
jglick, line 79:
```suggestion
    public static boolean DISABLED =
```
as a rule, so that problematic processes can be disabled in an emergency without a controller restart. (if SpotBugs complains, `@SuppressFBWarning`)

Dohbedoh (approved):

PR #570, opened 2024-08-12
Fix Markdown syntax of Non proxy hosts and Proxy host

Dohbedoh:
Recently discovered that wildcard syntax including several `*` are mis-interpreted by Markdown parsers.. For example `10.1.*.*` would be `10.1..` with a bold dot..

### Submitter checklist


Dohbedoh:
Or maybe we should just fix this entire content. The markdown is borked in the file and Proxy configuration does not make sense in an `update-center.md` anymore (if it ever did)..

jglick (approved):
Good so far as it goes. Other changes could be follow-ups?

PierreBtz:
> Or maybe we should just fix this entire content.

As far as changing things in depth, we could contemplate using a more machine friendly language :)

PR #577, opened 2024-09-05
[chore] Add "up" and "running" as stop words

PierreBtz:
If an instance contains jobs named "up" and/or "running", those end up being anonymized, leading to a log entry like: `Jenkins is fully item_outer_pub and item_square_life`.

This is not really convenient when browsing through the logs and looking for "up and running".

With this change, the log entry becomes `Jenkins is up and running`.

More generally, and outside of the context of this PR, I wonder if some groups of words should not be automatically excluded, like two letter words, adverbs etc...

### Submitter checklist

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/filter/DefaultStopWords.java
+++ b/src/main/java/com/cloudbees/jenkins/support/filter/DefaultStopWords.java
@@ -31,6 +31,8 @@ public class DefaultStopWords implements StopWords {
                 "everyone",
                 "system",
                 "admin",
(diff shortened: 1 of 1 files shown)


Dohbedoh:
> More generally, and outside of the context of this PR, I wonder if some groups of words should not be automatically excluded, like two letter words, adverbs etc...

Yeah could be included as part of supporting regexps: https://issues.jenkins.io/browse/JENKINS-71841

Dohbedoh (approved):

PR #580, opened 2024-09-17
Add ability to filter more Config Files from OtherConfigFilesComponent

Dohbedoh:
Add ability to provide more filters to the `OtherConfigFilesComponent` if some `xml` files are meant to be excluded in extensions and plugins.

### Submitter checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,7 +189,7 @@ This version changes the configuration fields to be more intuitive. Please updat
 you update support-core to version 2.68 or later.
 
 
-## Support Bundle sensitive data auto redaction
+## Support Bundle - Sensitive data auto redaction
 Starting from version 2.72.2, Support Core plugin could automatically redact any passwords stored in the following files:
 * `nodes.md`
 * `about.md`
@@ -210,6 +210,25 @@ will be changed to:
 -username.net.passwd=REDACTED
 ````
 To disable this feature, delete all of the security stop words from the security-stop-words.txt file (but not the file itself, because it will be automatically regenerated with a default values). Any changes made to the security-stop-words.txt file are applied after a Jenkins instance restart.
+
+### Support Bundle - Other Jenkins Configuration Files Filters
+
+The `Other Jenkins Configuration Files (Encrypted secrets are redacted)` option
+collects `$JENKINS_HOME/*.xml` with redacted secrets with some exceptions such as the
+`com.cloudbees.jenkins.support.filter.ContentMappings.xml` and `credentials.xml`.
+Additional filter can be provided in a text file located at
+`$JENKINS_HOME/support/other-config-files-filters.txt`. Each non-blank line of
+the file is treated as a regular expression filter. For example, the following 
+exclude files with `credentials` or `secrets` in their names and only includes 
+files that start with `org.jenkins`:
+
+```
+((?!(credentials|secrets)).)*
+^org\.jenkins.*
+```
+
+The system property `com.cloudbees.jenkins.support.configfiles.OtherConfigFilesComponent.otherConfigFilesFilterFile`
+can be used to override the file location.
 
 ## Changelog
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -211,6 +211,25 @@ will be changed to:
 ````
 To disable this feature, delete all of the security stop words from the security-stop-words.txt file (but not the file itself, because it will be automatically regenerated with a default values). Any changes made to the security-stop-words.txt file are applied after a Jenkins instance restart.
 
+### Support Bundle - Other Jenkins Configuration Files Filters
+
+The `Other Jenkins Configuration Files (Encrypted secrets are redacted)` option
+collects `$JENKINS_HOME/*.xml` with redacted secrets with some exceptions such as the
+`com.cloudbees.jenkins.support.filter.ContentMappings.xml` and `credentials.xml`.
+Additional filter can be provided in a text file located at
```
jglick, line 219: So this can only _subtract_ files that should not be included in the bundle, right?

What is the purpose of the hand-maintained list? The PR description talks about config files maintained by various plugins, but then they ought to add their own `@Extension`s; why would an admin wish to customize the filters?

Dohbedoh: > So this can only subtract files that should not be included in the bundle, right?

Yes. The patters are concatenated in a AND-style pattern `(?=pattern1)(?=pattern2).*`.

> What is the purpose of the hand-maintained list?

To allow an end user to control this was well.

> The PR description talks about config files maintained by various plugins, but then they ought to add their own @Extensions; why would an admin wish to customize the filters?

Plugin maintainers know best if a config file should be excluded so I thought that the extension approach was maybe better suited than maintaining a list in Support Core ? Also through an extension, plugin maintainer can "enforce" a file to be filtered. As opposed to for example a configurable text field in a global configuration.

Now I realize that this would require to add support-core as an optional dependency, even if not requiring the plugin. And that is not ideal.. 

WDYT would be a better approach ?

jglick: > this would require to add support-core as an optional dependency

That seems fine. Various plugins already have such an optional dep in order to add a `Component`.

> To allow an end user to control this was well.

Right but why would a Jenkins admin _need_ to do that? If plugins are already taking care of pruning the list down to files that are appropriate to include, then there should be no need for that.

It is hard to really evaluate this PR because there is no context / motivation.

Dohbedoh: Right, in general you wouldn't *need* to do this.
I can remove this additional feature then.

> It is hard to really evaluate this PR because there is no context / motivation.

Noticed recentely that a Proprietary plugin serialize a credentials cache in root config file (so similar to `credentials.xml`).

Dohbedoh: I removed the Regexp from file feature and simplify the interface with a `List<String>` of file names to exclude.

src/main/java/com/cloudbees/jenkins/support/configfiles/OtherConfigFilesComponent.java
```diff
@@ -99,4 +95,52 @@ public ComponentCategory getCategory() {
...
+            final Pattern allPattern = Pattern.compile("(" + String.join("|", filenames) + ")");
+            return (dir, name) -> !allPattern.matcher(name).matches();
+        }
+
+        /**
+         * Return a list of files names to exclude.
+         * @return list of file names to exclude
+         */
+        @NonNull
+        List<String> getFilenames();
```
jglick, line 126: This API feels like a step backwards from the `FilenameFilter`. Why not something like

```java
boolean include(File f);
```

? An implementation can choose to match the file basename against a static list of names, or do something else.

Dohbedoh: Sure. Makes sense as well.

jglick (approved):

PR #584, opened 2024-10-04
[JENKINS-73280] Migrate to Jakarta EE 9

Dohbedoh:
Started draft for Jakarta EE 9 migration.

### Testing done

TBD

### Submitter checklist
- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did


pom.xml
```diff
@@ -76,8 +76,8 @@
     <dependencies>
       <dependency>
         <groupId>io.jenkins.tools.bom</groupId>
-        <artifactId>bom-2.440.x</artifactId>
-        <version>3307.v2769886db_63b_</version>
+        <artifactId>bom-2.462.x</artifactId>
```
jglick, line 79: note https://github.com/jenkinsci/bom/pull/3677

pom.xml
```diff
@@ -128,6 +128,13 @@
       <artifactId>wordnet-random-name</artifactId>
       <version>1.6</version>
     </dependency>
+    <!--    <dependency>-->
+    <!--      &lt;!&ndash; Waiting for a release of https://github.com/jenkinsci/plugin-pom/pull/1004 &ndash;&gt;-->
+    <!--      <groupId>jakarta.servlet</groupId>-->
+    <!--      <artifactId>jakarta.servlet-api</artifactId>-->
+    <!--      <version>5.0.0</version>-->
+    <!--      <scope>provided</scope>-->
+    <!--    </dependency>-->
```
jglick, line 137:
```suggestion
```

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -184,7 +184,7 @@ public boolean isAnonymized() {
     }
 
     @RequirePOST
-    public void doDeleteBundles(StaplerRequest req, StaplerResponse rsp) throws ServletException, IOException {
+    public void doDeleteBundles(StaplerRequest2 req, StaplerResponse2 rsp) throws ServletException, IOException {
```
jglick, line 187: BTW better style is to use `HttpResponse` as a return value, in which case you can get rid of the `rsp` parameter and the `ServletException` throw. I am not sure if there is a clean replacement for `req.submittedForm` though, like a `@SomeAnnotation JSONObject form` parameter.

src/main/java/com/cloudbees/jenkins/support/SupportMetricsFilter.java
jglick, line 1: BTW this could be registered as an `HttpServletFilter`.

Dohbedoh: But the way we count the activeRequests, we need to wrap the `chain.doFilter` call.

pom.xml
Dohbedoh: Will bring the bom for `2.479.x`.

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
Dohbedoh: Switch to `HttpResponse`

Dohbedoh:
Rebased to fix the merge conflict. Should be good now.

PR #588, opened 2024-10-15
[JENKINS-73929] Handle custom buildsDir

Dohbedoh:
In case [buildsDir](https://www.jenkins.io/doc/book/managing/system-properties/#jenkins-model-jenkins-buildsdir) is customize, relativize against the buildsDir root directory to calculate the prefix. 

### Testing done

Tested with / without custom `buildsDir`
Tested with Multibranch branch job builds and standalone pipeline job builds

### Submitter checklist


src/main/java/com/cloudbees/jenkins/support/impl/RunDirectoryComponent.java
```diff
@@ -97,6 +91,26 @@ public boolean understandsSymlink() {
...
+                        Jenkins.get().getRawBuildsDir().replaceFirst("\\$\\{ITEM_FULL_NAME.*", ""),
+                        Map.of("JENKINS_HOME", Jenkins.get().getRootDir().getPath()))));
+
+        return Functions.isWindows()
+                ? buildsRootDir
+                        .toPath()
+                        .relativize(itemRootDir.toPath())
+                        .toString()
+                        .replace('\\', '/')
+                : buildsRootDir.toPath().relativize(itemRootDir.toPath()).toString();
```
jglick, line 111: or just
```suggestion
        return buildsRootDir
                        .toPath()
                        .relativize(itemRootDir.toPath())
                        .toString()
                        .replace(File.separatorChar, '/');
```
since `String.replace(char, char)` is a fast no-op if the args are equal.

src/main/java/com/cloudbees/jenkins/support/impl/RunDirectoryComponent.java
```diff
@@ -97,6 +91,26 @@ public boolean understandsSymlink() {
         }
     }
 
+    private static String getBuildRelativePath(@NonNull Run item, @NonNull File itemRootDir) {
+
+        // See Jenkins#isDefaultBuildDir that has restriction and cannot be used outside Jenkins core
```
jglick, line 96: I think it is `Jenkins.getBuildDirFor` you want to look at. What if `Jenkins.buildsDir` is set to, e.g., `${ITEM_ROOTDIR}/all-the-builds`?

jglick (commented):
Good though it does not handle all cases.

src/main/java/com/cloudbees/jenkins/support/impl/RunDirectoryComponent.java
Dohbedoh: Yeah, makes sense.

src/main/java/com/cloudbees/jenkins/support/impl/RunDirectoryComponent.java
Dohbedoh: Right... I think the simplest is to relativize if this is the default configuration, otherwise just use the item full name. We are not looking for the exact location of the build directory but rather deciding on what will be the file location in the bundle generated. And we should avoid absolute path.

By default, you can simply relativize and for example for a build dir using the default location `/var/jenkins_home/jobs/myfolder/jobs/myjob/builds/1` it can be easily relativized to `myfolder/jobs/myjob/builds/1`. 
But when the `buildsDir` is customize we don't want to use the absolute path, but choose a relevant path. I think the item full name should be fine. For example if the `buildsDir` was not the default one, the previous job example would yield `myfolder/myjob/builds/1`.

Dohbedoh:
Tested generation of bundle for a job, a build, a node and also standard Support Bundle. With and without the `buildsDir` customization.

Dohbedoh:
Tested generation of bundle for a job, a build, a node and also standard Support Bundle. With and without the `buildsDir` customization.

src/main/java/com/cloudbees/jenkins/support/api/ObjectComponent.java
```diff
@@ -86,8 +88,12 @@ public static <T extends AbstractModelObject> List<ObjectComponentDescriptor<T>>
      */
     @Override
     public <C extends AbstractModelObject> boolean isApplicable(Class<C> clazz) {
-        Class<T> type = (Class<T>) ((ParameterizedType) getClass().getGenericSuperclass()).getActualTypeArguments()[0];
-        return type.isAssignableFrom(clazz);
+        Type parameterization = Types.getBaseClass(getClass(), ObjectComponent.class);
+        if (parameterization instanceof ParameterizedType pt) {
+            return Types.erasure(Types.getTypeArgument(pt, 0)).isAssignableFrom(clazz);
```
jglick, line 93: Not sure what this change is going, but FWIW if you need to use `org.jvnet.tiger_types` you have already lost any sort of debuggability. Just make subclasses explicitly implement `isApplicable`.

src/main/java/com/cloudbees/jenkins/support/impl/RunDirectoryComponent.java
```diff
@@ -41,19 +40,18 @@ public RunDirectoryComponent(String includes, String excludes, boolean defaultEx
...
-                    ? new File(Jenkins.get().getRootDir(), "jobs")
-                            .toPath()
-                            .relativize(itemRootDir.toPath())
-                            .toString()
-                            .replace('\\', '/')
-                    : new File(Jenkins.get().getRootDir(), "jobs")
-                            .toPath()
-                            .relativize(itemRootDir.toPath())
-                            .toString();
+            String relativeToRoot =
```
jglick, line 46: TBH I am not clear on why we need to know this at all. We are just creating a ZIP file with info about one build, right? So why is there any prefix at all? Or if there is a prefix, why does it need to have anything to do with `$JENKINS_HOME` or other paths—why not simply use `item.getParent().getFullPath() + "/" + item.getNumber()`?

Dohbedoh: When introducing the Item / Build bundle feature, I had in mind to bring a feature that would allow to generate a bundle for a selection of items / builds. Hence the prefix to avoid conflicts.

jglick: OK…

src/main/java/com/cloudbees/jenkins/support/api/ObjectComponent.java
jglick: > Looking in core, I found out that we use `org.jvnet.tiger_types` in such kind of cases.

Yes, and we cannot fix that because we would be forced to break compatibility of poorly designed APIs used by plugins. There should be no such constraint here.

src/main/java/com/cloudbees/jenkins/support/impl/RunDirectoryComponent.java
Dohbedoh: The mix of Item full name and file system path (we use a `DirScanner` to retrieve files) would bring some inconsistency when we look at the different `Item` bundles (so with `AbstractItemDirectoryComponent`) and `Run` bundles (so with `RunDirectoryComponent`). 

If the component use the item full name as prefix, then we will have `${JenkinsItemPath}/${systemFilePathUnderItemDir}`. When taking the bundle of a folder item and taking a bundle of one of its child, the 2 bundles are not showing the same path to the `child` item. For example, bundle taken from `testMulti` and bundle taken from `testMulti/main`:

folder-bundle-vs-child-bunbdle

This is the reason why I tried to keep the file system path. Should a feature like I mentioned exist, it may show items in different location in the zip.. 

But I'm fine with using the Job full name for now. And revisit this later if/when we work on such a feature.

jglick: > the 2 bundles are not showing the same path to the child item

I think we should be excluding the `jobs/` or `branches/` subdirs from (folder) item exports?

jglick (approved):
Seems OK setting aside ongoing discussion.

PR #590, opened 2024-10-18
JENKINS-73984: More efficient filtering of input streams

basil:
I was made aware of an case where `LaunchLogsFileContent` was loading the entire agent launch log file into memory. More efficient to read at most one line at a time into memory, allowing the other lines to be garbage-collected.

### Testing done

`mvn clean verify`, new unit test

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/api/LaunchLogsFileContent.java
+++ b/src/main/java/com/cloudbees/jenkins/support/api/LaunchLogsFileContent.java
@@ -21,7 +21,7 @@ public class LaunchLogsFileContent extends FileContent {
 
     @Override
     protected InputStream getInputStream() throws IOException {
-        Function<String, String> filter = line -> PasswordRedactor.get().redact(line);
+        Function<String, String> filter = PasswordRedactor.get()::redact;
         return new FilteredInputStream(new FileInputStream(file), Charset.defaultCharset(), filter);
     }
 
--- a/src/main/java/com/cloudbees/jenkins/support/api/LaunchLogsFileContent.java
+++ b/src/main/java/com/cloudbees/jenkins/support/api/LaunchLogsFileContent.java
@@ -1,16 +1,14 @@
 package com.cloudbees.jenkins.support.api;
 
+import com.cloudbees.jenkins.support.filter.FilteredInputStream;
 import com.cloudbees.jenkins.support.filter.PasswordRedactor;
 import com.cloudbees.jenkins.support.impl.SlaveLaunchLogs;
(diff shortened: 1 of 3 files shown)


src/main/java/com/cloudbees/jenkins/support/api/LaunchLogsFileContent.java
```diff
@@ -23,14 +21,8 @@ public LaunchLogsFileContent(String name, String[] filterableParameters, File fi
...
     protected InputStream getInputStream() throws IOException {
-        try (FileInputStream inputStream = new FileInputStream(file)) {
-            List<String> strings = IOUtils.readLines(inputStream, Charset.defaultCharset());
-            byte[] bytes = strings.stream()
-                    .map(line -> PasswordRedactor.get().redact(line))
-                    .collect(Collectors.joining("\n", "", "\n"))
-                    .getBytes(Charset.defaultCharset());
-            return new ByteArrayInputStream(bytes);
-        }
+        Function<String, String> filter = line -> PasswordRedactor.get().redact(line);
```
jglick, line 24: May be slightly more efficient to avoid calling `lookupSingleton` for each line:
```suggestion
        Function<String, String> filter = PasswordRedactor.get()::redact;
```

src/main/java/com/cloudbees/jenkins/support/filter/FilteredInputStream.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.function.Function;
+
+/**
+ * A custom {@link InputStream} that filters lines of text read from an underlying {@link InputStream} using a provided
+ * mapping function. Each line read from the input stream is processed by the function before being returned, allowing
+ * for modifications such as redaction or transformation of the line content.
+ *
+ * @author Basil Crow
+ */
+public class FilteredInputStream extends InputStream {
```
jglick, line 19: Could possibly be replaced by https://javadoc.jenkins.io/hudson/console/LineTransformationOutputStream.Delegating.html

jglick (approved):
Amending c6d20da4f372f03bd3e4844f0df2f109df68a63c FTR.

src/main/java/com/cloudbees/jenkins/support/filter/FilteredInputStream.java
basil: Is that not an `OutputStream`, while in this case we need an `InputStream`?

Dohbedoh (approved):
Thanks @basil. I had some WIP and got interrupted with other tasks.. But reached a similar solution. LGTM.

PR #210, opened 2020-02-18
[JENKINS-36929] Collect Agents protocols information

Dohbedoh:
[JENKINS-36929](https://issues.jenkins-ci.org/browse/JENKINS-36929): Collect information about the agent protocols

PR #592, opened 2024-10-22
Removing `AgentProtocols`

jglick:
Reverting #210. Contingent on https://github.com/jenkinsci/jenkins/pull/9903 being merged.


MarkEWaite (approved):
Changes look good to me.  Passes the plugin BOM with an incremental build of 2.483

I was a little bit surprised that the [AgentProtocols reference](https://github.com/jenkinsci/support-core-plugin/blob/bf214ba25aa56eabb05e1d1e814241f9beaad6d1/src/test/java/com/cloudbees/jenkins/support/CasCTest.java#L41) in the CasCTest did not need to be changed.  I assume that is because this is not yet testing with Jenkins 2.483 or with configuration as code that drops AgentProtocols

jglick:
I suppose that means that the CasC setter here does not verify that the ids being requested are actually legitimate components, so typos etc. would be quietly ignored.

daniel-beck:
This seems wrong without a newer core dependency preventing installation on older cores for which this may provide useful information.

PR #601, opened 2024-12-10
API to supersede components if not required

nevingeorgesunny:
Added a way for components to supersede other components in the support bundle creation process

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
+++ b/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
@@ -596,8 +596,13 @@ public class SupportPlugin extends Plugin {
 
         manifest.append("Requested components:\n\n");
         ContentContainer contentsContainer = new ContentContainer(contentFilter, components);
+        Set<Class<? extends Component>> supersededComponents = new HashSet<>();
+        components.forEach(c -> supersededComponents.addAll(c.getSupersededComponents()));
         for (Component component : components) {
             try {
+                if (supersededComponents.contains(component.getClass())) {
+                    continue;
+                }
                 manifest.append("  * ").append(component.getDisplayName()).append("\n\n");
                 LOGGER.log(Level.FINE, "Start processing " + component.getDisplayName());
                 long startTime = System.currentTimeMillis();
--- a/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
+++ b/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
@@ -596,13 +596,12 @@ public class SupportPlugin extends Plugin {
 
         manifest.append("Requested components:\n\n");
         ContentContainer contentsContainer = new ContentContainer(contentFilter, components);
-        Set<Class<? extends Component>> supersededComponents = new HashSet<>();
-        components.forEach(c -> supersededComponents.addAll(c.getSupersededComponents()));
         for (Component component : components) {
             try {
-                if (supersededComponents.contains(component.getClass())) {
+                if (components.stream().anyMatch(c -> c.supersedes(component))) {
                     continue;
                 }
+
                 manifest.append("  * ").append(component.getDisplayName()).append("\n\n");
                 LOGGER.log(Level.FINE, "Start processing " + component.getDisplayName());
                 long startTime = System.currentTimeMillis();
--- a/src/main/java/com/cloudbees/jenkins/support/api/Component.java
+++ b/src/main/java/com/cloudbees/jenkins/support/api/Component.java
@@ -154,7 +154,7 @@ public abstract class Component implements ExtensionPoint {
     }
 
     public Set<Class<? extends Component>> getSupersededComponents() {
-        return Collections.emptySet();
+        return Set.of();
     }
 
     /**
--- a/src/main/java/com/cloudbees/jenkins/support/api/Component.java
+++ b/src/main/java/com/cloudbees/jenkins/support/api/Component.java
@@ -153,6 +153,10 @@ public abstract class Component implements ExtensionPoint {
         return ComponentCategory.UNCATEGORIZED;
     }
 
+    public Set<Class<? extends Component>> getSupersededComponents() {
+        return Collections.emptySet();
+    }
+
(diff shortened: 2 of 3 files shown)


src/main/java/com/cloudbees/jenkins/support/api/Component.java
```diff
@@ -153,6 +153,10 @@ public ComponentCategory getCategory() {
         return ComponentCategory.UNCATEGORIZED;
     }
 
+    public Set<Class<? extends Component>> getSupersededComponents() {
+        return Collections.emptySet();
```
jglick, line 157: or just
```suggestion
        return Set.of();
```

src/main/java/com/cloudbees/jenkins/support/api/Component.java
```diff
@@ -153,6 +153,10 @@ public ComponentCategory getCategory() {
         return ComponentCategory.UNCATEGORIZED;
     }
 
+    public Set<Class<? extends Component>> getSupersededComponents() {
```
jglick, line 156: Needs Javadoc.

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -596,8 +596,13 @@ private static List<Content> appendManifestContents(
 
         manifest.append("Requested components:\n\n");
         ContentContainer contentsContainer = new ContentContainer(contentFilter, components);
+        Set<Class<? extends Component>> supersededComponents = new HashSet<>();
+        components.forEach(c -> supersededComponents.addAll(c.getSupersededComponents()));
         for (Component component : components) {
             try {
+                if (supersededComponents.contains(component.getClass())) {
```
jglick, line 603: The use of `Class` in APIs is usually a bit of an antipattern since it makes assumptions about implementation hierarchies; and actually `@Extension` is not strictly limited to singleton classes. Maybe a `boolean supersedes(Component)`?

Dohbedoh: Right. Some components have class Hierarchy, for example [ProcFilesRetriever](https://github.com/jenkinsci/support-core-plugin/blob/1538.v0a_a_a_e0f2e1c0/src/main/java/com/cloudbees/jenkins/support/impl/ProcFilesRetriever.java).

You mean a `Component#supersedes(component)` ? That would require a O(n^2) loop to populated the list of superseded components ? But I guess should be fine, we are not talking about 100s of components.

Dohbedoh:
Could we please also add a test to cover this.

src/main/java/com/cloudbees/jenkins/support/api/Component.java
```diff
@@ -153,8 +153,13 @@ public ComponentCategory getCategory() {
         return ComponentCategory.UNCATEGORIZED;
     }
 
-    public Set<Class<? extends Component>> getSupersededComponents() {
-        return Collections.emptySet();
+    /**
+     * Returns true if a component is superseded by this component.
+     * useful if we write a component that makes another one obsolete.
+     * @return
```
jglick, line 159: either delete
```suggestion
```
or write something

nevingeorgesunny:
> Could we please also add a test to cover this.

Done added a test case

src/main/java/com/cloudbees/jenkins/support/api/Component.java
nevingeorgesunny: done removed

src/test/java/com/cloudbees/jenkins/support/SupportPluginTest.java
```diff
@@ -39,6 +39,11 @@ public class SupportPluginTest {
     @Rule
     public TemporaryFolder temp = new TemporaryFolder();
 
+    @Test
+    public void testSomething(){
+        j.getInstance();
+    }
+
     @Test
```
jglick:
```suggestion
```

??

src/test/java/com/cloudbees/jenkins/support/SupportPluginTest.java
```diff
@@ -103,4 +108,62 @@ public long getTime() throws IOException {
...
+                    public void addContents(@NonNull Container container) {
+                        container.add(new Content("test/testWriteBundleWithJenkinsRule.md") {
+                            @Override
+                            public void writeTo(OutputStream os) throws IOException {
+                                os.write("test content".getBytes(StandardCharsets.UTF_8));
+                            }
+                        });
+                    }
+                },
+                ExtensionList.lookup(Component.class).get(AboutJenkins.class),
```
jglick, line 144: use `lookupSingleton`

src/test/java/com/cloudbees/jenkins/support/SupportPluginTest.java
```diff
@@ -103,4 +108,62 @@ public long getTime() throws IOException {
...
+
+    /**
+     * Test that a component can supersede another component.
+     * This is similar to testGenerateBundleExceptionHandler() but we are superseding
+     * AboutJenkins and BuildQueue components. even if is added to the list of components to create.
+     * @throws Exception
+     */
+    @Test
+    public void testSupersedesComponent() throws Exception {
+        List<Component> components = Arrays.asList(
```
jglick, line 115: or `List.of`

src/test/java/com/cloudbees/jenkins/support/SupportPluginTest.java
```diff
@@ -103,4 +108,62 @@ public long getTime() throws IOException {
...
+
+                    @NonNull
+                    @Override
+                    public String getDisplayName() {
+                        return "Test Component";
+                    }
+
+                    @Override
+                    public boolean supersedes(Component component) {
+                        return Set.of(AboutJenkins.class, BuildQueue.class).contains(component.getClass());
```
jglick, line 131: perhaps more clearly just
```suggestion
                        return component instanceof AboutJenkins || component instanceof BuildQueue;
```

jglick (approved):

Dohbedoh (approved):

PR #617, opened 2025-01-27
Generate support bundle Async

nevingeorgesunny:
## The problem that I am trying to solve 
Currently, when users click on the "Generate Bundle" button on the /jenkins/support/ page, There is no user feedback.In case the generation process takes too long, the request may result in a 503 Timeout

## Here is the fix

We will generate support bundle anync in backend which we poll for its status from UI.
The support bundle will be saved in the tmp folder, this will be  deleted once the user has downloaded the files

here is a gif for the UX

GIF Recording 2025-02-10 at 10 12 43 AM


src/test/java/com/cloudbees/jenkins/support/SupportActionTest.java
```diff
@@ -309,11 +314,17 @@ private WebResponse doBundle(String action, String bundle, String user, String e
     }
 
     private ZipFile downloadBundle(String s) throws IOException, SAXException {
```
nevingeorgesunny, line 316: All the test remain the same , as  the only thing that changed was how the zip file was obtained, previoulsy it was a sync call , now its async , we just fetch the zip file after the support bundle is created

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -318,6 +340,50 @@ public void doGenerateAllBundles(StaplerRequest2 req, StaplerResponse2 rsp) thro
...
+
+    /**
+     * Generates a support bundle with selected components from the UI. in async
+     * @param req The stapler request
+     * @param rsp The stapler response
+     * @throws ServletException If an error occurred during form submission
+     * @throws IOException If an input or output exception occurs
+     */
+    @RequirePOST
+    public HttpRedirect doGenerateAllBundlesAsync(StaplerRequest2 req, StaplerResponse2 rsp)
```
nevingeorgesunny, line 355: I have created a new method , and left the existing methods as it is , as it being used by REST apis

src/main/java/com/cloudbees/jenkins/support/impl/AboutBrowser.java
```diff
@@ -37,9 +37,27 @@ public Set<Permission> getRequiredPermissions() {
         return Collections.emptySet();
     }
 
+    private Area screenResolution;
+    private StaplerRequest2 currentRequest;
```
nevingeorgesunny, line 41: added this logic.only in the case of asyn `AboutBrowser` component creation , 
as it needs cookie , these values will be set [here](https://github.com/jenkinsci/support-core-plugin/blob/931b8d6205d51663527519e3c734e8a0d0a718df/src/main/java/com/cloudbees/jenkins/support/SupportAction.java#L364C1-L377C1) before we start the genrate bundle process

src/test/java/com/cloudbees/jenkins/support/SupportActionTest.java
```diff
@@ -364,16 +420,7 @@ public void takeSnapshotAndMakeSureSomethingHappens() throws Exception {
...
-            HtmlPage p = wc.goTo(root.getUrlName());
-
-            HtmlForm form = p.getFormByName("bundle-contents");
-            HtmlButton submit = (HtmlButton) form.getElementsByTagName("button").get(0);
-            Page zip = submit.click();
-            File zipFile = File.createTempFile("test", "zip");
-            IOUtils.copy(zip.getWebResponse().getContentAsStream(), Files.newOutputStream(zipFile.toPath()));
-
-            ZipFile z = new ZipFile(zipFile);
+            ZipFile z = downloadSupportBundleByButtonClick();
```
nevingeorgesunny, line 423: I modified the existing test to use the new button click - poll for download button method for test this

src/test/java/com/cloudbees/jenkins/support/SupportActionTest.java
```diff
@@ -595,4 +643,22 @@ private void assertBundleNotContains(ZipFile zip, Collection<String> fileNames)
             assertNull(file + " was found in the bundle", zip.getEntry(file));
         }
     }
+
+    public void cleanUp() {
```
nevingeorgesunny, line 647: clean up required for removing the support bundle that i created in teh /tmp file after tests that is useing the UI support bundle dowload method

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -92,6 +107,13 @@ public class SupportAction implements RootAction, StaplerProxy {
      */
     private final Logger logger = Logger.getLogger(SupportAction.class.getName());
 
+    private static final String SUPPORT_BUNDLE_FILE_NAME = "support-bundle.zip";
+    private static final String SUPPORT_BUNDLE_CREATION_FOLDER = Paths.get(System.getProperty("java.io.tmpdir"))
+            .resolve("support-bundle")
+            .toString();
```
gbhat618, line 113: the system property `java.io.tmpdir` is always expected to exist? or could lead to NPE?

.DS_Store
gbhat618, line 1: `.DS_Store` probably need to put in `.gitignore`

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -92,6 +107,13 @@
...
+
+    private static final Map<UUID, SupportBundleAsyncGenerator> generatorMap = new ConcurrentHashMap<>();
```
gbhat618, line 115: nit: could be renamed to `generatorByTaskId`

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -439,4 +505,101 @@ public boolean isSelected() {
             return selected;
         }
     }
+
+    public ProgressiveRendering getGenerateSupportBundle(@QueryParameter String taskId) throws Exception {
```
gbhat618, line 509: nit: could be renamed to bit short `getGeneratorByTaskId(..)`

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -439,4 +505,101 @@ public boolean isSelected() {
...
+
+        public SupportBundleAsyncGenerator init(UUID taskId, List<Component> components) {
+            this.taskId = taskId;
+            this.components = components;
+            return this;
+        }
+
+        @Override
+        protected void compute() throws Exception {
+            if (!supportBundleGenerationInProgress) {
```
gbhat618, line 529: nit: can consider to early return to reduce nested block.. if (supportBundleGenerationInProgress) { return; } ... rest of the code

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -439,4 +505,101 @@ public boolean isSelected() {
...
+                }
+
+                try (FileOutputStream fileOutputStream =
+                        new FileOutputStream(new File(outputDir, SUPPORT_BUNDLE_FILE_NAME))) {
+                    SupportPlugin.setRequesterAuthentication(Jenkins.getAuthentication2());
+                    try {
+                        try (ACLContext ignored = ACL.as2(ACL.SYSTEM2)) {
+                            SupportPlugin.writeBundle(fileOutputStream, components);
+                        } catch (IOException e) {
+                            logger.log(Level.FINE, e.getMessage(), e);
```
gbhat618, line 546: this IOException means support bundle generation was not successful right? maybe needs to log at a higher level, perhaps `WARNING` ? - while suggesting `WARNING` i am assuming within the lifecycle of one bundle generation it won't keep logging multiple times...

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -439,4 +505,101 @@ public boolean isSelected() {
...
+            }
+        }
+
+        @NonNull
+        @Override
+        protected JSON data() {
+            JSONObject json = new JSONObject();
+            json.put("isCompleted", isCompleted);
+            json.put("pathToBundle", pathToBundle);
+            json.put("status", 1);
```
gbhat618, line 567: maybe can add an inline comment what the number `1` means (seems like success), or better the comment can link to an existing javadoc or comment having all possible values for `status`.. (perhaps it has only two values `0` and `1` maybe..)

src/test/java/com/cloudbees/jenkins/support/SupportActionTest.java
```diff
@@ -316,6 +331,47 @@ private ZipFile downloadBundle(String s) throws IOException, SAXException {
...
+        HtmlButton submit = (HtmlButton) form.getElementsByTagName("button").get(0);
+        HtmlPage page = submit.click();
+
+        HtmlAnchor downloadLink = null;
+        int maxRetries = 10;
+        int retryCount = 0;
+        int waitTime = 1000;
+        File zipFile;
+
+        while (retryCount < maxRetries) {
```
gbhat618, line 347: perhaps using `await().timeout.(..)until(() -> ...)` from hamcrest library (used a lot of places inside jenkins for this kind of waiting scenarios in test)

src/test/java/com/cloudbees/jenkins/support/SupportActionTest.java
```diff
@@ -595,4 +643,22 @@ private void assertBundleNotContains(ZipFile zip, Collection<String> fileNames)
             assertNull(file + " was found in the bundle", zip.getEntry(file));
         }
     }
+
+    public void cleanUp() {
+        try {
+            if (Files.exists(SUPPORT_BUNDLE_CREATION_FOLDER)) {
```
gbhat618, line 649: nit: could consider early return approach for less nesting.. `if (!Files.exists(..) {return; } code...`

src/test/java/com/cloudbees/jenkins/support/SupportActionTest.java
```diff
@@ -68,6 +70,12 @@
  * @author Kohsuke Kawaguchi
  */
 public class SupportActionTest {
+
+    private static final Logger LOGGER = Logger.getLogger(SupportActionTest.class.getName());
+
+    private static final Path SUPPORT_BUNDLE_CREATION_FOLDER =
+            Paths.get(System.getProperty("java.io.tmpdir")).resolve("support-bundle");
```
gbhat618, line 77: maybe using the `org.junit.rules.TemporaryFolder` helps in easy cleanup later :think (used for temp folders in several place in jenkins & plugins)

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
gbhat618: it picks default. sorry didn't know

.DS_Store
nevingeorgesunny: Yeah this got committed by accident , removed

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
nevingeorgesunny: yeah makes sense, renamed

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
nevingeorgesunny: yeah makes sense, renamed

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
nevingeorgesunny: yep , set to warning

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
nevingeorgesunny: status", 1) was a leftover from my testing , i removed it

src/test/java/com/cloudbees/jenkins/support/SupportActionTest.java
nevingeorgesunny: I was trying to do a integration  test , where the code will created a download link ,, which points to the file created [here](https://github.com/jenkinsci/support-core-plugin/blob/d98700d499ea7ec898a6e8bfa3eac6e32bdad3bf/src/main/java/com/cloudbees/jenkins/support/SupportAction.java#L543C1-L544C1) 

`TemporaryFolder` can be useful for passing an outputstream and getting the file directly like [here](https://github.com/jenkinsci/support-core-plugin/blob/8facfe41cc70e7efc2c7533bb3f1e38b0c582874/src/test/java/com/cloudbees/jenkins/support/SupportActionTest.java#L604C7-L605C77) 
I mean i think i can pass the temporal file in to SupportAction if the scope is test , but then i think it would add a few if and else conditions just for test

src/test/java/com/cloudbees/jenkins/support/SupportActionTest.java
nevingeorgesunny: done , this is more readable

src/test/java/com/cloudbees/jenkins/support/SupportActionTest.java
nevingeorgesunny: yeah that makes sense

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
nevingeorgesunny: yeah makes sense , done

src/test/java/com/cloudbees/jenkins/support/SupportActionTest.java
```diff
@@ -316,6 +332,42 @@ private ZipFile downloadBundle(String s) throws IOException, SAXException {
...
+            try {
+                return pageRef.get().getAnchorByText("Download Support Bundle") != null;
+            } catch (ElementNotFoundException e) {
+                pageRef.set((HtmlPage) pageRef.get().refresh());
+                return false;
+            }
+        });
+
+        downloadLink = pageRef.get().getAnchorByText("Download Support Bundle");
+        if (downloadLink != null) {
```
gbhat618, line 358: instead of doing `if-else` probably better to `assertThat("download link shouldn't be null", downloadLink, not(nullValue())`

gbhat618 (approved):

gbhat618 (approved):
I checked out the code, tested in my local, async bundle generation andUI navigation works. (although I didn't test for generating large bundles taking several minutes and such things)

Reviewed the java code, LGTM.
(I don't have deep knowledge of support-core-plugin though)

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
Vlatombe: Probably more straight to the point as
```
Files.createTempDirectory("support-bundle")
```

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
Vlatombe: Why plural? afaiu you generate one bundle with all selected components, so `doGenerateBundleAsync` would be clearer to me.

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -318,6 +339,50 @@ public void doGenerateAllBundles(StaplerRequest2 req, StaplerResponse2 rsp) thro
...
+                aboutBrowser.setScreenResolution(Functions.getScreenResolution());
+                aboutBrowser.setCurrentRequest(Stapler.getCurrentRequest2());
+                aboutBrowser.setGeneratedAsync(true);
+            }
+
+            if (component instanceof ReverseProxy) {
+                ReverseProxy reverseProxy = (ReverseProxy) component;
+                reverseProxy.setCurrentRequest(Stapler.getCurrentRequest2());
+            }
+        }
```
Vlatombe, line 375: Ok so the high level problem is that some components can't support async because they need data from the current request. Could add a new method on `Component` to indicate that fact.

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -318,6 +339,50 @@ public void doGenerateAllBundles(StaplerRequest2 req, StaplerResponse2 rsp) thro
...
+            return new HttpRedirect("support");
+        }
+        final List<Component> components = getComponents(req, json);
+
+        for (Component component : components) {
+            if (component instanceof AboutBrowser) {
+                AboutBrowser aboutBrowser = (AboutBrowser) component;
+                aboutBrowser.setScreenResolution(Functions.getScreenResolution());
+                aboutBrowser.setCurrentRequest(Stapler.getCurrentRequest2());
+                aboutBrowser.setGeneratedAsync(true);
```
Vlatombe, line 368: I don't think you need to report that, since in the end this shouldn't be a user choice.

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
jglick: Right, I think that was my suggestion a week ago: default to allow asynch generation, but for a couple specific components, override the method to flag that they need the request from the form submission and so should be run synchronously.

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
jglick: Unfortunately not. See description at https://github.com/jenkinsci/jenkins-test-harness/pull/838.

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -92,6 +107,12 @@ public class SupportAction implements RootAction, StaplerProxy {
      */
     private final Logger logger = Logger.getLogger(SupportAction.class.getName());
 
+    private static final String SUPPORT_BUNDLE_CREATION_FOLDER = Paths.get(System.getProperty("java.io.tmpdir"))
+            .resolve("support-bundle")
+            .toString();
```
jglick, line 112: Better to type the (pseudo-)constant as a `Path`.

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -439,4 +504,109 @@ public boolean isSelected() {
             return selected;
         }
     }
+
+    public ProgressiveRendering getGeneratorByTaskId(@QueryParameter String taskId) throws Exception {
```
jglick, line 508:
```suggestion
    public ProgressiveRendering getGeneratorByTaskId(String taskId) throws Exception {
```
This is not a web method, or at least it is not being used as one.

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -439,4 +504,109 @@ public boolean isSelected() {
...
+                } finally {
+                    SupportPlugin.clearRequesterAuthentication();
+                }
+            } finally {
+                logger.fine("Response completed");
+            }
+
+            pathToBundle = outputDir.getAbsolutePath() + "/" + supportBundleName;
+            isCompleted = true;
+            progress(1);
```
jglick, line 560: Would be nicer to add a callback to `SupportPlugin.writeBundle` so that the progress bar can actually reflect how many `Component`s have been written.

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -439,4 +504,109 @@ public boolean isSelected() {
...
+            if (!outputDir.exists()) {
+                if (!outputDir.mkdirs()) {
+                    throw new IOException("Failed to create directory: " + outputDir.getAbsolutePath());
+                }
+            }
+
+            try (FileOutputStream fileOutputStream = new FileOutputStream(new File(outputDir, supportBundleName))) {
+                SupportPlugin.setRequesterAuthentication(Jenkins.getAuthentication2());
+                try {
+                    try (ACLContext ignored = ACL.as2(ACL.SYSTEM2)) {
```
jglick, line 546: Was this idiom already in the plugin? Seems counterintuitive and even risky. I would expect all `Component`s to expect to run as some user authn (`SYSTEM` for a periodic hourly bundle) and to call `as(SYSTEM)` themselves if and when appropriate.

jglick, line 546: see #623

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -439,4 +504,109 @@ public boolean isSelected() {
...
+                        SupportPlugin.writeBundle(fileOutputStream, components);
+                    } catch (IOException e) {
+                        logger.log(Level.WARNING, e.getMessage(), e);
```
jglick, line 549:
```suggestion
                        logger.log(Level.WARNING, null, e);
```
if you have nothing more informative to say

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -439,4 +504,109 @@ public boolean isSelected() {
...
+            isCompleted = true;
+            progress(1);
+        }
+
+        @NonNull
+        @Override
+        protected JSON data() {
+            JSONObject json = new JSONObject();
+            json.put("isCompleted", isCompleted);
+            json.put("pathToBundle", pathToBundle);
```
jglick, line 568: AFAICT this is unused?

src/main/resources/com/cloudbees/jenkins/support/SupportAction/progressPage.jelly
```diff
@@ -0,0 +1,42 @@
...
+        </style>
+        <l:main-panel>
+            <h2>Generating Support Bundle</h2>
+            <p id="progressMessage">Generating a support bundle for this Jenkins instance. This may take a few minutes.</p>
+            <l:progressiveRendering handler="${it.getGeneratorByTaskId(request.getParameter('taskId'))}" callback="updateProgressUI"/>
+            <l:progressAnimation/>
+            <a id="downloadButton" style="display:none;" class="btn btn-primary" href="#">Download Support Bundle</a>
+        </l:main-panel>
+
+        <script>
```
jglick, line 17: CSP violation: all scripts must be in a separate `*.js` file included via `st:adjunct`.

src/main/resources/com/cloudbees/jenkins/support/SupportAction/progressPage.jelly
```diff
@@ -0,0 +1,42 @@
...
+                    progressBar.style.display = 'none';
+                }
+                if (data.isCompleted === true) {
+                    var ellipsisProgressBar = document.querySelector('.lds-ellipsis');
+                    if (ellipsisProgressBar) {
+                        ellipsisProgressBar.style.display = 'none';
+                    }
+                    downloadButton.style.display = "block";
+                    downloadButton.href = "downloadBundle?taskId=" + data.taskId; // Update this path accordingly
+                    document.getElementById("progressMessage").textContent = "Support bundle has been generated.";
```
jglick, line 30: There should be some way to trigger the download automatically, so that the link would only be a fallback in case that is broken in some browser.

src/main/resources/com/cloudbees/jenkins/support/SupportAction/progressPage.jelly
```diff
@@ -0,0 +1,42 @@
...
+            <l:progressAnimation/>
+            <a id="downloadButton" style="display:none;" class="btn btn-primary" href="#">Download Support Bundle</a>
+        </l:main-panel>
+
+        <script>
+            function updateProgressUI(data) {
+                var progressBar = document.querySelector('.app-progress-bar');
+                if (progressBar) {
+                    progressBar.style.display = 'none';
+                }
```
jglick, line 22: This should be handled by `ProgressiveRendering` automatically?

src/main/resources/com/cloudbees/jenkins/support/SupportAction/progressPage.jelly
```diff
@@ -0,0 +1,42 @@
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:l="/lib/layout">
+    <l:layout title="Progress Page">
```
jglick, line 3:
```suggestion
    <l:layout title="Support bundle generation">
```
perhaps

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
nevingeorgesunny: So I had something in my , I will create the contents that have dependency on the request first like [this](https://github.com/jenkinsci/support-core-plugin/blob/c8a1eb7d43d0aaaa7891a219903e272c34c452a7/src/main/java/com/cloudbees/jenkins/support/SupportAction.java#L362C8-L379C10) 
It will create a support-bundle.zip file in /tmp

then the async process will be called [here](https://github.com/jenkinsci/support-core-plugin/blob/c8a1eb7d43d0aaaa7891a219903e272c34c452a7/src/main/java/com/cloudbees/jenkins/support/SupportAction.java#L550)  to process the remaining components 
this will be saved to a new zip file

and finally both the zip file will be merged to one  [here](https://github.com/jenkinsci/support-core-plugin/blob/c8a1eb7d43d0aaaa7891a219903e272c34c452a7/src/main/java/com/cloudbees/jenkins/support/SupportAction.java#L557C1-L557C17)

--- an alternate  idea I had was , to process the request dependent components firsts and keep the contents in memory and pass it to the context of the asyn process , But I was thinking if these contents are too large it could cause issues

src/main/resources/com/cloudbees/jenkins/support/SupportAction/progressPage.jelly
nevingeorgesunny: I was doing this because there was no callback mentioned  in your comments [here](https://github.com/jenkinsci/support-core-plugin/pull/617#discussion_r1949651952).

But with the addition of the callback this will not be required

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
nevingeorgesunny: So i created something like this [here](https://github.com/jenkinsci/support-core-plugin/blob/c8a1eb7d43d0aaaa7891a219903e272c34c452a7/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java#L379C1-L380C1)  to set the progress , on every component visit.

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
nevingeorgesunny: update from SU , instead of creating 2 files and then merging it , 
just create the request dependent components save it to disk and use it while creating async friendly support bundle zip file

src/main/resources/com/cloudbees/jenkins/support/SupportAction/progressPage.jelly
nevingeorgesunny: I have added [progressPage.js](https://github.com/jenkinsci/support-core-plugin/pull/617/files#diff-4b4ae585968e41b5fac0880eacef95ecc729215a9f103e3f58f9e261b7ed4abb)

this will trigger an auto download

src/main/resources/com/cloudbees/jenkins/support/SupportAction/progressPage.jelly
nevingeorgesunny: im now using <st:adjunct includes="js/progressPage"/>

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
nevingeorgesunny: removed ,

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -439,4 +504,109 @@ public boolean isSelected() {
...
+                        logger.log(Level.WARNING, e.getMessage(), e);
+                    }
+                } finally {
+                    SupportPlugin.clearRequesterAuthentication();
+                }
+            } finally {
+                logger.fine("Response completed");
+            }
+
+            pathToBundle = outputDir.getAbsolutePath() + "/" + supportBundleName;
```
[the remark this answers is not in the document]
nevingeorgesunny: pathToBundle is removed

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
nevingeorgesunny: removed

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
nevingeorgesunny: set to Path

src/test/java/com/cloudbees/jenkins/support/SupportActionTest.java
```diff
@@ -595,4 +639,23 @@ private void assertBundleNotContains(ZipFile zip, Collection<String> fileNames)
             assertNull(file + " was found in the bundle", zip.getEntry(file));
         }
     }
+
+    public void cleanUpSupportBundleInTempFolder() {
```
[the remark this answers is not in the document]
nevingeorgesunny: yes it is working fine without it , removed

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
nevingeorgesunny: I have created an abstract method , to process some components sync

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
nevingeorgesunny: I now adding the request dependent components in the async flow like [this](https://github.com/jenkinsci/support-core-plugin/blob/3a128ef335c7503596257b6eaab4f68e0dc62bc1/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java#L494C17-L526C18)

src/test/java/com/cloudbees/jenkins/support/SupportActionTest.java
nevingeorgesunny: in this method `downloadSupportBundleByButtonClick ` i need the zip file,  `downloadSupportBundleByButtonClick` is more like a utils class that is used by other test methods https://github.com/jenkinsci/support-core-plugin/blob/3a128ef335c7503596257b6eaab4f68e0dc62bc1/src/test/java/com/cloudbees/jenkins/support/SupportActionTest.java#L108

src/main/resources/com/cloudbees/jenkins/support/SupportAction/progressPage.jelly
nevingeorgesunny: this logic is removed

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
nevingeorgesunny: adapted to match the PR

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
nevingeorgesunny: This issues was removed while I addressed https://github.com/jenkinsci/support-core-plugin/pull/617#discussion_r1949699443

nevingeorgesunny:
looks like the windows tests are failing, let me check . even after a re triggers

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
jglick: (if it is accepted, TBD)

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -351,34 +348,39 @@ public void doGenerateAllBundles(StaplerRequest2 req, StaplerResponse2 rsp) thro
      * @throws IOException If an input or output exception occurs
      */
     @RequirePOST
-    public HttpRedirect doGenerateAllBundlesAsync(StaplerRequest2 req, StaplerResponse2 rsp)
+    public HttpRedirect doGenerateAllBundleAsync(StaplerRequest2 req, StaplerResponse2 rsp)
```
jglick, line 351: Why not just `generateBundleAsync`? Why the _all_?

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -351,34 +348,39 @@ public void doGenerateAllBundles(StaplerRequest2 req, StaplerResponse2 rsp) thro
...
-                aboutBrowser.setScreenResolution(Functions.getScreenResolution());
-                aboutBrowser.setCurrentRequest(Stapler.getCurrentRequest2());
-                aboutBrowser.setGeneratedAsync(true);
+        // There are some components that need the request components to be processed
+        // these components cannot be processed async
+        // so process them first and then process the other components async
+        List<Component> syncComponent =
+                components.stream().filter(c -> !c.canBeGeneratedAsync()).toList();
+        if (!syncComponent.isEmpty()) {
+            File outputDir = new File(SUPPORT_BUNDLE_CREATION_FOLDER + "/" + taskId);
```
jglick, line 367: Maybe easier to create `support-bundle-$taskId.zip` directly in `SUPPORT_BUNDLE_CREATION_FOLDER` rather than needing to create and clean up a subdirectory.

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -541,31 +543,20 @@ protected void compute() throws Exception {
...
-                try {
-                    try (ACLContext ignored = ACL.as2(ACL.SYSTEM2)) {
-                        SupportPlugin.writeBundle(fileOutputStream, components);
-                    } catch (IOException e) {
-                        logger.log(Level.WARNING, e.getMessage(), e);
-                    }
-                } finally {
-                    SupportPlugin.clearRequesterAuthentication();
-                }
+                SupportPlugin.writeBundle(
```
jglick, line 546: (If #623 is not merged first, then this would need call `setRequesterAuthentication`.)

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -352,12 +357,46 @@ public static void writeBundle(OutputStream outputStream) throws IOException {
...
+    /**
+     * Generate a bundle for all components that are selected in the Global Configuration.
+     *
+     * @param outputStream an {@link OutputStream}
+     * @param components a list of {@link Component} to include in the bundle
+     * @param progressCallback a callback to report progress back to the UI see ProgressiveRendering.progress
+     * @param outputPath the path with the support bundle will be created in the cases of async generations
+     * @throws IOException if an error occurs while generating the bundle.
+     */
+    public static void writeBundle(
```
jglick, line 381: Does not need to be `public` I guess.

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -444,6 +488,42 @@ public static void writeBundle(
                                         + name);
                     }
                 }
+
+                // process for async components
+                if (!components.stream()
+                                .filter(c -> !c.canBeGeneratedAsync())
+                                .toList()
+                                .isEmpty()
```
jglick, line 496: Or could just check to see if the synch bundle is in fact there. (https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#newInputStream-java.nio.file.Path-java.nio.file.OpenOption...- will throw `NoSuchFileException` so just `catch` that; no need to precheck existence.)

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -444,6 +488,42 @@ public static void writeBundle(
...
+                        && outputPath != null) {
+                    try {
+                        File zipFile = outputPath.resolve(SYNC_SUPPORT_BUNDLE).toFile();
+                        try (ZipInputStream zis = new ZipInputStream(new FileInputStream(zipFile))) {
+                            ZipEntry entry;
+                            while ((entry = zis.getNextEntry()) != null) {
+                                if (entry.getName().equals("manifest.md")) {
+                                    // no need to add the manifest.md as it will be created in the async flow
+                                    continue;
+                                }
```
jglick, line 506: Cleaner to pass a `boolean` param to the bundle creation method indicating whether `manifest.md` is wanted to begin with.

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -444,6 +488,42 @@ public static void writeBundle(
...
+
+                                binaryOut.putNextEntry(entry);
+
+                                // Copy the content
+                                byte[] buffer = new byte[4096];
+                                int bytesRead;
+                                while ((bytesRead = zis.read(buffer)) != -1) {
+                                    binaryOut.write(buffer, 0, bytesRead);
+                                }
```
jglick, line 515:
```suggestion
                                zis.transferTo(binaryOut);
```

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -444,6 +488,42 @@ public static void writeBundle(
...
+
+                                binaryOut.flush();
```
jglick, line 517: Unnecessary AFAIK.
```suggestion
```

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -444,6 +488,42 @@ public static void writeBundle(
...
+                            }
+                        }
+                    } catch (Exception e) {
+                        LOGGER.warning("Error while processing sync components in asyn mode: " + e.getMessage());
```
jglick, line 521:
```suggestion
                        LOGGER.warning("Error while processing sync components in async mode: " + e.getMessage());
```
or actually do not swallow stack traces
```suggestion
                        LOGGER.log(Level.WARNING, "Error while processing sync components in async mode", e);
```

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -444,6 +488,42 @@ public static void writeBundle(
...
+                    } finally {
+                        binaryOut.closeEntry();
```
jglick, line 523: Unnecessary, though existing code uses this idiom. (Calling `putNextEntry` later suffices, or `close` implicit in a `try`-with-resources block in the case of the last entry.)

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -596,6 +676,8 @@ private static List<Content> appendManifestContents(
 
         manifest.append("Requested components:\n\n");
         ContentContainer contentsContainer = new ContentContainer(contentFilter, components);
+        int totalComponents = components.size();
+        int curerrentItration = 0;
```
jglick, line 680: typos

src/main/java/com/cloudbees/jenkins/support/api/ComponentVisitor.java
```diff
@@ -12,5 +12,5 @@ public interface ComponentVisitor {
      * @param component the {@link Component} being visited
      * @param <T> must be a subclass of {@link Component}
      */
-    <T extends Component> void visit(Container container, T component);
+    <T extends Component> void visit(Container container, T component, double progress);
```
jglick, line 15: Does not really belong here. https://github.com/jenkinsci/support-core-plugin/blob/8944371be83714ab937fe76e135d3f40a8ed4893/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java#L396 could instead be calculating the progress.

src/main/java/com/cloudbees/jenkins/support/actions/SupportObjectAction.java
```diff
@@ -141,12 +141,16 @@ public final void doGenerateAndDownload(StaplerRequest2 req, StaplerResponse2 rs
...
+                SupportPlugin.writeBundle(
+                        rsp.getOutputStream(),
+                        components,
+                        new ComponentVisitor() {
+                            @Override
+                            public <C extends Component> void visit(Container container, C component, double progress) {
+                                ((ObjectComponent<T>) component).addContents(container, object);
+                            }
+                        },
+                        null);
```
jglick, line 153: Note that these per-component **Support** links do _not_ use background generation. Probably fine; these are typically very quick to run.

jglick (changes requested):
Seems to work well. On hold: https://github.com/jenkinsci/support-core-plugin/pull/617#discussion_r1956593950

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
nevingeorgesunny: fixed

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
nevingeorgesunny: done

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
nevingeorgesunny: removed

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
nevingeorgesunny: I have set `SupportPlugin.setRequesterAuthentication` back

Maybe I read the comments of https://github.com/jenkinsci/support-core-plugin/pull/623 wrong 
From understanding  ,

we need` SupportPlugin.setRequesterAuthentication` to set the auth context for the flow , this is only checked in 
https://github.com/jenkinsci/support-core-plugin/blob/17f9e6937564fc2af840d10d29c863279d646a0b/src/main/java/com/cloudbees/jenkins/support/impl/AboutUser.java#L39C1-L40C1
as far as i can

right ?

and the seconds point 
Impersonating as SYSTEM  should not be done in the UI generate bundle flow 
it should only be for the periodic bundle generations  

as of now I removed Impersonating as SYSTEM in the asyn and syn flow 

once https://github.com/jenkinsci/support-core-plugin/pull/623 wrong  is merged I will remove ` SupportPlugin.setRequesterAuthentication` usage

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
nevingeorgesunny: used here https://github.com/jenkinsci/support-core-plugin/blob/17f9e6937564fc2af840d10d29c863279d646a0b/src/main/java/com/cloudbees/jenkins/support/SupportAction.java#L553

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
nevingeorgesunny: now im passing a boolean to skip this check

src/main/java/com/cloudbees/jenkins/support/api/ComponentVisitor.java
nevingeorgesunny: I have moved this logic to https://github.com/jenkinsci/support-core-plugin/blob/9b57a88922507d51963e8be4b8a1b897340011b9/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java#L424

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -92,6 +104,12 @@ public class SupportAction implements RootAction, StaplerProxy {
      */
     private final Logger logger = Logger.getLogger(SupportAction.class.getName());
 
+    private static final Path SUPPORT_BUNDLE_CREATION_FOLDER =
+            Paths.get(System.getProperty("java.io.tmpdir")).resolve("support-bundle");
+    public static final String SYNC_SUPPORT_BUNDLE = "support-bundle.zip";
+
+    private static final Map<UUID, SupportBundleAsyncGenerator> generatorByTaskId = new ConcurrentHashMap<>();
```
jglick, line 111:
```suggestion
    private final Map<UUID, SupportBundleAsyncGenerator> generatorByTaskId = new ConcurrentHashMap<>();
```
(it is a field of a singleton, so still just one, but works better with some `JenkinsRule` tests)

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -318,6 +336,60 @@ public void doGenerateAllBundles(StaplerRequest2 req, StaplerResponse2 rsp) thro
...
+        // so process them first and then process the other components async
+        List<Component> syncComponent =
+                components.stream().filter(c -> !c.canBeGeneratedAsync()).toList();
+        if (!syncComponent.isEmpty()) {
+            File outputDir = new File(SUPPORT_BUNDLE_CREATION_FOLDER + "/" + taskId);
+            if (!outputDir.exists()) {
+                if (!outputDir.mkdirs()) {
+                    throw new IOException("Failed to create directory: " + outputDir.getAbsolutePath());
+                }
+            }
```
jglick, line 372: Use `Files`.

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -439,4 +511,104 @@ public boolean isSelected() {
             return selected;
         }
     }
+
+    public ProgressiveRendering getGeneratorByTaskId(String taskId) throws Exception {
+        return generatorByTaskId.get(UUID.fromString(taskId));
+    }
+
+    public static class SupportBundleAsyncGenerator extends ProgressiveRendering {
+        private final Logger logger = Logger.getLogger(SupportBundleAsyncGenerator.class.getName());
```
jglick, line 520: Should just use the one from the outer class.

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -439,4 +511,104 @@ public boolean isSelected() {
...
+                return;
+            }
+
+            this.supportBundleName = BundleFileName.generate();
+            this.supportBundleGenerationInProgress = true;
+            logger.fine("Generating support bundle... task id " + taskId);
+            File outputDir = new File(SUPPORT_BUNDLE_CREATION_FOLDER + "/" + taskId);
+            if (!outputDir.exists()) {
+                if (!outputDir.mkdirs()) {
+                    throw new IOException("Failed to create directory: " + outputDir.getAbsolutePath());
```
jglick, line 546: ditto, one call in `Files`

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -439,4 +511,104 @@ public boolean isSelected() {
...
+        if (!bundleFile.exists()) {
+            rsp.sendError(HttpServletResponse.SC_NOT_FOUND, "Support bundle file not found");
+            return;
+        }
+
+        rsp.setContentType("application/zip");
+        rsp.addHeader("Content-Disposition", "attachment; filename=" + supportBundleName);
+        try (ServletOutputStream outputStream = rsp.getOutputStream();
+                FileInputStream inputStream = new FileInputStream(bundleFile)) {
+            IOUtils.copy(inputStream, outputStream);
```
jglick, line 593: there is now a method in `InputStream` itself for this

jglick: and for that matter you do not need to use `FileInputStream`, there is a method in `Files` to copy a `Path` to an `OutputStream`

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -352,12 +357,76 @@ public static void writeBundle(OutputStream outputStream) throws IOException {
...
+     * Generate a bundle for all synchronous components.
+     * This is useful when generating a support bundle asynchronously. There are some components that can't be generated
+     * asynchronously as they need some context from the request. So these components must be first processed synchronously
+     * and then the remaining components can be processed asynchronously.
+     *
+     * @param outputStream an {@link OutputStream}
+     * @param components a list of {@link Component} to include in the bundle
+     * @throws IOException if an error occurs while generating the bundle.
+     */
+    public static void writeBundleForSyncComponents(
```
jglick, line 383: Avoid making things `public` unless they really have to be.

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -352,12 +357,76 @@ public static void writeBundle(OutputStream outputStream) throws IOException {
...
+                },
+                null,
+                false);
+    }
+
+    /**
+     * Generate a bundle for all components that are selected in the Global Configuration.
+     *
+     * @param outputStream an {@link OutputStream}
+     * @param components a list of {@link Component} to include in the bundle
```
jglick, line 402: The summary sentence seems to contradict the `@param`?

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
jglick: Which is in the same package. So it could be using default access.

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -605,6 +699,7 @@ private static List<Content> appendManifestContents(
                 manifest.append("  * ").append(component.getDisplayName()).append("\n\n");
                 LOGGER.log(Level.FINE, "Start processing " + component.getDisplayName());
                 long startTime = System.currentTimeMillis();
+                // Calculate progress
```
jglick, line 702:
```suggestion
```
?

src/test/java/com/cloudbees/jenkins/support/SupportActionTest.java
```diff
@@ -316,6 +335,41 @@ private ZipFile downloadBundle(String s) throws IOException, SAXException {
...
+        HtmlForm form = p.getFormByName("bundle-contents");
+        HtmlButton submit = (HtmlButton) form.getElementsByTagName("button").get(0);
+        HtmlPage page = submit.click();
+
+        HtmlAnchor downloadLink = null;
+        File zipFile;
+        AtomicReference<HtmlPage> pageRef = new AtomicReference<>(page);
+
+        HtmlPage finalPage = page;
+        await().timeout(10, TimeUnit.HOURS).until(() -> {
```
jglick, line 350: Ten hours! The test will time out after three minutes.

src/test/java/com/cloudbees/jenkins/support/SupportActionTest.java
```diff
@@ -316,6 +335,41 @@ private ZipFile downloadBundle(String s) throws IOException, SAXException {
...
+            }
+        });
+
+        downloadLink = pageRef.get().getAnchorByText("click here");
+        if (downloadLink != null) {
+            // Download the zip file
+            WebResponse response = downloadLink.click().getWebResponse();
+            zipFile = File.createTempFile("downloaded", ".zip");
+            try (OutputStream os = Files.newOutputStream(zipFile.toPath())) {
+                IOUtils.copy(response.getContentAsStream(), os);
```
jglick, line 365: also an easier method in `Files`

jglick (approved):
Minor comments, but seems to work as advertised.

src/test/java/com/cloudbees/jenkins/support/SupportActionTest.java
nevingeorgesunny: I using Files.copy(... now

src/test/java/com/cloudbees/jenkins/support/SupportActionTest.java
nevingeorgesunny: Updated it to 10 seconds , I left it by mistake , was there for testing something

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
nevingeorgesunny: done

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
nevingeorgesunny: I update the summary to match the @param 
`Generate a bundle for the specified synchronous components. ` for synchronous and `Generate a bundle for the specified components ` for the generic

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -352,12 +357,76 @@ public static void writeBundle(OutputStream outputStream) throws IOException {
...
+                true);
+    }
+
+    /**
+     * Generate a bundle for all synchronous components.
+     * This is useful when generating a support bundle asynchronously. There are some components that can't be generated
+     * asynchronously as they need some context from the request. So these components must be first processed synchronously
+     * and then the remaining components can be processed asynchronously.
+     *
+     * @param outputStream an {@link OutputStream}
```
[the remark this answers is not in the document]
nevingeorgesunny: removed

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
nevingeorgesunny: made it to package private

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
nevingeorgesunny: im now using 
```
rsp.setContentType("application/zip");
rsp.addHeader("Content-Disposition", "attachment; filename=" + supportBundleName);
try (ServletOutputStream outputStream = rsp.getOutputStream()) {
    Files.copy(bundleFile.toPath(), outputStream);
}
        
```

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -439,4 +511,104 @@ public boolean isSelected() {
...
+                                FileUtils.deleteDirectory(outputDir);
+                                generatorByTaskId.remove(taskId);
+                                logger.fine(() -> "Cleaned up temporary directory " + outputDir);
+
+                            } catch (IOException e) {
+                                logger.log(Level.WARNING, () -> "Unable to delete " + outputDir);
+                            }
+                        },
+                        1,
+                        TimeUnit.HOURS);
```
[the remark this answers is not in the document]
nevingeorgesunny: I was thinking that , what if the user left the page open and left it it there for a while , the download button will not be valid if I immediately  delete , But makes sense that 1 hour can be a bit too long ,

I update it to 15 min

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
nevingeorgesunny: using Files. to achieve this

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
nevingeorgesunny: update it to use `SupportAction `

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
nevingeorgesunny: update it to use Files

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
nevingeorgesunny: I have removed it for generatorByTaskId

but im using a few static fileds like 
```
private static final Path SUPPORT_BUNDLE_CREATION_FOLDER =
        Paths.get(System.getProperty("java.io.tmpdir")).resolve("support-bundle");
public static final String SYNC_SUPPORT_BUNDLE = "support-bundle.zip";
  ```
  
  which are being used in other classes

Comment (jglick) [reply src/main/java/com/cloudbees/jenkins/support/SupportAction.java]: Sure but those are not mutable state.

File: src/main/java/com/cloudbees/jenkins/support/SupportAction.java [hunk 55]
```diff
@@ -364,13 +364,16 @@ public HttpRedirect doGenerateBundleAsync(StaplerRequest2 req, StaplerResponse2
...
-            File outputDir = new File(SUPPORT_BUNDLE_CREATION_FOLDER + "/" + taskId);
-            if (!outputDir.exists()) {
-                if (!outputDir.mkdirs()) {
-                    throw new IOException("Failed to create directory: " + outputDir.getAbsolutePath());
+            Path outputDir = SUPPORT_BUNDLE_CREATION_FOLDER.resolve(taskId.toString());
+            if (!Files.exists(outputDir)) {
+                try {
+                    Files.createDirectories(outputDir);
+                } catch (IOException e) {
+                    throw new IOException("Failed to create directory: " + outputDir.toAbsolutePath(), e);
```
jglick, line 372: Probably unnecessary to wrap and rethrow; the original `IOException` would include sufficient details.

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -364,13 +364,16 @@ public HttpRedirect doGenerateBundleAsync(StaplerRequest2 req, StaplerResponse2
...
                 }
             }
-            try (FileOutputStream fileOutputStream = new FileOutputStream(new File(outputDir, SYNC_SUPPORT_BUNDLE))) {
+            try (FileOutputStream fileOutputStream =
+                    new FileOutputStream(new File(outputDir.toString(), SYNC_SUPPORT_BUNDLE))) {
```
jglick, line 376: Preferable to use `Files.newOutputStream`.

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -517,7 +520,7 @@ public ProgressiveRendering getGeneratorByTaskId(String taskId) throws Exception
     }
 
     public static class SupportBundleAsyncGenerator extends ProgressiveRendering {
-        private final Logger logger = Logger.getLogger(SupportBundleAsyncGenerator.class.getName());
+        private final Logger logger = Logger.getLogger(SupportAction.class.getName());
```
jglick, line 523: No I meant you could just use https://github.com/jenkinsci/support-core-plugin/blob/effed5e2e7f6fcf6edeb6e562795bd3524c4c9ed/src/main/java/com/cloudbees/jenkins/support/SupportAction.java#L105

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -540,18 +543,20 @@ protected void compute() throws Exception {
             this.supportBundleName = BundleFileName.generate();
             this.supportBundleGenerationInProgress = true;
             logger.fine("Generating support bundle... task id " + taskId);
-            File outputDir = new File(SUPPORT_BUNDLE_CREATION_FOLDER + "/" + taskId);
-            if (!outputDir.exists()) {
-                if (!outputDir.mkdirs()) {
-                    throw new IOException("Failed to create directory: " + outputDir.getAbsolutePath());
+            Path outputDir = SUPPORT_BUNDLE_CREATION_FOLDER.resolve(taskId.toString());
+            if (!Files.exists(outputDir)) {
```
jglick, line 547: ditto

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -540,18 +543,20 @@ protected void compute() throws Exception {
...
+                try {
+                    Files.createDirectories(outputDir);
+                } catch (IOException e) {
+                    throw new IOException("Failed to create directory: " + outputDir.toAbsolutePath(), e);
```
jglick, line 551: ditto

src/main/java/com/cloudbees/jenkins/support/SupportAction.java
```diff
@@ -588,9 +593,8 @@ public void doDownloadBundle(@QueryParameter("taskId") String taskId, StaplerRes
 
         rsp.setContentType("application/zip");
         rsp.addHeader("Content-Disposition", "attachment; filename=" + supportBundleName);
-        try (ServletOutputStream outputStream = rsp.getOutputStream();
-                FileInputStream inputStream = new FileInputStream(bundleFile)) {
-            IOUtils.copy(inputStream, outputStream);
+        try (ServletOutputStream outputStream = rsp.getOutputStream()) {
```
jglick, line 596: stylistically, just
```suggestion
        try (OutputStream outputStream = rsp.getOutputStream()) {
```
or
```suggestion
        try (var outputStream = rsp.getOutputStream()) {
```
suffices

src/test/java/com/cloudbees/jenkins/support/SupportActionTest.java
```diff
@@ -347,7 +345,7 @@ private ZipFile downloadSupportBundleByButtonClick() throws IOException, SAXExce
         AtomicReference<HtmlPage> pageRef = new AtomicReference<>(page);
 
         HtmlPage finalPage = page;
-        await().timeout(10, TimeUnit.HOURS).until(() -> {
+        await().timeout(10, TimeUnit.SECONDS).until(() -> {
```
jglick, line 348: or simply
```suggestion
        await().until(() -> {
```
since this is the default timeout

PR #618, opened 2025-01-28
[JENKINS-73599] Externalize Proxy details in its own component + fix markdowns

Dohbedoh:
[JENKINS-73599](https://issues.jenkins.io/browse/JENKINS-73599). Proxy Configuration and Update Center are now separated in Jenkins core so it makes sense to separate the information in the bundle as well.

### Submitter checklist

### DIFF
--- a/src/test/java/com/cloudbees/jenkins/support/impl/ProxyConfigurationTest.java
+++ b/src/test/java/com/cloudbees/jenkins/support/impl/ProxyConfigurationTest.java
@@ -1,0 +1,47 @@
+package com.cloudbees.jenkins.support.impl;
+
+import static org.hamcrest.CoreMatchers.containsString;
+import static org.hamcrest.CoreMatchers.not;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+import com.cloudbees.jenkins.support.SupportTestUtils;
+import com.cloudbees.jenkins.support.api.Component;
+import hudson.ExtensionList;
+import hudson.ProxyConfiguration;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+import org.junit.Rule;
+import org.junit.Test;
+import org.jvnet.hudson.test.Issue;
+import org.jvnet.hudson.test.JenkinsRule;
+
+public class ProxyConfigurationTest {
+
+    @Rule
+    public JenkinsRule j = new JenkinsRule();
+
+    @Test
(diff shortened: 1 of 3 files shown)


jglick:
https://github.com/jenkinsci/plugin-pom?tab=readme-ov-file#format-as-you-build

src/main/java/com/cloudbees/jenkins/support/impl/UpdateCenter.java
```diff
@@ -40,25 +37,23 @@ public String getDisplayName() {
...
-
-                    out.println("======");
-
-                    out.println("Last updated: " + updateCenter.getLastUpdatedString());
-
-                    addProxyInformation(out, filter);
-                } finally {
-                    out.flush();
+            public void printTo(PrintWriter out, @NonNull ContentFilter filter) {
+                hudson.model.UpdateCenter updateCenter = Jenkins.get().getUpdateCenter();
```
jglick, line 41: (ignore WS)

src/test/java/com/cloudbees/jenkins/support/impl/ProxyConfigurationTest.java
```diff
@@ -0,0 +1,47 @@
...
+                1234,
+                "proxyUser",
+                "proxyPass",
+                String.join("\n", noProxyHosts),
+                "http://localhost:8080"));
+
+        String ucMdToString =
+                SupportTestUtils.invokeComponentToString(Objects.requireNonNull(ExtensionList.lookup(Component.class)
+                        .get(com.cloudbees.jenkins.support.impl.ProxyConfiguration.class)));
+        assertThat(ucMdToString, containsString(" - Host: `proxy.server.com`"));
```
jglick, line 39: Perhaps off topic, but have we considered throwing out the convention of formatting files in support bundles using Markdown, and switching to YAML?

jglick (approved):

src/test/java/com/cloudbees/jenkins/support/impl/ProxyConfigurationTest.java
jglick: @PierreBtz we can certainly make such a change in the plugin, but does @cloudbees have some internal tool which parses the current text files? If so, it would need to be modified (simplified, I hope).

PierreBtz: Yes we do... and I don't like parsing markdown :)

PR #623, opened 2025-02-10
Removing `requesterAuthentication`

jglick:
Reviewing #617, I was alarmed to find that this plugin defaults to impersonating `SYSTEM` during interactive bundle generation, and (as of the old a6de1be6ea1e3b77ea7bbb6c6893745fbf1d11ce) relies on the author of a `Component` to actually check the requester’s authentication where applicable. This goes against the principle of least surprise and safety by default. In fact most components require `ADMINISTER` anyway (the option to do otherwise dates all the way back to 066bd8d33a88b7866b23c20f66ec56313f57924e), but not all. I could not find any actually problematic cases, but for example https://github.com/jenkinsci/support-core-plugin/blob/3435ec4057ae19e04dee06d2eba8c5f1128e603e/src/main/java/com/cloudbees/jenkins/support/impl/ItemsContent.java#L87 is uncomfortable because it is reporting _statistics_ about the system based on details not accessible to the user. I think we should only use `SYSTEM` during periodic bundle generation. Not sure if this has any practical impact, since 6b177ea7cc7347e13fa87174472400bbbe78d422 and then 84ae1b3a1d326cabcfd755c92c6a4f949a937ffe locked down bundle generation anyway (so why do we offer some components to `MANAGE`?!), even in #198 AFAICT.

### DIFF
--- a/src/test/java/com/cloudbees/jenkins/support/CheckFilterTest.java
+++ b/src/test/java/com/cloudbees/jenkins/support/CheckFilterTest.java
@@ -28,6 +28,7 @@ import hudson.model.FreeStyleProject;
 import hudson.model.User;
 import hudson.model.labels.LabelAtom;
 import hudson.model.queue.QueueTaskFuture;
+import hudson.security.ACL;
 import java.io.BufferedInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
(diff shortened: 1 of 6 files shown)


src/test/java/com/cloudbees/jenkins/support/CheckFilterTest.java
```diff
@@ -219,24 +222,25 @@ private static String getFirstEnvVar() {
...
-            fileSet.add(of("items.md", "jobs", false));
+            // fileSet.add(of("items.md", "jobs", false));
             // checksum.md5 is not generated in tests
 
             // AboutJenkins -> nodes.md
             // The agent node name should be filtered
             fileSet.add(of("nodes.md", AGENT_NAME, true));
             // AboutUser
-            fileSet.add(of("user.md", "anonymous", true));
+            fileSet.add(of("user.md", "SYSTEM", true));
```
jglick, line 238: Did not in fact include `anonymous` before; did not create `user.md` at all.

src/test/java/com/cloudbees/jenkins/support/CheckFilterTest.java
```diff
@@ -219,24 +226,25 @@ private static String getFirstEnvVar() {
...
             fileSet.add(of("about.md", "b", false));
-            fileSet.add(of("items.md", "jobs", false));
+            // fileSet.add(of("items.md", "jobs", false));
             // checksum.md5 is not generated in tests
 
             // AboutJenkins -> nodes.md
             // The agent node name should be filtered
             fileSet.add(of("nodes.md", AGENT_NAME, true));
             // AboutUser
-            fileSet.add(of("user.md", "anonymous", true));
```
jglick, line 234: Never was used before. Now using `unchecked` to catch such mistakes in the test.

nevingeorgesunny (approved):
LGTM

PR #624, opened 2025-02-14
Stop agents before exiting test cases, and reliably close `all_*.log` when agents exit

jglick:
As noted in https://github.com/jenkinsci/support-core-plugin/pull/617#issuecomment-2659388971 there are mostly consistent test failures on Windows. Not sure why these are just appearing now (and I cannot reproduce locally on Windows 10), but at any rate `SupportLogHandler` opens a file handle on e.g. `agent-work-dirs\agent0\support\all_2025-02-14_13.33.29.log` which will not be closed unless the agent process exits.

([CloudBees internal issue](https://cloudbees.atlassian.net/browse/BEE-55699))


src/test/java/com/cloudbees/jenkins/support/CheckFilterTest.java
```diff
@@ -75,6 +76,14 @@ public class CheckFilterTest {
     @Rule
     public LoggerRule logging = new LoggerRule().record(AsyncResultCache.class, Level.FINER);
 
+    @After
+    public void stopAgents() throws Exception {
```
jglick, line 80: Maybe this is something that should be added to `JenkinsRule`? Compare https://github.com/jenkinsci/jenkins-test-harness/pull/643.

jglick:
Apparently did not suffice. Maybe this file is not actually closed when the agent terminates? Or the current attempt to terminate the agent is not effective because it just closes the connection and does not exit the JVM?

jglick:
https://github.com/jenkinsci/support-core-plugin/blob/7fcc00b4544597652056b0d7d28fb15ac32fa969/src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java#L879 initializes a log handler which opens this file handle in https://github.com/jenkinsci/support-core-plugin/blob/7fcc00b4544597652056b0d7d28fb15ac32fa969/src/main/java/com/cloudbees/jenkins/support/SupportLogHandler.java#L242 and AFAICT an agent would not normally close this handle until the JVM exits. https://github.com/jenkinsci/jenkins-test-harness/blob/052a595c77a54099ceeac0002c0d4f816838092e/src/main/java/org/jvnet/hudson/test/SimpleCommandLauncher.java#L82-L91 ought to be terminating that JVM, but I guess this just _initiates_ termination of the JVM, it does not _wait_ for it.

jglick:
Incorrect. `SimpleCommandLauncher` seems to be waiting for the process to exit, but its listener is not being called synchronously by this thread:

```
java.lang.Throwable
	at org.jvnet.hudson.test.SimpleCommandLauncher$1.onClosed(SimpleCommandLauncher.java:85)
	at hudson.remoting.Channel.terminate(Channel.java:1219)
	at hudson.remoting.Channel$CloseCommand.execute(Channel.java:1438)
	at hudson.remoting.Channel$1.handle(Channel.java:664)
	at hudson.remoting.SynchronousCommandTransport$ReaderThread.run(SynchronousCommandTransport.java:86)
```

jglick:
Almost, but [not quite](https://github.com/jenkinsci/support-core-plugin/pull/624/checks?check_run_id=37432234654).

MarkEWaite (dismissed):
Thanks very much for the investigation!

jglick:
So [this failure](https://github.com/jenkinsci/support-core-plugin/runs/37435427493) is puzzling me.

```
11.263 [id=227]	INFO	c.c.j.support.SupportTestUtils#stopAgents: successfully deleted C:\Jenkins\agent\workspace\57ae19e04dee06d2eba8c5f1128e603e\target\tmp\j h17291846234312879372\agent-work-dirs
…
11.374 [id=227]	INFO	o.j.h.t.TemporaryDirectoryAllocator#dispose: deleting C:\Jenkins\agent\workspace\57ae19e04dee06d2eba8c5f1128e603e\target\tmp\j h17291846234312879372
…
java.nio.file.FileSystemException: C:\Jenkins\agent\workspace\57ae19e04dee06d2eba8c5f1128e603e\target\tmp\j h17291846234312879372\agent-work-dirs\slave0\support\all_2025-02-18_23.39.21.log: The process cannot access the file because it is being used by another process
```

would seem to indicate that something _recreated_ this file after its containing directory had been deleted.

jglick:
Two tracks now: https://github.com/jenkinsci/jenkins-test-harness/pull/922; and trying to see if I can get `SupportLogHandler.close` to be called when the connection is closed.

jglick:
No luck so far on the latter approach:

```diff
diff --git src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
index 3e57d01..c09f9d6 100644
--- src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
+++ src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
@@ -57,6 +57,7 @@ import hudson.model.Descriptor;
 import hudson.model.Node;
 import hudson.model.PeriodicWork;
 import hudson.model.TaskListener;
+import hudson.remoting.Channel;
 import hudson.remoting.ChannelClosedException;
 import hudson.remoting.Future;
 import hudson.remoting.VirtualChannel;
@@ -877,6 +878,12 @@ public class SupportPlugin extends Plugin {
             LogHolder.AGENT_LOG_HANDLER.setLevel(level);
             LogHolder.AGENT_LOG_HANDLER.setDirectory(new File(rootPath.getRemote(), SUPPORT_DIRECTORY_NAME), "all");
             ROOT_LOGGER.addHandler(LogHolder.AGENT_LOG_HANDLER);
+            Channel.currentOrFail().addListener(new Channel.Listener() {
+                @Override
+                public void onClosed(Channel channel, IOException cause) {
+                    LogHolder.AGENT_LOG_HANDLER.close();
+                }
+            });
             return null;
         }
     }
```

So far as I can tell, `onClosed` does not get called.

jglick:
[Still failing](https://github.com/jenkinsci/support-core-plugin/pull/624/checks?check_run_id=37440112487) so I am running out of hypotheses.

jglick:
[With better logs](https://github.com/jenkinsci/support-core-plugin/pull/624/checks?check_run_id=37441880269) we can see that the agent process is terminated (exit code zero), and yet 59ms later the attempt to delete its working directory fails. [This remark](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-lockfile#remarks)

> If a process terminates with a portion of a file locked or closes a file that has outstanding locks, the locks are unlocked by the operating system. However, the time it takes for the operating system to unlock these locks depends upon available system resources. Therefore, it is recommended that your process explicitly unlock all files it has locked when it terminates. If this is not done, access to these files may be denied if the operating system has not yet unlocked them.

suggests that https://github.com/jenkinsci/jenkins-test-harness/pull/922 (while probably a step in the right direction) cannot suffice. Either I figure out how to make the agent release its locks before it exits, or the deletion code needs to just retry for a while.

src/test/java/com/cloudbees/jenkins/support/CheckFilterTest.java
jglick: No, actually `Jenkins.cleanUp` was already doing this.

jglick:
Seems like https://github.com/jenkinsci/support-core-plugin/blob/e23e0bae8e5f6a0ad15789205ad975b5076c5f1b/src/main/java/com/cloudbees/jenkins/support/SafeLog.java#L38-L70 could be useful to package as a utility somewhere, but not sure where. Definitely not the first time I have needed to robustly collect log messages from both controller & agent for diagnostic purposes.

pom.xml
```diff
@@ -70,6 +70,8 @@
     <!-- https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/ -->
     <jenkins.baseline>2.479</jenkins.baseline>
     <jenkins.version>${jenkins.baseline}.1</jenkins.version>
+    <!-- TODO until in parent -->
+    <jenkins-test-harness.version>2403.v256947ecb_c8a_</jenkins-test-harness.version>
```
jglick, line 74: https://github.com/jenkinsci/jenkins-test-harness/pull/922

src/main/java/com/cloudbees/jenkins/support/SupportLogHandler.java
```diff
@@ -234,6 +234,7 @@ private void setFile(File file) throws FileNotFoundException {
                 parentFile.mkdirs();
             }
 
+            StreamUtils.closeQuietly(null); // ensure class is loaded so close() can succeed
```
jglick, line 237: Does nothing but loads this class. Otherwise there can occasionally be a `NoClassDefFoundError` during agent shutdown, at which time it is too late to load new classes from `RemoteClassLoader` because the connection has already been closed.

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -862,6 +862,15 @@ public Void call() {
             // avoid double installation of the handler. JNLP agents can reconnect to the controller multiple times
             // and each connection gets a different RemoteClassLoader, so we need to evict them by class name,
             // not by their identity.
+            closeAll();
+            Runtime.getRuntime().addShutdownHook(new Thread(LogInitializer::closeAll, "close log handlers"));
```
jglick, line 866: Ensures that the file handle is closed before the JVM exits. Likely irrelevant on Linux, but seems to matter on Windows, where the OS will release mandatory file locks “sometime” after the process exits but not necessarily quickly enough for a test.

dwnusbaum (approved):

MarkEWaite (approved):
Thanks very, very much!

jglick:
[Yet another flake](https://github.com/jenkinsci/support-core-plugin/runs/37494222455) :eyes:

jtnord (approved):

jtnord (approved):

dwnusbaum (approved):

PR #642, opened 2025-04-03
Added plugin name terms and more common terms to stop words

Dohbedoh:
Adding more stop words after looking at a recent anonymized bundle, and also terms from installed plugin names.

### Submitter checklist

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/filter/PluginsStopWords.java
+++ b/src/main/java/com/cloudbees/jenkins/support/filter/PluginsStopWords.java
@@ -1,0 +1,25 @@
+package com.cloudbees.jenkins.support.filter;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import hudson.Extension;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Locale;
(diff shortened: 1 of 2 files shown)


Dohbedoh:
Wondering if maybe we could / should add the descriptor symbols (with `org.jenkinsci.plugins.structs.SymbolLookup`)..

src/main/java/com/cloudbees/jenkins/support/filter/PluginsStopWords.java
```diff
@@ -0,0 +1,25 @@
...
+import java.util.Set;
+import jenkins.model.Jenkins;
+
+@Extension
+public class PluginsStopWords implements StopWords {
+    @NonNull
+    @Override
+    public Set<String> getWords() {
+        final Set<String> pluginsWords = new HashSet<>();
+        Jenkins.get().pluginManager.getPlugins().forEach(pluginWrapper -> {
```
jglick, line 17: Note that this will include in-house plugins, but those would show up elsewhere anyway.

jglick (approved):

PR #646, opened 2025-04-17
Collect information about instance startup and shutdown

Vlatombe:
Collect timings for each init milestone
If the startup takes more than a defined initial delay, start collecting thread dumps at regular intervals, until the startup completes.

We consider that an instance taking more than 5 minutes is slow to start and is worth collecting thread dumps for (`com.cloudbees.jenkins.support.startup.StartupReport.INITIAL_DELAY_SECONDS`).
Then we take one thread dump every 30 seconds (`com.cloudbees.jenkins.support.startup.StartupReport.RECURRENCE_PERIOD_SECONDS`)

The timings are collected in `startup-timings.md` with the following format (values for a development instance)

---

| Milestone                          | Time     |
|------------------------------------|----------|
| Started initialization             | No data  |
| Listed all plugins                 | No data  |
| Prepared all plugins               | 17 sec   |
| Started all plugins                | 8.2 sec  |
| Augmented all extensions           | 0.7 sec  |
| System config loaded               | 1.8 sec  |
| System config adapted              | 0.16 sec |
| Loaded all jobs                    | 1.7 sec  |
| Configuration for all jobs updated | 0.44 sec |

Total startup time : 30 sec

---

I will also add a similar logic to collect information when the instance terminates and takes more than the expected time.

[CloudBees Internal issue](https://cloudbees.atlassian.net/browse/BEE-57504)

### Submitter checklist
- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/ThreadDumps.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/ThreadDumps.java
@@ -80,18 +80,19 @@ public class ThreadDumps extends ObjectComponent<Computer> {
             throw new IllegalArgumentException("delay must be greater than 0");
         }
         var format = new SimpleDateFormat(DATE_FORMAT);
-        for (int i = 0; i < iterations; i++) {
-            File threadDumpFile = fileList.file(format.format(new Date(timestamp)) + ".txt");
-            try (FileOutputStream fileOutputStream = new FileOutputStream(threadDumpFile)) {
-                threadDump(fileOutputStream);
-                fileList.add(threadDumpFile);
-                Thread.sleep(delay);
-            } catch (IOException ioe) {
-                LOGGER.log(loggingLevel, "Support Core plugin can't generate automatically thread dump", ioe);
-            } catch (InterruptedException ie) {
-                LOGGER.log(loggingLevel, "The thread was interrupted by unknown reasons. It may be a bug", ie);
-            } finally {
-                timestamp += delay;
+        File threadDumpFile = fileList.file(format.format(new Date(timestamp)) + ".txt");
+        try (FileOutputStream fileOutputStream = new FileOutputStream(threadDumpFile)) {
+            threadDump(fileOutputStream);
+            fileList.add(threadDumpFile);
+        } catch (IOException ioe) {
+            LOGGER.log(loggingLevel, "Failed to generate thread dump", ioe);
+        } finally {
+            if (iterations > 1) {
+                Timer.get()
+                        .schedule(
+                                () -> collectMultiple(fileList, timestamp + delay, delay, iterations - 1, loggingLevel),
+                                delay,
+                                TimeUnit.MILLISECONDS);
             }
         }
     }
--- a/src/main/java/com/cloudbees/jenkins/support/impl/ThreadDumps.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/ThreadDumps.java
@@ -63,6 +63,9 @@ public class ThreadDumps extends ObjectComponent<Computer> {
      * @param iterations the number of thread dumps to collect
      */
     public static void collectMultiple(FileListCap fileList, long timestamp, long delay, int iterations) {
+        if (delay <= 0) {
+            throw new IllegalArgumentException("delay must be greater than 0");
+        }
         var format = new SimpleDateFormat(DATE_FORMAT);
         for (int i = 0; i < iterations; i++) {
             File threadDumpFile = fileList.file(format.format(new Date(timestamp)) + ".txt");
--- a/src/main/java/com/cloudbees/jenkins/support/impl/ThreadDumps.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/ThreadDumps.java
@@ -63,6 +63,19 @@ public class ThreadDumps extends ObjectComponent<Computer> {
      * @param iterations the number of thread dumps to collect
      */
     public static void collectMultiple(FileListCap fileList, long timestamp, long delay, int iterations) {
+        collectMultiple(fileList, timestamp, delay, iterations, Level.WARNING);
+    }
+
+    /**
+     * Collects multiple thread dumps and saves them to the specified file list.
+     * @param fileList the file list where the thread dumps will be saved
+     * @param timestamp the initial timestamp
+     * @param delay the delay between two consecutive thread dumps
+     * @param iterations the number of thread dumps to collect
+     * @param loggingLevel the logging level to use for logging errors
+     */
+    public static void collectMultiple(
+            FileListCap fileList, long timestamp, long delay, int iterations, Level loggingLevel) {
         if (delay <= 0) {
             throw new IllegalArgumentException("delay must be greater than 0");
         }
@@ -74,9 +87,9 @@ public class ThreadDumps extends ObjectComponent<Computer> {
                 fileList.add(threadDumpFile);
                 Thread.sleep(delay);
             } catch (IOException ioe) {
-                LOGGER.log(Level.WARNING, "Support Core plugin can't generate automatically thread dump", ioe);
+                LOGGER.log(loggingLevel, "Support Core plugin can't generate automatically thread dump", ioe);
             } catch (InterruptedException ie) {
-                LOGGER.log(Level.WARNING, "The thread was interrupted by unknown reasons. It may be a bug", ie);
+                LOGGER.log(loggingLevel, "The thread was interrupted by unknown reasons. It may be a bug", ie);
             } finally {
                 timestamp += delay;
             }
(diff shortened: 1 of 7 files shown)


src/main/java/com/cloudbees/jenkins/support/impl/ThreadDumps.java
```diff
@@ -45,11 +48,41 @@
...
+     * @param delay the delay between two consecutive thread dumps
+     * @param iterations the number of thread dumps to collect
+     */
+    public static void collectMultiple(FileListCap fileList, long timestamp, long delay, int iterations) {
+        if (delay <= 0) {
+            throw new IllegalArgumentException("delay must be greater than 0");
+        }
+        var format = new SimpleDateFormat(DATE_FORMAT);
+        for (int i = 0; i < iterations; i++) {
+            File threadDumpFile = fileList.file(format.format(new Date(timestamp)) + ".txt");
```
Vlatombe, line 71: Suppressed the prefix `-i`as the timestamp is always different anyway

src/main/java/com/cloudbees/jenkins/support/impl/ThreadDumps.java
```diff
@@ -45,11 +48,41 @@
 @Extension(ordinal = -100.0) // run this last as it blocks the channel
 public class ThreadDumps extends ObjectComponent<Computer> {
 
-    private final Logger logger = Logger.getLogger(ThreadDumps.class.getName());
```
Vlatombe, line 48: Long due

src/main/java/com/cloudbees/jenkins/support/impl/ThreadDumps.java
jglick: (referring to code moved from `SlowRequestThreadDumpsGenerator`)

src/main/java/com/cloudbees/jenkins/support/impl/ThreadDumps.java
```diff
@@ -45,11 +48,54 @@
...
+        if (delay <= 0) {
+            throw new IllegalArgumentException("delay must be greater than 0");
+        }
+        var format = new SimpleDateFormat(DATE_FORMAT);
+        for (int i = 0; i < iterations; i++) {
+            File threadDumpFile = fileList.file(format.format(new Date(timestamp)) + ".txt");
+            try (FileOutputStream fileOutputStream = new FileOutputStream(threadDumpFile)) {
+                threadDump(fileOutputStream);
+                fileList.add(threadDumpFile);
+                Thread.sleep(delay);
```
jglick, line 88: Should be using `Timer` to (re)schedule a task rather than blocking a thread like this.

src/main/java/com/cloudbees/jenkins/support/impl/ThreadDumps.java
```diff
@@ -45,11 +48,54 @@
...
+            } catch (IOException ioe) {
+                LOGGER.log(loggingLevel, "Support Core plugin can't generate automatically thread dump", ioe);
```
jglick, line 90: while we are here, this is bad English and too verbose
```suggestion
                LOGGER.log(loggingLevel, "Failed to generate thread dump", ioe);
```

src/main/java/com/cloudbees/jenkins/support/startup/ShutdownComponent.java
```diff
@@ -0,0 +1,109 @@
...
+@Extension
+public class ShutdownComponent extends UnfilteredFileListCapComponent {
+    private static final Logger LOGGER = Logger.getLogger(ShutdownComponent.class.getName());
+
+    private static final int THREAD_DUMPS_TO_RETAIN =
+            SystemProperties.getInteger(ShutdownComponent.class.getName() + ".THREAD_DUMPS_TO_RETAIN", 40);
+    private static final int TOTAL_ITERATIONS =
+            SystemProperties.getInteger(ShutdownComponent.class.getName() + ".TOTAL_ITERATIONS", 4);
+    private static final int DELAY_BETWEEN_THREAD_DUMPS_MS =
+            SystemProperties.getInteger(ShutdownComponent.class.getName() + ".DELAY_BETWEEN_THREAD_DUMPS_MS", 1000);
```
jglick, line 58: I think we should introduce a method in `SystemProperties` to get a `Duration` where you can specify a default time unit but the parser would also accept explicit time units like `2s` etc.

src/main/java/com/cloudbees/jenkins/support/startup/ShutdownComponent.java
```diff
@@ -0,0 +1,109 @@
...
+    public static class OnShutdown extends ItemListener {
+        @Override
+        public void onBeforeShutdown() {
+            var logs = ShutdownComponent.get().logs;
+            Executors.newSingleThreadExecutor(new ExceptionCatchingThreadFactory(new NamingThreadFactory(
+                            new ClassLoaderSanityThreadFactory(new DaemonThreadFactory()),
+                            ShutdownComponent.class.getSimpleName())))
+                    .execute(() -> {
+                        try {
+                            Thread.sleep(TimeUnit.SECONDS.toMillis(INITIAL_DELAY));
```
jglick, line 92: Why not just `schedule` it later?

src/main/java/com/cloudbees/jenkins/support/startup/StartupComponent.java
```diff
@@ -0,0 +1,119 @@
...
+        private final Map<InitMilestone, Long> timePerMilestone;
+
+        StartupContent(Map<InitMilestone, Long> timePerMilestone) {
+            super("startup-timings.md");
+            this.timePerMilestone = timePerMilestone;
+        }
+
+        @Override
+        protected void printTo(PrintWriter out, @NonNull ContentFilter filter) {
+            var startTime = ManagementFactory.getRuntimeMXBean().getStartTime();
```
jglick, line 75: Beware that this is _JVM_ startup time, so if any servlet listener runs before Jenkins starts (two such examples come to mind in CloudBees CI), you will get a very skewed measurement.

src/main/java/com/cloudbees/jenkins/support/startup/StartupReport.java
```diff
@@ -0,0 +1,166 @@
...
+    private static final int DELAY_BETWEEN_THREAD_DUMPS_MS =
+            SystemProperties.getInteger(StartupReport.class.getName() + ".DELAY_BETWEEN_THREAD_DUMPS_MS", 1000);
+
+    private final FileListCap logs =
+            new FileListCap(new File(SupportPlugin.getRootDirectory(), "startup-threaddumps"), THREAD_DUMPS_TO_RETAIN);
+
+    /**
+     * Record the timestamp at which each milestone was reached.
+     */
+    private Map<InitMilestone, Long> timesPerMilestone = new ConcurrentHashMap<>();
```
jglick, line 71: Using `java.time` types would be more legible and less prone to confusing instants with durations.

jglick (commented):
Looks good so far.

src/main/java/com/cloudbees/jenkins/support/startup/StartupComponent.java
```diff
@@ -0,0 +1,119 @@
...
+        return ComponentCategory.CONTROLLER;
+    }
+
+    private static class StartupContent extends PrefilteredPrintedContent {
+        public static final String MILESTONE_HEADER = "Milestone";
+        public static final String TIME_HEADER = "Time";
+        private final Map<InitMilestone, Long> timePerMilestone;
+
+        StartupContent(Map<InitMilestone, Long> timePerMilestone) {
+            super("startup-timings.md");
```
jglick, line 69: Do I understand right that this one file is always added to the bundle, and it contains timing from this session? And additionally we will have a directory with timestamped thread dumps when startup has been slow (in any recent Jenkins session), similar to how `SlowRequestThreadDumpsGenerator` works?

Vlatombe: Yes, this is the idea.

src/main/java/com/cloudbees/jenkins/support/startup/ShutdownComponent.java
Vlatombe: A while ago I introduced https://github.com/jenkinsci/remoting/blob/3ea58fc6d3356698eb8b14210135149b8f1e6018/src/main/java/org/jenkinsci/remoting/util/DurationStyle.java#L38-L39 in remoting. Could unrestrict it so that Jenkins core can facade it through `SystemProperties`

jglick (approved):

src/main/java/com/cloudbees/jenkins/support/startup/ShutdownComponent.java
Vlatombe: https://github.com/jenkinsci/remoting/pull/800 + https://github.com/jenkinsci/jenkins/pull/10561

PR #659, opened 2025-05-26
remove usages of commons-lang from core

mawinter69:
To be able to remove commons-lang from core all direct and indirect usages must be replaced with commons-lang3 or native Java

use XS_DATETIME_FORMATTER2 instead of deprecated XS_DATETIME_FORMATTER from core
use commons-lang3 api plugin instead of commons-lang provided by core

### Submitter checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -118,6 +118,10 @@
     </dependency>
     <!-- plugin dependencies -->
     <dependency>
+      <groupId>io.jenkins.plugins</groupId>
+      <artifactId>commons-lang3-api</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>metrics</artifactId>
     </dependency>
--- a/src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
+++ b/src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
@@ -22,7 +22,7 @@ import javax.xml.transform.TransformerFactory;
 import javax.xml.transform.sax.SAXSource;
 import javax.xml.transform.stream.StreamResult;
 import org.apache.commons.io.FileUtils;
-import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.xml.sax.Attributes;
 import org.xml.sax.InputSource;
 import org.xml.sax.SAXException;
--- a/src/main/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilter.java
+++ b/src/main/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilter.java
@@ -37,7 +37,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Pattern;
 import java.util.stream.StreamSupport;
-import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
 
--- a/src/main/java/com/cloudbees/jenkins/support/impl/DirectoryComponent.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/DirectoryComponent.java
@@ -19,7 +19,7 @@ import java.util.Set;
 import java.util.logging.Logger;
 import java.util.stream.Stream;
 import jenkins.model.Jenkins;
-import org.apache.commons.lang.ArrayUtils;
+import org.apache.commons.lang3.ArrayUtils;
 import org.apache.tools.ant.DirectoryScanner;
 import org.apache.tools.ant.Project;
 import org.apache.tools.ant.types.FileSet;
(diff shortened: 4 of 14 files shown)


pom.xml
```diff
@@ -98,6 +98,11 @@
       <artifactId>support-log-formatter</artifactId>
       <version>1.3</version>
     </dependency>
+    <!-- plugin dependencies -->
+    <dependency>
+      <groupId>io.jenkins.plugins</groupId>
+      <artifactId>commons-lang3-api</artifactId>
```
jglick, line 104: Would prefer to drop this dep and just replace usages of `StringUtils` with simple equivalents from the Java Platform API.

mawinter69: There is one place that uses a complex function: `StringUtils.splitByCharacterTypeCamelCase(camelCase)`

jglick: https://github.com/apache/commons-lang/blob/7302e9a308f0a5d59f17f72b6f5326e76cd71e3a/src/main/java/org/apache/commons/lang3/StringUtils.java#L7196-L7226 FTR. I think https://github.com/jenkinsci/support-core-plugin/blob/a2a9f06eed6188fb78f636ad9b2e6ba5c048ea88/src/main/java/com/cloudbees/jenkins/support/impl/LoadStats.java#L318-L328 could just be deleted and the code name used directly.

jglick:
https://github.com/jenkinsci/plugin-pom/blob/master/README.md#format-as-you-build

src/main/java/com/cloudbees/jenkins/support/impl/LoadStats.java
```diff
@@ -314,16 +313,4 @@ public boolean shouldBeFiltered() {
...
         }
     }
-
-    /**
-     * Converts "aCamelCaseString" into "A sentence case string".
-     *
-     * @param camelCase camelCase.
-     * @return Sentence case.
-     */
-    private static String camelCaseToSentenceCase(String camelCase) {
```
Dohbedoh, line 324: Can we keep the camel case to sentence logic ? 
https://github.com/jenkinsci/jenkins/blob/24a2e7c78278e591706a572896c0cabd69f4d705/core/src/main/resources/hudson/model/Messages.properties#L374-L381.

Dohbedoh (approved):

src/main/java/com/cloudbees/jenkins/support/configfiles/SecretHandler.java
```diff
@@ -50,8 +49,8 @@ class SecretHandler {
      * FALLBACK will be disable in case you define a system property like -Dsupport-core-plugin.SecretHandler.ENABLE_FALLBACK=false
      * Otherwise will be enabled.
      */
-    private static boolean ENABLE_FALLBACK = !StringUtils.equalsIgnoreCase(
-            System.getProperty("support-core-plugin.SecretHandler.ENABLE_FALLBACK", "TRUE"), "FALSE");
+    private static boolean ENABLE_FALLBACK =
+            !"FALSE".equalsIgnoreCase(System.getProperty("support-core-plugin.SecretHandler.ENABLE_FALLBACK", "TRUE"));
```
jglick, line 53: BTW this is done more easily with `Boolean.getProperty` or `SystemProperties.getBoolean`.

src/main/java/com/cloudbees/jenkins/support/impl/DirectoryComponent.java
```diff
@@ -19,7 +19,7 @@
 import java.util.logging.Logger;
 import java.util.stream.Stream;
 import jenkins.model.Jenkins;
-import org.apache.commons.lang.ArrayUtils;
+import org.apache.commons.lang3.ArrayUtils;
```
jglick, line 22: This dep seems to be coming as a third-level transitive dep via `configuration-as-code`, which is `optional`, so this might break if that dep were disabled. Anyway it is pretty gratuitous.

jglick (changes requested):
https://github.com/jenkinsci/support-core-plugin/pull/659#discussion_r2172628968 possible regression

src/main/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilter.java
```diff
@@ -93,7 +92,7 @@ public synchronized void reload() {
                 });
 
         NameProvider.all().forEach(provider -> provider.names()
-                .filter(StringUtils::isNotBlank)
+                .filter(s -> s != null && !s.trim().isEmpty())
```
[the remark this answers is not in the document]
jglick: Strange, perhaps an artifact of a bad test?

src/main/java/com/cloudbees/jenkins/support/impl/DirectoryComponent.java
```diff
@@ -221,15 +222,12 @@ public void scan(File dir, FileVisitor visitor) throws IOException {
...
-                                .map(s -> dir.toPath().relativize(Paths.get(s)).toString())
-                                .toArray());
-                int var6 = var5.length;
-
-                for (int var7 = 0; var7 < var6; ++var7) {
-                    String f = var5[var7];
+                ArrayList<String> var5 = new ArrayList<>(Arrays.asList(dirScanner.getIncludedFiles()));
+                var5.addAll(Stream.of(dirScanner.getNotFollowedSymlinks())
+                        .map(s -> dir.toPath().relativize(Paths.get(s)).toString())
+                        .toList());
```
jglick, line 228: BTW `Stream.concat` could also be used.

jglick (approved):

mawinter69:
test failure seems unrelated. Worked fine locally on windows

PR #672, opened 2025-07-30
Upgrade Commons Lang from 2 to 3

basil:
Migrate from deprecated/EOL Commons Lang 2 to Commons Lang 3.


mawinter69:
More or less a duplicate of #659

jglick (approved):
Prefer #659 if it were fixed, but in the meantime this will do.

PR #681, opened 2025-08-28
Modernizing the UI and adding a HASH

ironcerocloudbees:
This is an attempt to enhance the Support Bundle Generation view. The improvements introduced are:

### Moving the list of options from the main page to a pop-up/widget.

Users don't always need to mark or unmark any options if they just want to generate a default support bundle. Moving the list to a pop-up or widget allows us to add more elements to manage the checked options in the main view (see next improvement sections).

To open the widget we will need to click on `Select options`:

SCR-20250828-mckr

The new widget shows the following:

SCR-20250828-mczm

Please pay attention to the code displayed in the bottom-right corner. This is a HASH generated from the current selected (checked) options. Please refer to the "Custom options" section for more information about this new element.

You can close the widget by clicking on `Close` or pressing the `Esc` key.

### Choose a general set of options

Selecting which options should be checked can be complex for standards users, so it may be a good idea to have a dropdown element with some standard options (such as configuration files, performance, etc.). For the moment, only three options were added (default, default + configuration files, and performance). As soon as the user clicks or selects one general option in this dropdown, the options will be checked in the widget (by JS).

SCR-20250828-mgck

### Custom options

This could be the most interesting part of this development. 

Each option will have an ID in the list of options. Anytime that a user check or uncheck any option in the widget a HASH value will be calculated in base on the current selected options. This HASH will appear in two locations: in the bottom-right corner of the widget, and in the input textbox under the `Choose a general set of options`. 

SCR-20250828-mhsg

Selecting the `Custom` value in the `Choose a general set of options` dropdown, the textbox for the custom value will appear. If you type a HASH value in this textbox and press the `Enter` key (or click in the `Apply` button), the options in the widget will be selected according to this HASH value. After that, we could optionally click on `Select options` and alter this set of select options.

There is an issue with plugins that extend the `Component` class. The elements shown in the widget (and originally on that page) are classes that extend the `Component` class and satisfy some requirements. Those classes can be provided by any plugin (not only the support-core-plugin). This question makes the management of this HASH so hard (since we cannot ensure that all the child components will implement the `getHash` method (inherited from the `Component` class). Therefore, the decision was made to disregard the child components that would not correctly implement the `getHash` method. By doing that, we can share a HASH code that we can ensure will work.

## Testing done

The development was tested in three different ways:

1. A Junit test was created for the `getHash` method of the child components.
2. The plugin was deployed in Jenkins version `2.479.3` and passed the basic integration tests:
    1. Selecting all the general options and confirming that the options were checked correctly in the widget.
    2. Selecting the `Custom` general options and confirming that the correct options were checked with different known HASH codes.
    3. Checking and unckecking different options in the widget and confirming that the generated HASH was correct.
    4. Confirming the usage of events (like the `Esc` for the widget or the `Enter` for the custom textbox).
    5. Generating a support bundle and confirming that the options checked in the widget were included in the file.  

## Submitter checklist
- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/api/Component.java
+++ b/src/main/java/com/cloudbees/jenkins/support/api/Component.java
@@ -148,6 +148,8 @@ public abstract class Component implements ExtensionPoint {
         return getClass().getSimpleName();
     }
 
+    public int getHash() { return -1; };
+
     @Deprecated
     public void start(@NonNull SupportContext context) {}
 
--- a/src/main/java/com/cloudbees/jenkins/support/api/Component.java
+++ b/src/main/java/com/cloudbees/jenkins/support/api/Component.java
@@ -148,6 +148,14 @@ public abstract class Component implements ExtensionPoint {
         return getClass().getSimpleName();
     }
 
+    /**
+     * Used for getting a hash code with the options selected by the user.
+     * Each component (class that extends the Component class) should have a unique value.
+     * Only child components were treated in this plugin. Other plugins that extends the Component class
+     * should implement (override) the getHash method with a unique value.
+     * The UI will ignore any component with a "-1" value in the id (extracted from this getHash method)
+     * in order to generate the final hash value for the selected options.
+     */
     public int getHash() { return -1; };
 
     @Deprecated
--- a/src/main/resources/com/cloudbees/jenkins/support/Messages.properties
+++ b/src/main/resources/com/cloudbees/jenkins/support/Messages.properties
@@ -42,17 +42,18 @@ SupportPlugin_DeleteBundle=Delete bundle
 SupportPlugin_DownloadBundle=Download bundle
 SupportPlugin_ListBundle=Existing support bundles
 SupportPlugin_loadConfig=Read support plugin configuration
-SupportPlugin_GenerateSupportBundle=Generate Support Bundle
+
+SupportPlugin_ApplyCustomChanges=Apply
 SupportPlugin_ChooseGeneralOption=Choose a general set of options
 SupportPlugin_ChooseGeneralOptionApplied=Success: options have been checked automatically based on the general option selected (above)
 SupportPlugin_ChooseGeneralOptionApplying=Applying options based on the general option selected (above) ...
 SupportPlugin_ChooseOptions=Select options
 SupportPlugin_Close=Close
+SupportPlugin_GenerateSupportBundle=Generate Support Bundle
 SupportPlugin_GenericOption_ConfigurationFiles=Default with configuration files
-SupportPlugin_GenericOption_PerformanceData=Performance data
-SupportPlugin_GenericOption_Default=Default
 SupportPlugin_GenericOption_Custom=Custom
-SupportPlugin_ApplyCustomChanges=Apply
+SupportPlugin_GenericOption_Default=Default
+SupportPlugin_GenericOption_PerformanceData=Performance data
 
 
 SupportPlugin_Category_Agent=Agents
--- a/src/main/resources/com/cloudbees/jenkins/support/Messages.properties
+++ b/src/main/resources/com/cloudbees/jenkins/support/Messages.properties
@@ -42,6 +42,16 @@ SupportPlugin_DeleteBundle=Delete bundle
 SupportPlugin_DownloadBundle=Download bundle
 SupportPlugin_ListBundle=Existing support bundles
 SupportPlugin_loadConfig=Read support plugin configuration
+SupportPlugin_GenerateSupportBundle=Generate Support Bundle
+SupportPlugin_ChooseGeneralOption=Choose a general set of options
+SupportPlugin_ChooseGeneralOptionApplied=Options have been checked automatically based on the general option selected (above)
+SupportPlugin_ChooseOptions=Select options
+SupportPlugin_Close=Close
+SupportPlugin_GenericOption_ConfigurationFiles=Default with configuration files
+SupportPlugin_GenericOption_PerformanceData=Performance data
+SupportPlugin_GenericOption_Default=Default
+SupportPlugin_GenericOption_Custom=Custom
+
 
 SupportPlugin_Category_Agent=Agents
 SupportPlugin_Category_Builds=Builds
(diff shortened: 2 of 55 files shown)


src/main/resources/com/cloudbees/jenkins/support/Messages.properties
```diff
@@ -43,6 +43,19 @@ SupportPlugin_DownloadBundle=Download bundle
 SupportPlugin_ListBundle=Existing support bundles
 SupportPlugin_loadConfig=Read support plugin configuration
 
+SupportPlugin_ApplyCustomChanges=Apply
+SupportPlugin_ChooseGeneralOption=Choose a general set of options
```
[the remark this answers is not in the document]
ironcerocloudbees: Done, I needed to rename/replace all the references to options (not only in the messages.properties). I needed to perform all the tests after the change (because the change affected almost all the PR files).

PierreBtz:
@ironcerocloudbees sorry for the delay. My suggestion to get going would be to split this development in two since you are trying to tackle two issues at once:

1. Introduction of profiles

The current architecture is problematic, the plugin shouldn't have knowledge of downstream components.

I tend to agree with Jesse's approach of storing custom profiles as a [UserProperty](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/model/UserProperty.java) (or if want to keep things on the front end, a cookie maybe?).

One thing that this approach doesn't allow is custom profile sharing (I don't know if it's a requirement of yours or not?). If it is, you could maybe provide an url fragment to share with the component selection appearing as query parameters. Note that to keep things simple, this can be delivered independently on a subsequent PR.

2. a UX redesign (depends on 1 but can be delivered after)

No strong opinion on this I'm no UX designer (and maybe we could ask for the opinion of one...). I personally like that the main view is not polluted by a long list of components anymore.

src/main/resources/com/cloudbees/jenkins/support/SupportAction/index.jelly
```diff
@@ -60,42 +60,31 @@
...
+              <option value="H4sIAAAAAAAAE6WV3XIaMQyFXyaXGb9DsqGlHUgpP7k3XrGo2JYryUDevuOlKe0VYnq13vE5862PZG2g5EKk2m8BxP2AfMAsTmopxOowlehW0R+ho5R87lfqFUUxyGO47XypqUzO7X3ttxEsllfqYXKGUJXYxGiGOWW06p8rxv57hWr6nGXNGfMwmuSRePiQBXQl1qEZTsSHXaSTC0VcV2S9Z/B9O/tTUKT8sMACETNcNyzopy1V/XrZMen7hBlF2SserZEEyjscdhhBXDeuP2FsxS6UIauFOyM/tsU9qG+6B77y5C7ggun8fjHXdljKN1yinrUWt9pX7emUrTSJdGL4WUHUrSKdlpe12f6BvTz/k3ptnvvS2pTeK3SQFdhWzmEAnvvsB5uhq6KUZjSYuvR3Q1vlY6NYxWsvB6v2c2dVjvfwmekktjxG/cYobrqOqq2UU/BlI36AKYrSwD6ZEn+bL5gCiKzeRSHNQRmDdJQVssrD3IuxNS7+f26eza2YgN0L+D5SONx5jUb0JB+RKSfI+uYZ29/EVLw2XV5AAmNR4hkmNOFeQdtU/9Iuzc4HG2sJR2CBcTzZ41wwFWBFG+OvKXAr81HZ11TcFId9m9J/Ev8FWTBcDfUHAAA=">${it.genericComponentPerformanceData}</option>
```
jglick, line 87: This is not maintainable. Define a Java method to construct these hashes by some means.

ironcerocloudbees: Yep, this is another excellent idea. I modified this part as well. Now, we can override the `getDefautlPreChooseOptions` method in the component to choose the predefinied options where the component will be included:

```
@Override
    public SupportAction.PreChooseOptions[] getDefautlPreChooseOptions() {
        return new SupportAction.PreChooseOptions[]{ SupportAction.PreChooseOptions.Default, SupportAction.PreChooseOptions.ConfigurationFiles, SupportAction.PreChooseOptions.PerformanceData };
    }
```

Dohbedoh (commented):
Thanks for the contribution and sorry for reviewing so late...
Overall it looks OK to me but there is still some work to it until we can merge. I believe the Component ID approach is best due to the plugin extensions. FWIW I think the UI / CLI could be adapted to support selecting a profile.

> Also I wonder if we need the dropdown and the popup, or if we could just have buttons to select all/none of the components generally, or within one of the categories. That would be make it easy to collect all components pertaining to agents, for example. Plus a button to reset to defaults.

I think this was general idea we had the various time we discussed about this. I was imagining a reactive grid component with profile selection and even single component content download. But my expertise in UI / fronted is limited. The feature in this PR is still a step forward towards this and that's great.

Couple of points I thin kneed to be addressed:

* When testing this PR, I was looking for the list of components that each of those options would pre-select. Which I could not get. I thought that the "Select components" would show me the list for the option I just picked but actually it doesn't. I wish I could see what components are included.

* The default selection is changed. There is a method `isSelectedByDefault()` that should be integrated in the "Default" options to preserve it. NOTE: the review of the defaut selection is a topic in itself. There's been a growing lists of components that are selected by default and that need to be reviewed. Maybe as part of a different PR.

* Some UI tests will need to be updated I think.

NOTE: I am also wondering if the options list should be made extensible. If other plugin could contribute a profile for example.

src/main/java/com/cloudbees/jenkins/support/api/Component.java
```diff
@@ -193,6 +176,14 @@ public boolean supersedes(Component component) {
         return false;
     }
 
+    public SupportAction.PreChooseOptions[] getDefautlPreChooseOptions(){
```
jglick, line 179: typo

src/main/java/com/cloudbees/jenkins/support/api/Component.java
```diff
@@ -193,6 +176,14 @@ public boolean supersedes(Component component) {
...
+        return new SupportAction.PreChooseOptions[0];
+    }
+
+    public Boolean checkDefaultPreChooseOptions(SupportAction.PreChooseOptions preChooseOptions) {
```
jglick, line 183: Why `Boolean` rather than `boolean`?

PR #961, opened 2026-05-12
Ensure thread dump file timestamp is in UTC

duemir:
For consistency with slow request report files and general UTC preference, ensure multiple thread dump collection functions set UTC as a TZ explicitly.

`SlowRequestChecker` explicitly sets the time zone of its `SimpleDateFormat` which is used for file name generation.
https://github.com/jenkinsci/support-core-plugin/blob/1801.v76b_389d2deec/src/main/java/com/cloudbees/jenkins/support/slowrequest/SlowRequestChecker.java#L61-L65

The `ThreadDumps` utility functions for generating a sequence of thread dumps were not doing that. Hence, file names it creates for thread dumps can be in the local time zone of the server. The PR suggests explicitly setting the time zone to UTC in `ThreadDumps`.

### Submitter checklist
- [ ] ~Link to relevant issues in GitHub or Jira~
- [ ] ~Link to relevant pull requests, esp. upstream and downstream changes~

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/impl/ThreadDumps.java
+++ b/src/main/java/com/cloudbees/jenkins/support/impl/ThreadDumps.java
@@ -29,6 +29,7 @@ import java.util.Comparator;
 import java.util.Date;
 import java.util.Objects;
 import java.util.Set;
+import java.util.TimeZone;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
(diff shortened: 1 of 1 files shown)


jglick:
https://github.com/jenkinsci/support-core-plugin/blob/76b389d2deecc805acf928f3578033da646605d9/src/main/java/com/cloudbees/jenkins/support/slowrequest/SlowRequestChecker.java#L61-L65 for the more legible permalink.

Is https://github.com/jenkinsci/support-core-plugin/pull/961/checks?check_run_id=75504908276 reproducible or a flake?

jglick (approved):

PR #980, opened 2026-08-17
Evict unused ContentMappings entries after a configurable retention period

### LINKED ISSUE #811: [JENKINS-55995] support-core-plugin: mappings removed not removed from disk
<p><b>Expected behavior</b>:</p>
<ul>
	<li>Whenever you change the objects created in Jenkins, the mappings persisted are refreshed</li>
</ul>

<p><b>Actual behavior</b>:</p>
<ul>
	<li>When you remove an object, it isn't removed from the persisted storage (xml file).</li>
</ul>

<p><b>Steps</b>:</p>
<ul>
	<li>Add labels to the master node. Example: <em>info severe threads jvm</em></li>
	<li>Check anonymize the bundles</li>
	<li>Generate a bundle</li>
	<li>Review the <em>com.cloudbees.jenkins.support.filter.ContentMappings.xml</em> file</li>
	<li>It has the words previously defined</li>
	<li>Go again to the master node and remove one label. Example: <em>info</em></li>
	<li>Generate a bundle</li>
	<li>Review the <em>com.cloudbees.jenkins.support.filter.ContentMappings.xml</em> file</li>
	<li>The info word <b>still stays</b> in the file.</li>
</ul>

<p>Tested in support-core-2.54</p>

---
<details><summary><i>Originally reported by  <a href="https://issues.jenkins.io/secure/ViewProfile.jspa?name=mramonleon">mramonleon</a>, imported from: <a class="original-jira-link" href="https://issues.jenkins.io/browse/JENKINS-55995" target="_blank">support-core-plugin: mappings removed not removed from disk</a></i></summary>
<i><ul>
<li><b>assignee</b>:  <a href="https://issues.jenkins.io/secure/ViewProfile.jspa?name=allan_burdajewicz">allan_burdajewicz</a>
<li><b>status</b>: Open
<li><b>priority</b>: Major
<li><b>component(s)</b>: support-core-plugin
<li><b>resolution</b>: Unresolved
<li><b>votes</b>: 0
<li><b>watchers</b>: 2
<li><b>imported</b>: 20251212-090250
</ul></i>
<details><summary>Raw content of original issue</summary>

<pre>
<p><b>Expected behavior</b>:</p>
<ul>
	<li>Whenever you change the objects created in Jenkins, the mappings persisted are refreshed</li>
</ul>

<p><b>Actual behavior</b>:</p>
<ul>
	<li>When you remove an object, it isn't removed from the persisted storage (xml file).</li>
</ul>

<p><b>Steps</b>:</p>
<ul>
	<li>Add labels to the master node. Example: <em>info severe threads jvm</em></li>
	<li>Check anonymize the bundles</li>
	<li>Generate a bundle</li>
	<li>Review the <em>com.cloudbees.jenkins.support.filter.ContentMappings.xml</em> file</li>
	<li>It has the words previously defined</li>
	<li>Go again to the master node and remove one label. Example: <em>info</em></li>
	<li>Generate a bundle</li>
	<li>Review the <em>com.cloudbees.jenkins.support.filter.ContentMappings.xml</em> file</li>
	<li>The info word <b>still stays</b> in the file.</li>
</ul>

<p>Tested in support-core-2.54</p></pre>
</details>
</details>

mikecirioli:
`ContentMappings` grows without bound - deleted items never leave. Adds a 90-day retention sweep (tunable via `ContentMappings.RETENTION` system property).

"Seen" = reported by a `NameProvider` during reload, or matched during filtering. A deleted item whose name still appears in content stays alive as long as it keeps matching.

- Sweep runs end of `writeBundle`, after filtering has touched what it matched. Skipped when anonymization is off.
- Concurrent bundles: `touchMatched` re-inserts on match, not just touch - a second bundle can evict the mapping mid-flight, and for deleted items nothing recreates it.
- `onMatch` consumer captured at reload time so filtering never needs a live Jenkins (shutdown-safe).
- Migrated data: no `lastSeen` in XML → stamped now, not epoch. Upgrading doesn't sweep the table.
- Snapshot atomicity: pattern + maps + consumer behind one `AtomicReference`, closes a torn-read NPE.

Fixes #811

### Related

- #984 — the OOM itself (per-entry regex emission cost)
- #986, #987, #988 — correctness/cleanup from the same investigation


src/main/java/com/cloudbees/jenkins/support/filter/ContentMapping.java
```diff
@@ -53,9 +51,15 @@ public class ContentMapping implements ContentFilter {
...
 
+    // Mutable usage-tracking metadata, deliberately outside equals/hashCode/identity: when this mapping was last
+    // known to be live (per a NameProvider) or actually matched during filtering. Drives grace-period eviction in
+    // ContentMappings#evictStale(); a mapping that stops being touched eventually gets swept.
+    private volatile long lastSeen;
+
     private ContentMapping(@NonNull String original, @NonNull String replacement) {
         this.original = original;
         this.replacement = replacement;
+        this.lastSeen = System.currentTimeMillis();
```
jglick: I would recommend using `java.time` types for legibility and safety, converting to/from `long` only if and when necessary for use with some older API.

src/main/java/com/cloudbees/jenkins/support/filter/ContentMapping.java
```diff
@@ -118,15 +143,24 @@ private Object writeReplace() {
         SerializationProxy proxy = new SerializationProxy();
         proxy.original = original;
         proxy.replacement = replacement;
+        proxy.lastSeen = lastSeen;
```
jglick: Since this is in the serial form, `long` is OK.

src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
```diff
@@ -147,6 +149,60 @@ public void reload() {
...
     }
 
+    // Read live (not cached) so that ops can tune via system property without a restart, and so tests can
+    // override per-test.
+    private static long retentionDays() {
+        return Long.getLong(ContentMappings.class.getName() + ".RETENTION_DAYS", 90);
+    }
+
+    private static int maxEntries() {
+        return Integer.getInteger(ContentMappings.class.getName() + ".MAX_ENTRIES", 50_000);
```
jglick: `SystemProperties`

src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
```diff
@@ -147,6 +149,60 @@ public void reload() {
         });
     }
 
+    // Read live (not cached) so that ops can tune via system property without a restart, and so tests can
+    // override per-test.
+    private static long retentionDays() {
+        return Long.getLong(ContentMappings.class.getName() + ".RETENTION_DAYS", 90);
```
jglick: Use `SystemProperties.getDuration`, and use `Duration` internally. That also means you can drop the _days_ nomenclature (just call it _retention_): the type will be self-documenting.

src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
```diff
@@ -147,6 +149,60 @@ public void reload() {
...
+    /**
+     * Evicts mappings that haven't been live or actually matched within the retention window, then enforces an
+     * absolute cap (oldest-touched first) as a backstop regardless of the retention window. Must run after the
+     * current reload cycle has had a chance to touch every currently-live and currently-matched mapping, or it
+     * will evict based on stale bookkeeping from the previous cycle.
+     */
+    void evictStale() {
+        long cutoff = System.currentTimeMillis() - TimeUnit.DAYS.toMillis(retentionDays());
+        int removedByAge = 0;
+        Iterator<Map.Entry<String, ContentMapping>> it = mappings.entrySet().iterator();
```
jglick:
```suggestion
        var it = mappings.entrySet().iterator();
```

src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
```diff
@@ -147,6 +149,60 @@ public void reload() {
...
+        while (it.hasNext()) {
+            Map.Entry<String, ContentMapping> entry = it.next();
```
jglick:
```suggestion
            var entry = it.next();
```

src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
```diff
@@ -147,6 +149,60 @@ public void reload() {
...
+            if (entry.getValue().getLastSeen() < cutoff) {
+                it.remove();
+                removedByAge++;
+            }
+        }
+        if (removedByAge > 0) {
+            LOGGER.log(
+                    Level.FINE,
+                    "Evicted " + removedByAge + " stale content mappings (unused for " + retentionDays() + "+ days)");
```
jglick:
```suggestion
            LOGGER.fine(() -> "Evicted " + removedByAge + " stale content mappings (unused for " + retentionDays() + "+ days)");
```

src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
```diff
@@ -147,6 +149,60 @@ public void reload() {
...
+            mappings.values().stream()
+                    .map(m -> Map.entry(m.getOriginal(), m.getLastSeen()))
+                    .sorted(Map.Entry.comparingByValue())
+                    .limit(overflow)
+                    .map(Map.Entry::getKey)
+                    .forEach(mappings::remove);
+            LOGGER.log(
+                    Level.INFO,
+                    "ContentMappings exceeded " + maxEntries + " entries; evicted " + overflow
+                            + " oldest-touched mappings");
```
jglick:
```suggestion
            LOGGER.info(() -> "ContentMappings exceeded " + maxEntries + " entries; evicted " + overflow
                            + " oldest-touched mappings");
```

src/main/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilter.java
```diff
@@ -54,20 +54,40 @@ public class SensitiveContentFilter implements ContentFilter {
...
     }
 
     @Override
     public @NonNull String filter(@NonNull String input) {
-        return WordReplacer.replaceWords(input, mappingsPattern.get(), replacementsMap.get());
+        // Snapshot both maps once per call, alongside the pattern, so a concurrent reload() can't mix an old
+        // pattern with newer maps (or vice versa) partway through matching -- see replacementFor().
+        Map<String, String> replacements = replacementsMap.get();
+        Map<String, ContentMapping> matched = matchedMappings.get();
+        return WordReplacer.replaceWords(
```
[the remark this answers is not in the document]
jglick: Sounds like a legitimate concern. TBH I cannot follow all of this code so easily.

src/test/java/com/cloudbees/jenkins/support/filter/ContentMappingsTest.java
```diff
@@ -128,21 +128,137 @@ void contentMappingsRemovedSerialized(JenkinsRule r) throws Throwable {
...
+    @Test
+    void evictStaleRemovesMappingsPastRetentionWindow(JenkinsRule r) {
+        ContentMappings mappings = ContentMappings.get();
+        long now = System.currentTimeMillis();
+
+        ContentMapping stale = mappings.getMappingOrCreate("stale-entry", ContentMappingsTest::identityMapping);
+        stale.touch(now - TimeUnit.DAYS.toMillis(91));
+
+        ContentMapping fresh = mappings.getMappingOrCreate("fresh-entry", ContentMappingsTest::identityMapping);
+        fresh.touch(now - TimeUnit.DAYS.toMillis(1));
```
jglick: `java.time`

src/test/java/com/cloudbees/jenkins/support/filter/ContentMappingsTest.java
```diff
@@ -128,21 +128,137 @@ void contentMappingsRemovedSerialized(JenkinsRule r) throws Throwable {
...
+        System.setProperty(property, "10");
+        try {
+            ContentMappings mappings = ContentMappings.get();
+            long now = System.currentTimeMillis();
+            for (int i = 0; i < 15; i++) {
+                ContentMapping mapping =
+                        mappings.getMappingOrCreate("entry-" + i, ContentMappingsTest::identityMapping);
+                // Space out lastSeen (well within the retention window) so eviction order is deterministic:
+                // entry-0 is the oldest-touched, entry-14 the most recently touched.
+                mapping.touch(now - TimeUnit.MINUTES.toMillis(15 - i));
```
jglick: ditto

src/test/java/com/cloudbees/jenkins/support/filter/ContentMappingsTest.java
```diff
@@ -128,21 +128,137 @@ void contentMappingsRemovedSerialized(JenkinsRule r) throws Throwable {
...
+
+        Map<String, String> remaining = mappings.getMappings();
+        assertFalse(remaining.containsKey("stale-entry"), "mapping unused for over 90 days should be evicted");
+        assertTrue(remaining.containsKey("fresh-entry"), "mapping used within the retention window should survive");
+    }
+
+    @Test
+    void evictStaleEnforcesMaxEntriesCapOldestTouchedFirst(JenkinsRule r) {
+        String property = ContentMappings.class.getName() + ".MAX_ENTRIES";
+        String previous = System.getProperty(property);
```
jglick: `FlagRule`

src/test/java/com/cloudbees/jenkins/support/filter/ContentMappingsTest.java
```diff
@@ -128,21 +128,137 @@ void contentMappingsRemovedSerialized(JenkinsRule r) throws Throwable {
...
+        assertTrue(
+                migrated.getLastSeen() >= before && migrated.getLastSeen() <= after,
+                "migrated mapping should default lastSeen to \"now\", not the epoch (0)");
+    }
+
+    /**
+     * Stand-in for the pre-upgrade {@code ContentMapping.SerializationProxy} shape, which had no
+     * {@code lastSeen} field. Used only to empirically verify the downgrade-safety claim in the spec: that
+     * XStream (as used by {@code Persistence}/{@code XmlFile}) tolerates an unrecognized XML element on read
+     * rather than failing to unmarshal.
```
jglick: This seems unnecessary. We know well how XStream behaves. Anyway, what “downgrade safety claim”? We never claim that a `$JENKINS_HOME` installation will be usable after _downgrading_ anything, only upgrading. If you need to downgrade something, restore from backup.

jglick:
> Testing done

Since this is clearly all AI-written (and mostly AI-reviewed), it would be nice to hear about the problem being reproduced in an actual controller and that the proposed change does fix it (when system properties are tuned to make cleanup more aggressive than usual).

mikecirioli:
I broke the original  apart into smaller independent pieces and was able to make some additional improvements as well.  I'll cross-ref all the PR's once i finish creating them

src/main/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilter.java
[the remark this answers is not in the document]
mikecirioli: Fixed - the pattern and both maps are now one immutable Snapshot record behind a single AtomicReference, so there's no torn read. It starts as a never-matching empty snapshot too, so filter() before the first reload() is a no-op instead of an NPE.

src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
```diff
@@ -147,6 +151,35 @@ public void reload() {
         });
     }
 
+    // Read per-call rather than cached in a static field, so the value is picked up wherever it is supplied and
+    // so tests can override it. Changing it on a running controller still requires a restart.
```
jglick: Unlikely to ever matter, but—why would a restart be required?

src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
```diff
@@ -147,6 +151,35 @@ public void reload() {
...
+            var entry = it.next();
+            if (entry.getValue().getLastSeen().isBefore(cutoff)) {
+                it.remove();
+                removedByAge++;
+            }
+        }
+        if (removedByAge > 0) {
+            int removedCount = removedByAge;
+            LOGGER.fine(() -> "Evicted " + removedCount + " stale content mappings (unused for "
+                    + retention().toDays() + "+ days)");
```
jglick: JaCoCo notes this is not covered by tests. I generally recommend using `LoggerRule` (4) / `LogRecorder` (5) to show fine logs during tests of some component.

src/test/java/com/cloudbees/jenkins/support/filter/ContentMappingsTest.java
```diff
@@ -128,21 +129,72 @@ void contentMappingsRemovedSerialized(JenkinsRule r) throws Throwable {
...
-        // Mapping removed after restart
         mappings = ContentMappings.get();
+        assertThat(
+                "The mapping of a removed project should survive its grace period",
+                mappings.getMappings(),
+                hasEntry("ShortName", jobReplacement.toString()));
 
-        // Check if the mapping exists after restart
+        // Simulate the grace period elapsing by backdating lastSeen directly (touch(Instant) is
+        // package-private, reachable from this same-package test), then run the cycle again.
```
jglick:
```suggestion
        // Simulate the grace period elapsing by backdating lastSeen directly, then run the cycle again.
```
Unnecessary to comment on what can be seen clearly from an IDE.

src/test/java/com/cloudbees/jenkins/support/filter/ContentMappingsTest.java
```diff
@@ -128,21 +129,72 @@ void contentMappingsRemovedSerialized(JenkinsRule r) throws Throwable {
...
+        StreamSupport.stream(mappings.spliterator(), false)
+                .filter(mapping -> mapping.getOriginal().equals("ShortName"))
+                .findFirst()
+                .orElseThrow()
+                .touch(Instant.now().minus(Duration.ofDays(91)));
```
jglick: OK, though it is more convincing to set the system property to make the grace period very short—say, 15s—then `await()` for the expected eviction.

src/test/java/com/cloudbees/jenkins/support/filter/ContentMappingsTest.java
```diff
@@ -128,21 +129,72 @@ void contentMappingsRemovedSerialized(JenkinsRule r) throws Throwable {
...
+        assertTrue(remaining.containsKey("fresh-entry"), "mapping used within the retention window should survive");
+    }
+
+    @Test
+    void migratedMappingDefaultsLastSeenToNowNotEpoch(JenkinsRule r) {
+        String xml = Jenkins.XSTREAM2.toXML(ContentMapping.of("legacy-original", "legacy-replacement"));
+        assertTrue(xml.contains("lastSeen"), "sanity check: current wire format should include lastSeen");
+
+        // Simulate XML persisted before the lastSeen field existed.
+        String legacyXml = xml.replaceFirst("<lastSeen>[^<]*</lastSeen>", "");
```
jglick: This would better be done with `@OldData`.

jglick (approved):

src/main/java/com/cloudbees/jenkins/support/filter/ContentFilter.java
```diff
@@ -131,10 +131,12 @@ static void reloadAndSaveMappings(@NonNull ContentFilter filter) throws IOExcept
         if (BulkChange.contains(mappings)) {
             mappings.reload();
             filter.reload();
+            mappings.evictStale();
```
RanjitCS: have you tested this in a controller for ip_mappings? `evictStale()` runs before content is filtered but `reload()` skips `ip_ mappings` and `InetAddressContentFilter` only touches them at filter time this means old-but-live IPs are evicted then recreated with a new name. i think we should run `evictStale()` fter the bundles content has been filtered

mikecirioli: Agreed, and that is what it does now — `evictStale()` moved to the end of `writeBundle`, after all content is filtered. IP mappings are touched at filter time through `getMappingOrCreate`, so a live IP is refreshed before the sweep.

src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
```diff
@@ -130,6 +133,7 @@ private ContentMappings(@NonNull XmlProxy proxy) {
             @NonNull String original, @NonNull Function<String, ContentMapping> generator) {
         boolean isNew = !mappings.containsKey(original);
         ContentMapping mapping = mappings.computeIfAbsent(original, generator);
+        mapping.touch();
```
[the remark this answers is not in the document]
jglick: @mikecirioli is this real?

src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
```diff
@@ -140,13 +144,69 @@ private ContentMappings(@NonNull XmlProxy proxy) {
...
+     * has been filtered, so every currently-live and currently-matched mapping has been refreshed via {@code touch()}.
+     *
+     * <p>Uses {@code computeIfPresent} to express the remove-if-stale decision as one map operation. That is not
+     * atomic with respect to {@code touch()}, which mutates {@code lastSeen} in place rather than replacing the map
+     * value, so a touch landing between the staleness check and the removal is lost. {@link #touchMatched} covers
+     * the case where that matters: a mapping matched in real content is put back if a concurrently generated bundle
+     * has already swept it away. Do not rely on the mapping simply being recreated later -- for an item that has
+     * been deleted, nothing recreates it.
+     *
+     * <p>Must be called from within a {@link BulkChange} so the eviction is persisted.
```
jglick:
```suggestion
     * <p>Must be called from within a {@link hudson.BulkChange} so the eviction is persisted.
```
(or `import` it)

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
```diff
@@ -561,6 +562,11 @@ public static void writeBundle(
                     }
                 }
                 binaryOut.flush();
+                // Only evict when anonymization actually ran: with ContentFilter.NONE nothing calls touch(), so
+                // every mapping would look stale and the whole table would be dropped.
```
jglick: Not sure I understand. Is this real?

src/test/java/com/cloudbees/jenkins/support/filter/ContentMappingsTest.java
```diff
@@ -128,21 +133,100 @@ void contentMappingsRemovedSerialized(JenkinsRule r) throws Throwable {
...
         mappings = ContentMappings.get();
-
-        // Check if the mapping exists after restart
         assertThat(
-                "The mapping of a removed project shouldn't persist",
+                "The mapping of a removed project should survive its grace period",
                 mappings.getMappings(),
+                hasEntry("ShortName", jobReplacement.toString()));
+
+        System.setProperty(ContentMappings.class.getName() + ".RETENTION", "PT15S");
```
jglick, line 147: This format does not work AFAIK. Did you mean
```suggestion
        System.setProperty(ContentMappings.class.getName() + ".RETENTION", "15s");
```
?

src/test/java/com/cloudbees/jenkins/support/filter/ContentMappingsTest.java
```diff
@@ -214,4 +298,102 @@ void clear(JenkinsRule r) {
...
+
+        mappings.getMappingOrCreate(testName, ContentMappingsTest::identityMapping)
+                .touch(Instant.now().minus(Duration.ofDays(91)));
+
+        ContentFilters.get().setEnabled(false);
+        SupportPlugin.writeBundle(OutputStream.nullOutputStream(), List.of());
+
+        assertTrue(
+                ContentMappings.get().getMappings().containsKey(testName),
+                "with anonymization disabled nothing calls touch(), so eviction must not run");
```
jglick, line 411: But do you have any tests which verify that `writeBundle` _does_ call `evictStale` when there _is_ anonymization?

src/test/java/com/cloudbees/jenkins/support/filter/ContentMappingsTest.java
```diff
@@ -128,21 +133,100 @@ void contentMappingsRemovedSerialized(JenkinsRule r) throws Throwable {
...
+        assertTrue(remaining.containsKey("fresh-entry"), "mapping used within the retention window should survive");
+    }
+
+    @Test
+    void evictionLogReportsSubDayRetentionWindows(JenkinsRule r) {
+        ContentMappings mappings = ContentMappings.get();
+        mappings.getMappingOrCreate("stale-entry", ContentMappingsTest::identityMapping)
+                .touch(Instant.now().minus(Duration.ofDays(91)));
+
+        System.setProperty(ContentMappings.class.getName() + ".RETENTION", "PT15S");
```
jglick, line 201: ditto

src/test/java/com/cloudbees/jenkins/support/filter/ContentMappingsTest.java
```diff
@@ -128,21 +133,100 @@ void contentMappingsRemovedSerialized(JenkinsRule r) throws Throwable {
...
+        try (LogRecorder logger =
+                new LogRecorder().record(ContentMappings.class, Level.FINE).capture(1)) {
+            mappings.evictStale();
+
+            assertTrue(
+                    logger.getMessages().stream().anyMatch(msg -> msg.contains("retention window: PT15S")),
+                    "a sub-day retention window must be logged in full, not truncated to whole days");
+        } finally {
+            System.clearProperty(ContentMappings.class.getName() + ".RETENTION");
+        }
```
jglick, line 211: or `FlagExtension`

jglick (changes requested):
`SystemProperties.getDuration` does not honor `Duration.parse` format AFAIK.

As to the rest, it sounds OK but I am not familiar enough with the original to understand all the details.

src/test/java/com/cloudbees/jenkins/support/filter/ContentMappingsTest.java
mikecirioli: `SystemProperties.getDuration` uses `DurationStyle.detectAndParse` which handles both `SIMPLE` and `ISO8601`

src/test/java/com/cloudbees/jenkins/support/filter/ContentMappingsTest.java
mikecirioli: see previous reply (both formats are supported)

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
mikecirioli:
```suggestion
```

mikecirioli: is that clearer @jglick ?

src/test/java/com/cloudbees/jenkins/support/filter/ContentMappingsTest.java
mikecirioli: Added: `writeBundleEvictsStaleWhenAnonymizationEnabled`.

src/test/java/com/cloudbees/jenkins/support/filter/ContentMappingsTest.java
mikecirioli: Can’t scope it mid-method — the property is set after setup runs under the default window. Same for `removedItemMappingSurvivesGracePeriodThenEvicted`.

src/main/java/com/cloudbees/jenkins/support/SupportPlugin.java
jglick: Not really. Maybe just delete the comment altogether? IIUC the thing to evict would be an anonymization table. So if we are not anonymizing, obviously we do not need to do anything. There might be some edge cases when anonymization is turned off or on, but this comment does not seem to have been referring to that to begin with. The comment just seems like some weird internal Claude musings that are not helpful to human readers.

src/test/java/com/cloudbees/jenkins/support/filter/ContentMappingsTest.java
jglick: TIL. (When we document these, we always use the simple format.)

src/test/java/com/cloudbees/jenkins/support/filter/ContentMappingsTest.java
jglick: Which is why there is a mode which does not set the value initially, only resets it at test completion.

Not important, just FYI.

jglick (approved):
OK AFAICT (just review the warning from Copilot)

jglick:
(also https://github.com/jenkinsci/support-core-plugin/pull/980#issuecomment-5318056145, still unclear whether this & linked PRs have been tested in a real controller)

src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java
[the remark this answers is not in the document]
mikecirioli: No, AFAICT. `getMappingOrCreate` only runs during `reload()` or during content filtering,  both before `evictStale()` in the same bundle. A concurrent bundle's sweep can remove it, but the item is still live (a `NameProvider` reports it), so the next `reload()` recreates it immediately.

Line 157 (`touchMatched`) is the fix for the case where it does matter: a matched mapping from a snapshot where nothing will recreate it (deleted item). I think `Copilot` flagged it without reading the `putIfAbsent` on the next line.

mikecirioli:
running a quick validation on a live controller while this builds

mikecirioli:
All four tests pass on a real controller. Jenkins 2.541.3, plugin built from evict-final @ e5a1527a, 13 bundle generations, $3.36 on Sonnet.
```
┌────────────────────────┬────────┬──────────────────────────────────────────────────────────────────────────────────┐
│          Test          │ Result │                                     Evidence                                     │
├────────────────────────┼────────┼──────────────────────────────────────────────────────────────────────────────────┤
│ Eviction lifecycle     │ PASS   │ project-alpha survived under 90d, evicted under PT1S; live items kept            │
├────────────────────────┼────────┼──────────────────────────────────────────────────────────────────────────────────┤
│ Anonymization disabled │ PASS   │ project-beta mapping survived with lastSeen unchanged                            │
├────────────────────────┼────────┼──────────────────────────────────────────────────────────────────────────────────┤
│ Stale match survives   │ PASS   │ log line redacted to item_flawed_detail, mapping re-inserted with fresh lastSeen │
├────────────────────────┼────────┼──────────────────────────────────────────────────────────────────────────────────┤
│ Pseudonym stability    │ PASS   │ project-gamma → item_visual_dream unchanged across 3 more bundles                │
└────────────────────────┴────────┴──────────────────────────────────────────────────────────────────────────────────┘
```

PR #982, opened 2026-08-18
Fold case consistently when building and looking up anonymization keys

### LINKED ISSUE #981: Case-folding mismatch in SensitiveContentFilter: NPE during bundle generation, and some names silently never anonymized
## Summary

`SensitiveContentFilter` builds its lookup keys with one case-folding rule and matches with another. The compiled `Pattern` uses `CASE_INSENSITIVE | UNICODE_CASE`, which compares characters via `toLowerCase(toUpperCase(c))`, while the trie words and `replacementsMap` keys are built with `String.toLowerCase(Locale.ENGLISH)`:

```java
// reload()
String lowerCaseOriginal = name.toLowerCase(Locale.ENGLISH);
trie.add(lowerCaseOriginal);
replacementsMap.putIfAbsent(lowerCaseOriginal, ...);
...
Pattern.compile("(?<!\\w)" + trie.getRegex() + "(?!\\w)",
        Pattern.CASE_INSENSITIVE | Pattern.UNICODE_CASE);

// lookup, via WordReplacer.replaceWords(String, Pattern, Map)
replacements.get(s.toLowerCase(Locale.ENGLISH));
```

Those two rules do not always agree, which produces two distinct failures.

### 1. `NullPointerException` aborts bundle generation

When the pattern matches text that `String.toLowerCase` maps differently, the map lookup returns `null` and `Matcher.replaceAll(Function)` throws:

```
java.lang.NullPointerException: Cannot invoke "String.length()" because "replacement" is null
	at java.base/java.util.regex.Matcher.appendExpandedReplacement(Matcher.java:1058)
	at java.base/java.util.regex.Matcher.appendReplacement(Matcher.java:1041)
	at java.base/java.util.regex.Matcher.replaceAll(Matcher.java:1305)
	at com.cloudbees.jenkins.support.util.WordReplacer.replaceWords(WordReplacer.java:32)
	at com.cloudbees.jenkins.support.util.WordReplacer.replaceWords(WordReplacer.java:20)
	at com.cloudbees.jenkins.support.filter.SensitiveContentFilter.filter(SensitiveContentFilter.java:64)
```

Trigger: an object named `sap` produces the pattern literal `sap`, which **matches** `ſap` (U+017F LATIN SMALL LETTER LONG S) under Unicode case folding — but the lookup key is `"ſap".toLowerCase(ENGLISH)`, which leaves U+017F unchanged, so there is no map entry.

### 2. Some names are silently never anonymized (arguably worse)

If an object's **own name** contains a character whose lowercase mapping changes the string's *length*, the generated pattern cannot match anything at all — including that object's own name. The mapping is created, persisted, and listed in the management view as anonymized, but the real name is emitted in plaintext in every bundle with no error logged.

For an object named `İdalium` (U+0130 LATIN CAPITAL LETTER I WITH DOT ABOVE):

```
"İdalium"                       = U+0130 d a l i u m           (7 code points)
  → String.toLowerCase(ENGLISH) = U+0069 U+0307 d a l i u m    (8 code points)
                                         ^^^^^^ COMBINING DOT ABOVE, injected
```

That 8-code-point string becomes the regex literal, so the pattern searches for `i` + `U+0307` + `dalium`. Every realistic input fails at the second position:

| input | pos 1 | pos 2 | result |
|---|---|---|---|
| `İdalium` | `İ` vs `i` — folds, ok | `d` vs U+0307 | no match |
| `idalium` | `i` vs `i` — ok | `d` vs U+0307 | no match |
| `IDALIUM` | `I` vs `i` — ok | `D` vs U+0307 | no match |

Root cause in one line: `String.toLowerCase` may change a string's length, `Character.toLowerCase` may not — and the regex literal is built from the length-changing one.

## Reproducer

No Jenkins needed to see the divergence — this is plain JDK behaviour (verified on 21.0.11):

```java
import java.util.Locale;
import java.util.regex.*;

public class Demo {
    public static void main(String[] args) {
        // exactly how SensitiveContentFilter builds the pattern and the lookup key
        Pattern p = Pattern.compile("(?<!\\w)sap(?!\\w)",
                Pattern.CASE_INSENSITIVE | Pattern.UNICODE_CASE);
        Matcher m = p.matcher("start ſap end");   // U+017F long s
        System.out.println("matched: " + m.find());     // true
        System.out.println("key:     " + m.group().toLowerCase(Locale.ENGLISH));
        //  -> "ſap", which is NOT the map key "sap"  =>  map.get() == null  =>  NPE

        // failure mode 2: the literal itself is corrupted by a length-changing lowercase
        String literal = "İdalium".toLowerCase(Locale.ENGLISH);  // U+0130 -> i + U+0307
        Pattern q = Pattern.compile("(?<!\\w)" + literal + "(?!\\w)",
                Pattern.CASE_INSENSITIVE | Pattern.UNICODE_CASE);
        for (String in : new String[] {"İdalium", "idalium", "IDALIUM"}) {
            System.out.println(in + " -> " + q.matcher("x " + in + " y").find());  // all false
        }
    }
}
```

In the plugin, failure 1 surfaces as an aborted support bundle whenever such a character appears in filtered content; failure 2 surfaces as a name that is listed as anonymized but never actually redacted.

## Frequency

Rare. Only a handful of characters have lowercase mappings that change a string's length (U+0130 is essentially the canonical one), and the `ſ`-style divergence needs an unusual character to appear in filtered content. Reporting it because the feature accepts arbitrary Unicode names and claims to redact them, so a name it cannot match is a correctness gap regardless of frequency — and failure 2 fails *open*, which is the wrong direction for an anonymization feature.

## Environment

- support-core 1744.vf4f8e8f2b_0b_5 (the affected code path is long-standing and unchanged on `master`)
- Jenkins 2.541.3, JDK 21
- Anonymization enabled (`ContentFilters.enabled = true`)

## Fix

Deriving both the key and the lookup with the same per-code-point fold — `Character.toLowerCase(Character.toUpperCase(cp))`, which mirrors what the matcher actually does — resolves both failures and makes the `null` return impossible by construction. Verified that Java's `CASE_INSENSITIVE | UNICODE_CASE` single-character matching is exactly equality under that fold (4,000,000 random code-point pairs, 0 mismatches) and that the fold is idempotent across all 1,112,064 valid code points, so folding the pattern literal only ever broadens matching and never loses a match the current code catches.

Happy to open a PR with the fix and a regression test that fails on `master`.

mikecirioli:
`SensitiveContentFilter` compiles its pattern with `CASE_INSENSITIVE | UNICODE_CASE`, which folds via `toLowerCase(toUpperCase(c))`, but derives the trie words and `replacementsMap` keys with `String.toLowerCase(Locale.ENGLISH)`. The two disagree for some characters, with two results:

- The pattern matches text the map has no key for, so the replacement is `null` and `Matcher.replaceAll` throws. Literal `sap` matches `ſap` (U+017F), whose `toLowerCase` is unchanged.
- Where a name itself contains a character whose lowercase changes length, the literal is corrupted and matches nothing at all — so that name is never redacted. `İdalium` (U+0130) becomes literal `i` + U+0307 + `dalium`.

Both are fixed by deriving the key and the lookup with the same per-code-point fold.

Found while working on #811 (JENKINS-55995).

Fixes #981

### Related

Four independent PRs from the same pass over the anonymization code, plus the original report in #811:

- #980
- #982 *(this one)*
- #984
- #986

Suggested merge order: **#984 -> #986 -> #980 -> #982**. Any order works; this one just minimises rebasing.

### Testing done

New test in `SensitiveContentFilterTest` covering both characters plus ASCII guards; it fails on `master` with the NPE above.

### Submitter checklist
- [x] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilter.java
+++ b/src/main/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilter.java
@@ -61,7 +61,26 @@ public class SensitiveContentFilter implements ContentFilter {
 
     @Override
     public @NonNull String filter(@NonNull String input) {
-        return WordReplacer.replaceWords(input, mappingsPattern.get(), replacementsMap.get());
+        Map<String, String> replacements = replacementsMap.get();
+        return WordReplacer.replaceWords(input, mappingsPattern.get(), match -> replacements.get(fold(match)));
+    }
+
+    /**
+     * Canonicalizes a value for use as a key in the replacements map, folding case the same way the compiled
+     * {@link Pattern} does.
+     *
+     * <p>The pattern is compiled with {@link Pattern#CASE_INSENSITIVE} and {@link Pattern#UNICODE_CASE}, which
+     * compares characters via {@code toLowerCase(toUpperCase(c))}. {@link String#toLowerCase(Locale)} does not
+     * always agree with that, so deriving the key one way and the lookup the other lets the pattern match text
+     * that the map has no entry for. Two concrete divergences: {@code U+017F LATIN SMALL LETTER LONG S} matches
+     * the literal {@code s} but {@code toLowerCase} leaves it unchanged, and {@code U+0130 LATIN CAPITAL LETTER I
+     * WITH DOT ABOVE} matches the literal {@code i} but {@code toLowerCase} decomposes it to {@code i U+0307}.
+     * Deriving both sides with this fold keeps them in agreement.
+     */
+    static String fold(String value) {
+        StringBuilder folded = new StringBuilder(value.length());
+        value.codePoints().forEach(cp -> folded.appendCodePoint(Character.toLowerCase(Character.toUpperCase(cp))));
+        return folded.toString();
     }
 
     @Override
@@ -79,35 +98,38 @@ public class SensitiveContentFilter implements ContentFilter {
                 // Filter out IP mappings
                 .filter(mapping -> !mapping.getReplacement().startsWith("ip_"))
                 .forEach(contentMapping -> {
-                    String lowerCaseOriginal = contentMapping.getOriginal().toLowerCase(Locale.ENGLISH);
-                    if (!stopWords.contains(lowerCaseOriginal)) {
+                    String original = contentMapping.getOriginal();
+                    // Stop words are stored lower cased, so they are matched that way; the trie and the map are
+                    // keyed on the case fold instead, so that filter() can look up whatever the pattern matched.
+                    if (!stopWords.contains(original.toLowerCase(Locale.ENGLISH))) {
+                        String key = fold(original);
                         replacementsMap.put(
-                                lowerCaseOriginal,
+                                key,
                                 contentMapping
                                         .getReplacement()
                                         .replaceAll("\\\\", "\\\\\\\\")
                                         .replaceAll("\\$", "\\\\\\$"));
-                        trie.add(lowerCaseOriginal);
+                        trie.add(key);
                     }
                 });
 
         NameProvider.all()
                 .forEach(provider -> provider.names().filter(s -> !s.isBlank()).forEach(name -> {
-                    String lowerCaseOriginal = name.toLowerCase(Locale.ENGLISH);
                     // NOTE: We could well create a WordTrie for the stop words and use it as a filter instead of the
                     // conditional here. Or find a better way to deal with insensitive key mapping in general.
                     // But the reload is already quite fast anyway. (~1s for 10^4 items with 1 CPU / 2 GB memory
                     // container)
-                    if (!stopWords.contains(lowerCaseOriginal)) {
+                    if (!stopWords.contains(name.toLowerCase(Locale.ENGLISH))) {
                         ContentMapping mapping = mappings.getMappingOrCreate(
                                 name, original -> ContentMapping.of(original, provider.generateFake()));
+                        String key = fold(name);
                         // Matcher#appendReplacement needs to have the `\` and `$` escaped.
                         replacementsMap.putIfAbsent(
-                                lowerCaseOriginal,
+                                key,
                                 mapping.getReplacement()
                                         .replaceAll("\\\\", "\\\\\\\\")
                                         .replaceAll("\\$", "\\\\\\$"));
-                        trie.add(lowerCaseOriginal);
+                        trie.add(key);
                     }
                 }));
         this.mappingsPattern.set(Pattern.compile(
--- a/src/test/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilterTest.java
+++ b/src/test/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilterTest.java
@@ -108,4 +108,26 @@ class SensitiveContentFilterTest {
         assertThat(filter.filter(os)).isEqualTo(os);
         assertThat(filter.filter(label)).startsWith("label_").isNotEqualTo(label);
     }
+
+    @Test
+    void anonymizeUnicodeCaseVariants(JenkinsRule j) {
+        ContentMappings mappings = ContentMappings.get();
+        mappings.getMappingOrCreate("sapphirium", original -> ContentMapping.of(original, "item_test_alpha"));
+        mappings.getMappingOrCreate("idalium", original -> ContentMapping.of(original, "item_test_beta"));
+
+        SensitiveContentFilter filter = SensitiveContentFilter.get();
+        filter.reload();
+
+        // U+017F LATIN SMALL LETTER LONG S matches the literal 's' under CASE_INSENSITIVE|UNICODE_CASE, but
+        // String#toLowerCase leaves it untouched -- so keying the map that way loses the entry.
+        assertThat(filter.filter("build ſapphirium done")).isEqualTo("build item_test_alpha done");
+
+        // U+0130 LATIN CAPITAL LETTER I WITH DOT ABOVE matches the literal 'i', but String#toLowerCase
+        // decomposes it to 'i' + U+0307 COMBINING DOT ABOVE.
+        assertThat(filter.filter("job İdalium ran")).isEqualTo("job item_test_beta ran");
+
+        // plain ASCII case-insensitivity must keep working
+        assertThat(filter.filter("SAPPHIRIUM")).isEqualTo("item_test_alpha");
+        assertThat(filter.filter("sapphirium")).isEqualTo("item_test_alpha");
+    }
 }


src/main/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilter.java
```diff
@@ -79,35 +98,38 @@ public synchronized void reload() {
...
 
         NameProvider.all()
                 .forEach(provider -> provider.names().filter(s -> !s.isBlank()).forEach(name -> {
-                    String lowerCaseOriginal = name.toLowerCase(Locale.ENGLISH);
                     // NOTE: We could well create a WordTrie for the stop words and use it as a filter instead of the
                     // conditional here. Or find a better way to deal with insensitive key mapping in general.
                     // But the reload is already quite fast anyway. (~1s for 10^4 items with 1 CPU / 2 GB memory
                     // container)
-                    if (!stopWords.contains(lowerCaseOriginal)) {
+                    if (!stopWords.contains(name.toLowerCase(Locale.ENGLISH))) {
```
jglick, line 122: I have a stylistic preference for
```suggestion
                    if (!stopWords.contains(name.toLowerCase(Locale.ROOT))) {
```
but I do not _think_ it makes any practical difference.

src/main/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilter.java
```diff
@@ -61,7 +61,26 @@ public static SensitiveContentFilter get() {
...
+    /**
+     * Canonicalizes a value for use as a key in the replacements map, folding case the same way the compiled
+     * {@link Pattern} does.
+     *
+     * <p>The pattern is compiled with {@link Pattern#CASE_INSENSITIVE} and {@link Pattern#UNICODE_CASE}, which
+     * compares characters via {@code toLowerCase(toUpperCase(c))}. {@link String#toLowerCase(Locale)} does not
+     * always agree with that, so deriving the key one way and the lookup the other lets the pattern match text
+     * that the map has no entry for. Two concrete divergences: {@code U+017F LATIN SMALL LETTER LONG S} matches
+     * the literal {@code s} but {@code toLowerCase} leaves it unchanged, and {@code U+0130 LATIN CAPITAL LETTER I
+     * WITH DOT ABOVE} matches the literal {@code i} but {@code toLowerCase} decomposes it to {@code i U+0307}.
```
jglick, line 77: Sorry, _what_? There is no reason for any Unicode craziness here AFAIK. Just use simple per-code-point lowercasing.

What is _the pattern_ being discussed, and why would it be using https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/regex/Pattern.html#UNICODE_CASE ?

TBH I am not sure why anything in anonymization is case-insensitive to begin with. Job names, for example, are assumed to be case-sensitive. I cannot think of any circumstance where it would be necessary for an anonymized support bundle to fold case. If you are running Jenkins on a Windows server (quite rare these days I think), then you could not create both `%JENKINS_HOME%\jobs\myfolder\` and `%JENKINS_HOME%\jobs\MyFolder\`, but whichever one you _did_ create, that would be its name in all texts that a support bundle might render.

jglick: https://github.com/jenkinsci/support-core-plugin/blob/7a1f11e6beae0423006b600a3d00fe4e9588ae57/src/main/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilter.java#L114 I guess?

https://github.com/jenkinsci/support-core-plugin/pull/473

> Note: I thought a further improvement could be to use case insensitive mapping for the replacement. Here, we convert mapping to lowercase to create a replacement map. And then during the processing of a file, we convert the line to lowercase when passing it to the matcher. That help us achieve case insensitive word replacement. I thought that maybe creating a "case insensitive replacement map" could help save even more time during processing... but maybe at a cost of more resource (memory) usage. Actually when using some of the case insensitive implementation, we don't save time. It's quite similar (or worse in case of the TreeMap with comparator).

…so why are we doing it? And why a3a8324333ece8a2dbc2c6cc3ab6c7f41f62aa0d?

I think we could start by just dropping `UNICODE_CASE`, which would I hope fix #981, and then maybe just drop all the case manipulation and KISS.

src/test/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilterTest.java
```diff
@@ -108,4 +108,26 @@ void shouldNotFilterOperatingSystem(JenkinsRule j) throws Exception {
...
+        assertThat(filter.filter("build ſapphirium done")).isEqualTo("build item_test_alpha done");
+
+        // U+0130 LATIN CAPITAL LETTER I WITH DOT ABOVE matches the literal 'i', but String#toLowerCase
+        // decomposes it to 'i' + U+0307 COMBINING DOT ABOVE.
+        assertThat(filter.filter("job İdalium ran")).isEqualTo("job item_test_beta ran");
+
+        // plain ASCII case-insensitivity must keep working
+        assertThat(filter.filter("SAPPHIRIUM")).isEqualTo("item_test_alpha");
+        assertThat(filter.filter("sapphirium")).isEqualTo("item_test_alpha");
+    }
```
RanjitCS, line 132: this wont pass after https://github.com/jenkinsci/support-core-plugin/pull/980 is merged. we use 
```
String key = fold(original);
```
while https://github.com/jenkinsci/support-core-plugin/pull/980/changes#diff-bf122a5dbcea9fd49f1038bc73ee99d52f5d2bdbf297a6deecec17719b2e9240R80-R87 looks up by `match.toLowerCase()`, we should use `fold()` for every key

jglick: > after #980 is merged

Seems like a good opportunity to try GH Stacks.

mikecirioli: AFAICT you can't use the native `gh stacks` with our fork-based PR model

jglick: Right, not without pushing origin branches.

Feel free to disregard, I just thought it might simplify the interactions between these PRs.

mikecirioli:
Superseded by #988 — moves case normalization out of the regex engine rather than aligning the key derivation with it.

Your question about why anonymization is case-insensitive at all is still open; #988 only makes the existing behaviour self-consistent.

PR #984, opened 2026-08-18
Emit the anonymization regex from a sorted word list instead of building a trie

### LINKED ISSUE #983: WordsTrie builds a multi-GB scratch tree to emit a regex that is then discarded
`SensitiveContentFilter.reload()` builds a `WordsTrie`, calls `getRegex()`, and drops it:

```java
final WordsTrie trie = new WordsTrie();
...
trie.add(lowerCaseOriginal);
...
Pattern.compile("(?<!\\w)" + trie.getRegex() + "(?!\\w)", CASE_INSENSITIVE | UNICODE_CASE);
// trie is now unreachable
```

The tree is pure scratch, but every `TrieNode` owns a `TreeMap<Character, TrieNode>`, so it ends up far larger than the regex it exists to produce.

Measured with ~1.5M mapping entries (54.7M characters of originals):

| | |
|---|---|
| `TrieNode` instances | 26.6M |
| tree (`TrieNode` + `TreeMap` + `Entry` + `EntrySet`) | **~3.2 GB** |
| regex produced | 30.8M chars |
| regex `String` + compiled `Pattern` retained afterwards | ~337 MB |

So the throwaway structure costs roughly 10x the artifact it produces, and it is rebuilt on every bundle generation — hourly by default, since automated bundles are enabled out of the box. On instances with large mapping tables this is the dominant memory cost of anonymization and can exhaust the heap.

No tree is needed to emit a prefix-shared regex: if the words are sorted, every group sharing a prefix is contiguous, so a recursive walk over the sorted range can emit the same regex with `O(longest word)` working memory. Locally that produces byte-identical output with peak dropping from ~3.2 GB to ~90 MB for the same input, and `add()` becoming a list append rather than a tree descent.

Related: #811, #980, #982.

mikecirioli:
`WordsTrie.getRegex()` is the OOM site. `SensitiveContentFilter.reload()` builds a trie of `TrieNode` objects — each owning a `TreeMap<Character, TrieNode>` — just to emit a regex string, then drops the trie. At large mapping counts the tree's per-entry overhead dwarfs both the mappings table and the regex it produces. The cost is proportional to the number of live entries, not to how many are stale, so eviction alone does not address installations with legitimately large mapping tables.

This PR replaces the tree with a sorted-array walk. Sorting makes each shared-prefix group contiguous, so the same regex can be emitted by recursing over ranges with `O(longest word)` working memory. `add()` becomes a list append; `getRegex()` sorts once, deduplicates in place, and recurses. No node objects are allocated.

The emitted regex is **byte-identical** to the old implementation — that is the safety property, so no matching behaviour changes. Verified by a golden-hash fixture covering 10,000 words with shared prefixes, metacharacters, duplicates, and surrogate pairs.

`add("")` used to fail at `charAt(0)`; a list append accepts it silently. The pre-fill loop over persisted mappings does not filter blanks the way the `NameProvider` loop does, so a blank original could reach the trie and produce an empty-matching expression. Blank words are now skipped.

Fixes #983

### Related

- #980 — evict stale mappings (addresses unbounded *growth*, not per-entry cost)
- #986, #987, #988 — correctness and cleanup from the same investigation
- #811 — original report


src/test/java/com/cloudbees/jenkins/support/filter/LegacyWordsTrie.java
```diff
@@ -0,0 +1,128 @@
...
+import java.util.TreeMap;
+import java.util.function.Function;
+import java.util.regex.Pattern;
+
+/**
+ * Verbatim copy of the pre-rewrite, tree-based {@code WordsTrie} implementation, kept test-only so
+ * {@link WordsTrieTest} can differentially assert that the sorted-list rewrite produces byte-identical regex
+ * output. Do not use in main code: this is the implementation being replaced because of its memory footprint.
+ */
+class LegacyWordsTrie {
```
mikecirioli: debated about including this - initially used to verify my approach

jglick: It does not make sense to commit this code to trunk sources, as it is of no interest going forward. I would rather suggest a more straightforward TDD approach: pick some _deterministic_ large input (`Generate` with a seed, or simply some modestly-sized text file in `src/test/resources/`); revert `src/main/` changes temporarily; verify that the test passes (so it accurately captures what the original algo produced); then reapply your proposed changes, and again verify that the test passes.

src/test/java/com/cloudbees/jenkins/support/filter/WordsTrieTest.java
```diff
@@ -0,0 +1,163 @@
...
+                "a" + grinning + "b",
+                "a" + grinning + "c",
+                fraktur + "oo",
+                "İstanbul",
+                "ſword",
+                "ſ",
+                "İ",
+                grinning,
+                fraktur));
+    }
```
jglick, line 155: https://github.com/jenkinsci/support-core-plugin/pull/982#discussion_r3808786745 I would suggest cleaning up case folding first, at which point other stuff should be simpler (and perhaps faster as well).

src/main/java/com/cloudbees/jenkins/support/filter/WordsTrie.java
```diff
@@ -69,122 +63,136 @@ public void add(String word) {
      * @return the regex String of this Trie.
      */
     public String getRegex() {
-        return root.getRegex();
+        if (words.isEmpty()) {
+            // No data, stop here
+            return null;
+        }
+        String[] sorted = words.toArray(new String[0]);
```
jglick: Personally I prefer
```suggestion
        String[] sorted = words.toArray(String[]::new);
```
though it hardly matters.

src/main/java/com/cloudbees/jenkins/support/filter/WordsTrie.java
```diff
@@ -41,26 +45,16 @@ public class WordsTrie {
...
-        TrieNode ref = root;
-        int i = 0;
-        while (i < word.length() - 1) {
-            // Fill in down the Trie
-            ref = ref.getOrCreate(word.charAt(i), s -> new TrieNode(false));
-            i++;
-        }
-        // We need to mark the last node as an end node
-        ref.getOrCreate(word.charAt(i), s -> new TrieNode(true)).end = true;
+        words.add(word);
```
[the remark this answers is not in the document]
Dohbedoh: This would have failed loudly on the current branch. Now it's silent and would cause borked outputs in the bundle data. We just need a guard against blank string here to be defensive.

Dohbedoh:
Thanks @mikecirioli. That looks good to me. I have use claude to run tests comparison between master and this on the similar factor that I used back in the days (deep vs flat Trie - random names vs. prefixed, large amount of data and lines, large vs low replacement hit, large vs small content mapping). Short summary is that output are byte identitcal, time benefit is not significant, memory benefit fix is huge and whetever I throw at it 7MB flat is used to generate the regexp.

Just the guard on the `add()` I feel is needed.

src/main/java/com/cloudbees/jenkins/support/filter/WordsTrie.java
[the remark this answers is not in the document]
mikecirioli: Fixed — `add()` now ignores blank words, with regression cases for blank alone, blank mixed with real words, and whitespace-only.

Ignoring rather than throwing: a blank original already sitting in someone's mappings file shouldn't abort bundle generation, and the old `charAt(0)` failure was incidental rather than intended. `isBlank()` rather than `isEmpty()` because whitespace-only would otherwise emit a literal whitespace alternative in the pattern.

Note the property generator still has a minimum word size of 1, so the explicit cases cover this rather than the property test.

src/test/java/com/cloudbees/jenkins/support/filter/WordsTrieTest.java
mikecirioli: ! went the other way on ordering, for a concrete reason: the case-folding change doesn't touch WordsTrie at all. It changes what the call site passes in - `trie.add(normalizeCase(original))` instead of `trie.add(original.toLowerCase(ENGLISH))` - so this class and its tests are unaffected either way, and there's no rework to avoid by waiting.

src/main/java/com/cloudbees/jenkins/support/filter/WordsTrie.java
```diff
@@ -69,122 +66,136 @@ public void add(String word) {
      * @return the regex String of this Trie.
      */
     public String getRegex() {
-        return root.getRegex();
+        if (words.isEmpty()) {
+            // No data, stop here
+            return null;
```
jglick, line 72: a `@CheckForNull` would be nice

jglick (approved):
OK to the extent that I follow it.

Dohbedoh (approved):

PR #986, opened 2026-08-18
Make anonymization pseudonyms unique and reproducible

### LINKED ISSUE #985: Generated pseudonyms are exhaustible and not reproducible
`ContentMappings` pseudonyms are an `adjective_noun` pair from `wordnet-random-name`, so only 611 * 2407 = 1,470,677 distinct values exist.

`RandomNameGenerator.next()` walks a cursor - `pos = Math.abs(pos + prime) % size`, with `gcd(prime, size) == 1` - so it is an exact permutation: collision-free until it exhausts the space, then it repeats in the same order. Checked against the shipped jar: `size=1470677`, `prime=3263443`, 1,470,677 distinct names, first repeat at draw 1,470,678.

Two consequences:

1. **Tables larger than 1.47M entries reuse pseudonyms**, and nothing checks `getReplacement()` for uniqueness. Two distinct originals then render identically in a bundle, and reversing a pseudonym through the mappings table is ambiguous. Tables that large do occur in practice.

2. **Pseudonyms are not reproducible.** The cursor is seeded from `System.currentTimeMillis()`, so the same original gets a different pseudonym on another instance, or after the mappings file is lost.

On severity: a collision leaks *less* information, not more, so this is a correctness bug in the promise that one pseudonym identifies one thing - not a security issue.

mikecirioli:
Pseudonyms are an `adjective_noun` pair drawn from `wordnet-random-name`, so only 611 * 2407 = 1,470,677 distinct values exist. `RandomNameGenerator.next()` walks a cursor (`pos = Math.abs(pos + prime) % size`, and `gcd(prime, size) == 1`), making it an exact permutation: collision-free until it exhausts the space, then it repeats in the same order. Tables larger than that reuse pseudonyms, and nothing checks `getReplacement()` for uniqueness - two distinct originals render identically, and reversing a pseudonym through the mappings table is ambiguous.

A pseudonym is a function of *when* an original was first seen - its position in the cursor walk - not of the original itself, and the walk starts at `System.currentTimeMillis()`. So the same original gets a different pseudonym on another instance, or after the mappings file is lost.

Both parts now derive from `HMACConfidentialKey.mac(byte[])`. The new 8-character hex tail carries the uniqueness - a third word would still leave ~1000 expected collisions at 1.5M entries, since hashing costs you the cursor's collision-freedom and puts you on the birthday bound - while the word pair stays for readability. Keying it from the instance secret is load-bearing: an unsalted hash would let anyone holding a bundle brute-force guessable names.

Existing mappings are untouched. `getMappingOrCreate` does `computeIfAbsent(original, generator)`, and the map is keyed by original, so the generator never runs for an original already in the table - already-minted pseudonyms are preserved and a table just holds a mix of both formats.

That does mean reproducibility only applies to newly-minted entries. A pre-existing original keeps whatever random pseudonym it was given, so two instances that already have divergent entries for it stay divergent, and losing the mappings file re-mints it to a stable but different value. Neither is a regression - both cases produce a random value today - but the guarantee arrives as entries turn over rather than retroactively. The same is true of collisions: a table that has already exhausted the namespace keeps whatever duplicate pseudonyms it accumulated, since those entries are preserved. Widening the namespace prevents new collisions rather than resolving existing ones - clearing the mappings file is the only way to retire them, at the cost of every pseudonym changing.

Two `@Restricted(NoExternalUse.class)` signatures change rather than gaining overloads: `DataFaker.apply` differs only in return type so it can't be overloaded, and a no-arg `NameProvider.generateFake()` can't satisfy the reproducibility invariant - it could only return something random, which is the bug. 

### From review

The MAC is now sized to what it uses. An 8-byte MAC is requested via the three-argument constructor and all of it is consumed: four bytes for the dictionary index, four for the hex tail. Previously a full 256-bit MAC was requested and 64 bits used.

Four bytes rather than three for the index also removes a modulo bias: 2^24 over a 1,470,677-word dictionary left some words roughly 9% likelier than others, where 2^32 brings that to about 0.03%.

The index derivation is funnelled through one method using `Math.floorMod`, which takes the sign of the divisor and therefore cannot return a negative index for any input. Java's `%` takes the sign of the *dividend*, which is the same trapdoor behind the library's own `Math.abs(Integer.MIN_VALUE)` path — reachable there because `Math.abs` returns `Integer.MIN_VALUE` unchanged. A test feeds hostile values straight to that method and fails with `IndexOutOfBoundsException` if `floorMod` is reverted to `%`.

On `Dictionary.INSTANCE` — it is package-private, so constructing our own `Dictionary` is the only option without reflection. `javap -p` confirms: the class, its no-argument constructor, `size()` and `word(int)` are all public; `INSTANCE` is not.

Determinism is now proven across an actual restart via `JenkinsSessionExtension` rather than within a single JVM, which is the property being claimed. The two tests exercising `HMACConfidentialKey` rather than this class are gone.

Fixes #985

### Related

Five PRs from the same pass over the anonymization code, plus the original report in #811:

- #980
- #984
- #986 *(this one)*
- #987
- #988

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/support/filter/DataFaker.java
+++ b/src/main/java/com/cloudbees/jenkins/support/filter/DataFaker.java
@@ -33,7 +33,7 @@ import java.util.function.Function;
 import jenkins.security.HMACConfidentialKey;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
-import org.kohsuke.randname.RandomNameGenerator;
+import org.kohsuke.randname.Dictionary;
 
 /**
  * Provides deterministic pseudonym generation for anonymization.
@@ -48,6 +48,7 @@ import org.kohsuke.randname.RandomNameGenerator;
 public class DataFaker implements ExtensionPoint, Function<Function<String, String>, Function<String, String>> {
 
     private static final HMACConfidentialKey PSEUDONYMS = new HMACConfidentialKey(DataFaker.class, "pseudonyms");
+    private static final Dictionary DICTIONARY = new Dictionary();
 
     /**
      * @return the singleton instance
@@ -66,14 +67,11 @@ public class DataFaker implements ExtensionPoint, Function<Function<String, Stri
     public Function<String, String> apply(@NonNull Function<String, String> nameTransformer) {
         return original -> {
             String hex = PSEUDONYMS.mac(original);
-            // Masked because RandomNameGenerator computes Math.abs(pos + prime) % size, and
-            // Math.abs(Integer.MIN_VALUE) is negative, which would yield a negative word index.
-            int seed = (int) (Long.parseLong(hex.substring(0, 8), 16) & 0x0FFFFFFFL);
+            int index = (int) (Long.parseLong(hex.substring(0, 8), 16) % DICTIONARY.size());
             String tail = hex.substring(8, 16);
 
-            RandomNameGenerator generator = new RandomNameGenerator(seed);
             String name = nameTransformer
-                    .apply(generator.next())
+                    .apply(DICTIONARY.word(index))
                     .toLowerCase(Locale.ENGLISH)
                     .replace(' ', '_');
             return name + "_" + tail;
--- a/src/main/java/com/cloudbees/jenkins/support/filter/DataFaker.java
+++ b/src/main/java/com/cloudbees/jenkins/support/filter/DataFaker.java
@@ -28,6 +28,7 @@ import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.Extension;
 import hudson.ExtensionList;
 import hudson.ExtensionPoint;
+import java.nio.charset.StandardCharsets;
 import java.util.Locale;
 import java.util.function.Function;
 import jenkins.security.HMACConfidentialKey;
@@ -40,16 +41,14 @@ import org.kohsuke.randname.Dictionary;
  *
  * <p>Derives stable pseudonyms from originals using HMAC-keyed word pairs and hex tails.
  * Same original always produces the same pseudonym within one installation (stable across
- * restarts and HA replicas sharing JENKINS_HOME). Deliberately differs across installations
+ * restarts and CloudBees CI HA replicas sharing JENKINS_HOME). Deliberately differs across installations
  * with different secret keys, preventing brute-force attacks on guessable originals.
- *
- * @since TODO
  */
 @Extension
 @Restricted(NoExternalUse.class)
 public class DataFaker implements ExtensionPoint, Function<Function<String, String>, Function<String, String>> {
 
-    private static final HMACConfidentialKey PSEUDONYMS = new HMACConfidentialKey(DataFaker.class, "pseudonyms");
+    private static final HMACConfidentialKey PSEUDONYMS = new HMACConfidentialKey(DataFaker.class, "pseudonyms", 8);
     private static final Dictionary DICTIONARY = new Dictionary();
 
     /**
@@ -57,6 +56,21 @@ public class DataFaker implements ExtensionPoint, Function<Function<String, Stri
      */
     public static DataFaker get() {
         return ExtensionList.lookupSingleton(DataFaker.class);
+    }
+
+    /**
+     * Maps MAC-derived bits to a dictionary word, guaranteed never to produce a negative index.
+     *
+     * <p>Uses {@code Math.floorMod} rather than {@code %} because Java's {@code %} operator
+     * takes the sign of the dividend, so a negative input yields a negative index.
+     * {@code floorMod} takes the sign of the divisor, guaranteeing a non-negative result
+     * for any input including {@code Long.MIN_VALUE}.
+     *
+     * @param macBits MAC-derived value to map to dictionary index
+     * @return dictionary word at the computed index
+     */
+    static String wordFor(long macBits) {
+        return DICTIONARY.word(Math.floorMod(macBits, DICTIONARY.size()));
     }
 
     /**
@@ -68,14 +82,16 @@ public class DataFaker implements ExtensionPoint, Function<Function<String, Stri
     @Override
     public Function<String, String> apply(@NonNull Function<String, String> nameTransformer) {
         return original -> {
-            String hex = PSEUDONYMS.mac(original).toLowerCase(Locale.ROOT);
-            // Index Dictionary directly rather than RandomNameGenerator.next() because next()
-            // computes Math.abs(pos + prime) % size and Math.abs(Integer.MIN_VALUE) is negative.
-            int index = (int) (Long.parseLong(hex.substring(0, 8), 16) % DICTIONARY.size());
-            String tail = hex.substring(8, 16);
+            byte[] mac = PSEUDONYMS.mac(original.getBytes(StandardCharsets.UTF_8));
+            // First 4 bytes for dictionary index, next 4 for the 8-hex-character tail
+            long idx = ((long) (mac[0] & 0xFF) << 24)
+                    | ((long) (mac[1] & 0xFF) << 16)
+                    | ((long) (mac[2] & 0xFF) << 8)
+                    | (mac[3] & 0xFF);
+            String tail = String.format("%02x%02x%02x%02x", mac[4] & 0xFF, mac[5] & 0xFF, mac[6] & 0xFF, mac[7] & 0xFF);
 
             String name = nameTransformer
-                    .apply(DICTIONARY.word(index))
+                    .apply(wordFor(idx))
                     .toLowerCase(Locale.ENGLISH)
                     .replace(' ', '_');
             return name + "_" + tail;
--- a/src/main/java/com/cloudbees/jenkins/support/filter/DataFaker.java
+++ b/src/main/java/com/cloudbees/jenkins/support/filter/DataFaker.java
@@ -30,19 +30,24 @@ import hudson.ExtensionList;
 import hudson.ExtensionPoint;
 import java.util.Locale;
 import java.util.function.Function;
-import java.util.function.Supplier;
+import jenkins.security.HMACConfidentialKey;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.kohsuke.randname.RandomNameGenerator;
 
 /**
- * Provides a way to generate random names.
+ * Provides deterministic pseudonym generation for anonymization.
+ *
+ * <p>Derives stable pseudonyms from originals using HMAC-keyed word pairs and hex tails.
+ * Same original always produces the same pseudonym across instances and over time.
  *
  * @since TODO
  */
 @Extension
 @Restricted(NoExternalUse.class)
-public class DataFaker implements ExtensionPoint, Function<Function<String, String>, Supplier<String>> {
+public class DataFaker implements ExtensionPoint, Function<Function<String, String>, Function<String, String>> {
+
+    private static final HMACConfidentialKey PSEUDONYMS = new HMACConfidentialKey(DataFaker.class, "pseudonyms");
 
     /**
      * @return the singleton instance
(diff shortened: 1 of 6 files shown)


src/main/java/com/cloudbees/jenkins/support/filter/DataFaker.java
```diff
@@ -30,19 +30,27 @@
...
 /**
- * Provides a way to generate random names.
+ * Provides deterministic pseudonym generation for anonymization.
+ *
+ * <p>Derives stable pseudonyms from originals using HMAC-keyed word pairs and hex tails.
+ * Same original always produces the same pseudonym within one installation (stable across
+ * restarts and HA replicas sharing JENKINS_HOME). Deliberately differs across installations
+ * with different secret keys, preventing brute-force attacks on guessable originals.
  *
  * @since TODO
```
jglick:
```suggestion
```
meaningless for a non-API

src/main/java/com/cloudbees/jenkins/support/filter/DataFaker.java
```diff
@@ -30,19 +30,27 @@
...
-import org.kohsuke.randname.RandomNameGenerator;
+import org.kohsuke.randname.Dictionary;
 
 /**
- * Provides a way to generate random names.
+ * Provides deterministic pseudonym generation for anonymization.
+ *
+ * <p>Derives stable pseudonyms from originals using HMAC-keyed word pairs and hex tails.
+ * Same original always produces the same pseudonym within one installation (stable across
+ * restarts and HA replicas sharing JENKINS_HOME). Deliberately differs across installations
```
jglick:
```suggestion
 * restarts and CloudBees CI HA replicas sharing JENKINS_HOME). Deliberately differs across installations
```

src/main/java/com/cloudbees/jenkins/support/filter/DataFaker.java
```diff
@@ -51,16 +59,26 @@ public static DataFaker get() {
...
-                .apply(generator.next())
-                .toLowerCase(Locale.ENGLISH)
-                .replace(' ', '_');
+    public Function<String, String> apply(@NonNull Function<String, String> nameTransformer) {
+        return original -> {
+            String hex = PSEUDONYMS.mac(original).toLowerCase(Locale.ROOT);
+            // Index Dictionary directly rather than RandomNameGenerator.next() because next()
+            // computes Math.abs(pos + prime) % size and Math.abs(Integer.MIN_VALUE) is negative.
+            int index = (int) (Long.parseLong(hex.substring(0, 8), 16) % DICTIONARY.size());
+            String tail = hex.substring(8, 16);
```
jglick: Might be clearer to use https://javadoc.jenkins.io/jenkins/security/HMACConfidentialKey.html#mac(byte%5B%5D) rather than doing string manipulation on hex.

src/main/java/com/cloudbees/jenkins/support/filter/DataFaker.java
```diff
@@ -30,19 +30,27 @@
...
+ * with different secret keys, preventing brute-force attacks on guessable originals.
  *
  * @since TODO
  */
 @Extension
 @Restricted(NoExternalUse.class)
-public class DataFaker implements ExtensionPoint, Function<Function<String, String>, Supplier<String>> {
+public class DataFaker implements ExtensionPoint, Function<Function<String, String>, Function<String, String>> {
+
+    private static final HMACConfidentialKey PSEUDONYMS = new HMACConfidentialKey(DataFaker.class, "pseudonyms");
```
jglick: Use https://javadoc.jenkins.io/jenkins/security/HMACConfidentialKey.html#%3Cinit%3E(java.lang.Class,java.lang.String,int) to pick the length of a MAC you actually expect to use. From what I can see, you are discarding all but the first 16 hex chars ~ 8 bytes ~ 64 bits of a 256-bit MAC.

src/main/java/com/cloudbees/jenkins/support/filter/DataFaker.java
```diff
@@ -51,16 +59,26 @@ public static DataFaker get() {
...
-    public Supplier<String> apply(@NonNull Function<String, String> nameTransformer) {
-        return () -> nameTransformer
-                .apply(generator.next())
-                .toLowerCase(Locale.ENGLISH)
-                .replace(' ', '_');
+    public Function<String, String> apply(@NonNull Function<String, String> nameTransformer) {
+        return original -> {
+            String hex = PSEUDONYMS.mac(original).toLowerCase(Locale.ROOT);
+            // Index Dictionary directly rather than RandomNameGenerator.next() because next()
+            // computes Math.abs(pos + prime) % size and Math.abs(Integer.MIN_VALUE) is negative.
```
jglick: This comment confuses me. Beyond whatever integer overflow bug might be in the lib, https://github.com/kohsuke/wordnet-random-name/blob/91f2497c0c98a0400474c484a95cc984981f2521/src/main/java/org/kohsuke/randname/RandomNameGenerator.java#L23 is clearly not suitable anyway since we need to pick a word based on an index from a MAC, not just pick the next word in some sequence.
```suggestion
```

src/main/java/com/cloudbees/jenkins/support/filter/DataFaker.java
```diff
@@ -30,19 +30,27 @@
...
  *
  * @since TODO
  */
 @Extension
 @Restricted(NoExternalUse.class)
-public class DataFaker implements ExtensionPoint, Function<Function<String, String>, Supplier<String>> {
+public class DataFaker implements ExtensionPoint, Function<Function<String, String>, Function<String, String>> {
+
+    private static final HMACConfidentialKey PSEUDONYMS = new HMACConfidentialKey(DataFaker.class, "pseudonyms");
+    private static final Dictionary DICTIONARY = new Dictionary();
```
jglick, line 52: Why not just use `Dictionary.INSTANCE`?

src/test/java/com/cloudbees/jenkins/support/filter/DataFakerTest.java
```diff
@@ -0,0 +1,148 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2018, CloudBees, Inc.
```
jglick:
```suggestion
 * Copyright (c) 2026, CloudBees, Inc.
```
presumably

src/test/java/com/cloudbees/jenkins/support/filter/DataFakerTest.java
```diff
@@ -0,0 +1,148 @@
...
+
+    @Test
+    void deterministicPseudonymGeneration(JenkinsRule r) {
+        DataFaker faker = new DataFaker();
+        String original = "testuser";
+
+        String pseudonym1 = faker.apply(name -> "user_" + name).apply(original);
+        String pseudonym2 = faker.apply(name -> "user_" + name).apply(original);
+
+        assertEquals(pseudonym1, pseudonym2, "Same original should produce identical pseudonym");
```
jglick: Maybe more interesting to use `JenkinsSessionExtension` to show that the value is consistent across restarts.

src/test/java/com/cloudbees/jenkins/support/filter/DataFakerTest.java
```diff
@@ -0,0 +1,148 @@
...
+
+        assertEquals(foo1, foo2, "Order of minting should not affect pseudonym for 'foo'");
+        assertEquals(bar1, bar2, "Order of minting should not affect pseudonym for 'bar'");
+    }
+
+    @Test
+    void collisionSanityCheck(JenkinsRule r) {
+        DataFaker faker = new DataFaker();
+        Set<String> pseudonyms = new HashSet<>();
+        int iterations = 200000;
```
jglick, line 110: How long does this take?

src/test/java/com/cloudbees/jenkins/support/filter/DataFakerTest.java
```diff
@@ -0,0 +1,148 @@
...
+        for (int i = 0; i < iterations; i++) {
+            String pseudonym = faker.apply(name -> "user_" + name).apply("original" + i);
+            pseudonyms.add(pseudonym);
+        }
+
+        assertEquals(iterations, pseudonyms.size(), "Expected zero collisions in " + iterations + " iterations");
+    }
+
+    @Test
+    void legacyPseudonymPreservation(JenkinsRule r) {
```
jglick, line 122: This should use `@LocalData`.

src/test/java/com/cloudbees/jenkins/support/filter/DataFakerTest.java
```diff
@@ -0,0 +1,148 @@
...
+    void keyedPseudonyms(JenkinsRule r) {
+        HMACConfidentialKey key1 = new HMACConfidentialKey(DataFaker.class, "pseudonyms");
+        HMACConfidentialKey key2 = new HMACConfidentialKey(DataFaker.class, "something-else");
+
+        String original = "testuser";
+        String mac1 = key1.mac(original);
+        String mac2 = key2.mac(original);
+
+        assertNotEquals(mac1, mac2, "Different keys should produce different MACs for same original");
+    }
```
jglick: Well, yes, but this is not testing `DataFaker` code at all. Delete it.

src/test/java/com/cloudbees/jenkins/support/filter/DataFakerTest.java
```diff
@@ -0,0 +1,148 @@
...
+    void macLengthSufficient(JenkinsRule r) {
+        HMACConfidentialKey key = new HMACConfidentialKey(DataFaker.class, "pseudonyms");
+        String mac = key.mac("test");
+
+        assertTrue(
+                mac.length() >= 16,
+                "MAC must be at least 16 hex chars to safely extract seed (8 chars) and tail (8 chars). Got: "
+                        + mac.length());
+        assertEquals(64, mac.length(), "Full HMAC-SHA256 should be 64 hex chars (32 bytes)");
+    }
```
jglick: Ditto.

src/test/java/com/cloudbees/jenkins/support/filter/DataFakerTest.java
```diff
@@ -0,0 +1,148 @@
...
+import org.junit.jupiter.api.Test;
+import org.jvnet.hudson.test.JenkinsRule;
+import org.jvnet.hudson.test.junit.jupiter.WithJenkins;
+
+@WithJenkins
+class DataFakerTest {
+
+    @Test
+    void pseudonymsMatchExpectedFormat(JenkinsRule r) {
+        DataFaker faker = new DataFaker();
```
jglick:
```suggestion
        DataFaker faker = ExtensionList.lookupSingleton(DataFaker.class);
```
and similarly elsewhere

src/test/java/com/cloudbees/jenkins/support/filter/DataFakerTest.java
```diff
@@ -0,0 +1,148 @@
...
+    void tailIsExactlyEightHexChars(JenkinsRule r) {
+        DataFaker faker = new DataFaker();
+
+        for (int i = 0; i < 100; i++) {
+            String pseudonym = faker.apply(name -> "user_" + name).apply("test" + i);
+            assertTrue(pseudonym.matches("^.+_[0-9a-f]{8}$"), "Pseudonym should end with underscore and 8 hex chars");
+
+            String tail = pseudonym.substring(pseudonym.lastIndexOf('_') + 1);
+            assertEquals(8, tail.length(), "Tail should be exactly 8 characters");
+            assertTrue(tail.matches("[0-9a-f]{8}"), "Tail should be 8 lowercase hex characters");
```
jglick:
```suggestion
```

Appears to add nothing to the previous assertion?

src/main/java/com/cloudbees/jenkins/support/filter/DataFaker.java
```diff
@@ -68,14 +82,16 @@ public static DataFaker get() {
...
-            // computes Math.abs(pos + prime) % size and Math.abs(Integer.MIN_VALUE) is negative.
-            int index = (int) (Long.parseLong(hex.substring(0, 8), 16) % DICTIONARY.size());
-            String tail = hex.substring(8, 16);
+            byte[] mac = PSEUDONYMS.mac(original.getBytes(StandardCharsets.UTF_8));
+            // First 4 bytes for dictionary index, next 4 for the 8-hex-character tail
+            long idx = ((long) (mac[0] & 0xFF) << 24)
+                    | ((long) (mac[1] & 0xFF) << 16)
+                    | ((long) (mac[2] & 0xFF) << 8)
+                    | (mac[3] & 0xFF);
+            String tail = String.format("%02x%02x%02x%02x", mac[4] & 0xFF, mac[5] & 0xFF, mac[6] & 0xFF, mac[7] & 0xFF);
```
jglick, line 91: Possibly written more concisely with `BigInteger` (not sure offhand).

mikecirioli:
Superseded by the combined PR (deterministic pseudonyms + case normalization + dead-code removal in one reviewable branch). This avoids the cross-PR conflict in SensitiveContentFilter.java and the ordering dependency between #987 and #988.

PR #987, opened 2026-08-21
Remove the unreachable in-place replacement path and its per-mapping data

mikecirioli:
> Draft: based on #980's branch, so the diff here includes #980 until that merges.
> This PR's own change only: https://github.com/mikecirioli/support-core-plugin/compare/JENKINS-55995...drop-dead-replacement-path
> Merge order: #984 → #980 → #986 → this → #988.

`ContentMapping.filter()` has no caller anywhere in `src/main` — it exists only because the class declares `implements ContentFilter`. But its constructor still runs for every mapping, building a set of up to four name variants into a `String[]` plus a second `String[]` of repeated references. On a 1.5M-entry table that is ~3.1M array objects, roughly 56 MB by object-layout arithmetic, retained to serve a method nobody calls.

Those variants were an attempt to also match HTML-escaped (`a&lt;b`) and folder-display (`folder » job`) forms of a name. Since the consumer is unreachable, that coverage is currently provided by nothing — `SensitiveContentFilter` feeds the trie only the raw original. Only 0.3% of real originals contain `/` and 0.01% contain an HTML-special character, so reviving it is not worth it.

Deleting `filter()` makes the in-place `WordReplacer` family unreachable in turn — the `String[]`/`StringBuilder` overloads and the singular `replaceWord` variants. The `Pattern`-based overloads are the live path and stay. `ContentMapping` no longer implements `ContentFilter`; it remains `@Restricted(NoExternalUse.class)`, so this is not a public API change.

### Testing done

The persisted form is unaffected: `ContentMapping` serializes via `writeReplace` and a proxy carrying only `original` and `replacement`, so the deleted fields never appeared on disk and no migration is involved.

`WordReplacerTest` loses the cases for the deleted methods, including a differential test that compared the in-place and `Pattern` implementations — with one side gone it has nothing left to compare. The `Pattern`-based tests are kept, including the full-Unicode scope test.

### Submitter checklist
- [x] Ensure you have provided tests that demonstrate the feature works or the issue is fixed


mikecirioli:
Superseded by the combined PR (deterministic pseudonyms + case normalization + dead-code removal in one reviewable branch). This avoids the cross-PR conflict in SensitiveContentFilter.java and the ordering dependency between #987 and #988.

PR #992, opened 2026-08-29
Deterministic pseudonyms, case normalization, dead-code removal

### LINKED ISSUE #981: Case-folding mismatch in SensitiveContentFilter: NPE during bundle generation, and some names silently never anonymized

mikecirioli:
Correctness and cleanup from the same investigation as #984 and #980. Supersedes #986, #987, #988 — combined here because they conflict in `SensitiveContentFilter.java` and #987's dead-code removal is only valid after #988 lands.

### Commits

**Derive pseudonyms from the original and widen their namespace** (was #986)
- `DataFaker` uses `HMACConfidentialKey` keyed on the original string → same original always produces the same pseudonym within one installation. Stable across restarts and HA replicas sharing `JENKINS_HOME`.
- Namespace widened from ~370k to ~370M combinations (611 words × 8 hex chars), eliminating the collision plateau the old cursor-based approach hit.

**Move case normalization out of the regex engine** (was #988)
- One per-code-point function (`Character.toLowerCase(Character.toUpperCase(cp))`) normalizes keys, trie words, stop words and input text. Pattern compiles with no flags; replacements splice into the original by offset.
- Fixes: NPE when `ſap` matched `sap`'s key (`toLowerCase` left `ſ` alone → null replacement); `İdalium` silently unredacted (`toLowerCase` grew the string, corrupting trie literals).
- Stop words compared via the same normalization, closing a bypass where two originals collapsed to one key but only one was gated.

**Remove the unreachable in-place replacement path** (was #987)
- `ContentMapping.filter()`, its per-mapping `originals`/`replacements` arrays, and the old `WordReplacer.replaceWord()`/`replaceWords()` statics — all dead once the offset-splice path above replaced them.

Fixes #981, #985

### Related

- #984 (merged) — regex emission memory fix
- #980 (merged) — stale mapping eviction


src/main/java/com/cloudbees/jenkins/support/filter/DataFaker.java
```diff
@@ -28,21 +28,28 @@
...
+import org.kohsuke.randname.Dictionary;
 
 /**
- * Provides a way to generate random names.
+ * Provides deterministic pseudonym generation for anonymization.
  *
- * @since TODO
+ * <p>Derives stable pseudonyms from originals using HMAC-keyed word pairs and hex tails.
+ * Same original always produces the same pseudonym within one installation (stable across
+ * restarts and CloudBees CI HA replicas sharing JENKINS_HOME). Deliberately differs across installations
```
jglick, line 44: That sounds good, but then do we still need `ContentMappings.mappings`?

```xml
<com.cloudbees.jenkins.support.filter.ContentMapping resolves-to="com.cloudbees.jenkins.support.filter.ContentMapping$SerializationProxy">
  <original>stuff</original>
  <replacement>item_socialist_lot_fd1134e5</replacement>
  <lastSeen>1790026397866</lastSeen>
</com.cloudbees.jenkins.support.filter.ContentMapping>
```

seems redundant if the `replacement` is derived from the `original` by an HMAC, right?

[CloudBees-internal link](https://cloudbees.atlassian.net/browse/BEE-69231)

mikecirioli: The `replacement` field is redundant now since `DataFaker.apply()` is a pure HMAC function of `original`. We don't need to store the pseudonym string; we can recompute it on demand.

  The map itself still has a job though. [`SensitiveContentFilter.reload()`](https://github.com/jenkinsci/support-core-plugin/blob/ace24ff36a03a192959085ce734f38e7b8504afb/src/main/java/com/cloudbees/jenkins/support/filter/SensitiveContentFilter.java#L154-L167) pre-fills the trie from 
  the persisted table before asking `NameProvider` for currently-live items:

  ```java
  // Pre-fill with existing mappings ...
  // This is required to filter out names of items that does not exist anymore, for which they could be
  // recorded in some content (such as log files that are anonymized when being written)
  // Once a job/user/node is deleted, NameProvider stops reporting it. The persisted table is the only record   that its name was ever sensitive.
```
 That's what lastSeen/touchMatched
  (https://github.com/jenkinsci/support-core-plugin/blob/ace24ff36a03a192959085ce734f38e7b8504afb/src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java#L156-L159) and evictStale()
  (https://github.com/jenkinsci/support-core-plugin/blob/ace24ff36a03a192959085ce734f38e7b8504afb/src/main/java/com/cloudbees/jenkins/support/filter/ContentMappings.java#L187-L208) are for: redaction of a deleted item's name fades out over the 90-day retention window instead of
  stopping the moment the item is deleted. Dropping the map would make that an immediate cutoff, which regresses anonymization.

  There's also DataFakerTest#legacyPseudonymPreservation (https://github.com/jenkinsci/support-core-plugin/blob/ace24ff36a03a192959085ce734f38e7b8504afb/src/test/java/com/cloudbees/jenkins/support/filter/DataFakerTest.java#L120-L133). Installs upgraded from before this HMAC scheme have
  non-HMAC pseudonyms already in ContentMappings.xml. If we stopped trusting the stored replacement and always recomputed via HMAC, those would change on upgrade.

  So: drop replacement from persistence, keep original + lastSeen, recompute the pseudonym via DataFaker on read, with a fallback for legacy entries during the retention window. Can split that into a follow-up rather than block this PR. Thoughts?

mikecirioli: Correction to my last reply: recomputing `replacement` on demand doesn't actually work.

Every `NameProvider` (ITEMS, USERS, NODES, COMPUTERS, VIEWS, LABELS) wraps `DataFaker` with a different prefix (`item_`, `user_`, etc). That prefix is only known while a `NameProvider` is actively reporting the name as live. For a mapping kept alive purely by `lastSeen` (its item was deleted, so no `NameProvider` reports it anymore), there's no way to know which prefix to regenerate with. `original` alone doesn't say whether it was a job, a user, or a node.

It seems the map isn't just tracking liveness, it's the only place that category information survives after deletion. Dropping `replacement` would need a new field to replace it (e.g. a stored prefix/category), which isn't smaller and doesn't remove complexity. I don't think there's a safe simplification here after all.

